PR #1, opened 2022-07-15
feat: merged chugsplash manager into registry

sam-goldman:
Merges ChugSplashManager into ChugSplashRegistry, making it easy to query events emitted from all deployments. This removes the concept of an 'owner' address which is in charge of each ChugSplashManager, and introduces the concept of a 'manager' address which is in charge of each project within ChugSplashRegistry. Function signatures have been updated to reflect this change.

Note that the yarn workspaces run order is now specified explicitly in the root package.json, which prevents an error that was occurring during the build. (The error was coming from [this line](https://github.com/smartcontracts/chugsplash/blob/develop/packages/demo/hardhat.config.ts#L5) in the demo/ sub-package, which references the plugins/ sub-package before it's built).


packages/contracts/contracts/ChugSplashRegistry.sol
```diff
@@ -1,50 +1,312 @@
...
-        // ChugSplashManager contract based on the project's name.
-        bytes32 salt = keccak256(bytes(_name));
-        registry[_name] = new ChugSplashManager{ salt: salt }(_owner);
-        emit ChugSplashProjectRegistered(_name, msg.sender, _owner, address(registry[_name]));
-        return address(registry[_name]);
+        ChugSplashProject storage project = projects[_name];
+        project.manager = _manager;
+        emit ChugSplashProjectRegistered(_name, msg.sender, _manager);
+
+        // TODO: We used to return the ChugSplashManager address here. Should we keep this new
```
smartcontracts, line 134: I think it's fine to remove now

packages/contracts/contracts/ChugSplashRegistry.sol
```diff
@@ -1,50 +1,312 @@
...
+     * @param _bundleSize Total number of actions in the bundle.
+     * @param _configUri URI pointing to the config file for the bundle.
+     */
+    function proposeChugSplashBundle(
+        string memory _name,
+        bytes32 _bundleHash,
+        uint256 _bundleSize,
+        string memory _configUri
+    ) public onlyManager(_name) {
+        ChugSplashProject storage project = projects[_name];
```
smartcontracts, line 153: An important here is to have the ability to have multiple proposed bundles for the same project at the same time (but only one active bundle at a time).

packages/contracts/contracts/ChugSplashRegistry.sol
```diff
@@ -1,50 +1,312 @@
...
+
+    /**
+     * Allows a manager to approve a bundle to be executed. Note that the bundle can be executed
+     * as soon as the bundle is approved.
+     *
+     * @param _name Name of the ChugSplash project.
+     * @param _bundleHash Hash of the bundle to approve.
+     */
+    function approveChugSplashBundle(string memory _name, bytes32 _bundleHash) public onlyManager(_name) {
+        ChugSplashProject storage project = projects[_name];
```
smartcontracts, line 173: Assuming we have multiple proposed bundles, we'll also want to check here that the `_bundleHash` corresponds to some bundle that has been previously proposed.

packages/contracts/contracts/ChugSplashRegistry.sol
```diff
@@ -1,50 +1,312 @@
...
+        bytes32[] memory _proof
+    ) public {
+        require(
+            isUpgrading(_name) == true,
+            "ChugSplashRegistry: no bundle has been approved for execution"
+        );
+
+        ChugSplashProject storage project = projects[_name];
+
+        // TODO: Confirm that this will do out-of-bounds checks.
```
smartcontracts, line 211: Yeah Solidity will do OOB checks here

packages/contracts/contracts/ChugSplashRegistry.sol
```diff
@@ -1,50 +1,312 @@
...
+            ),
+            "ChugSplashRegistry: invalid bundle action proof"
+        );
+
+        // Make sure the proxy has code in it and deploy the proxy if it doesn't. Since we're
+        // deploying via CREATE2, we can always correctly predict what the proxy address *should*
+        // be and can therefore easily check if it's already populated.
+        // TODO: See if there's a better way to handle this case because it messes with the gas
+        // cost of SET_CODE/SET_STORAGE operations in a somewhat unpredictable way.
+        ChugSplashProxy proxy = getProxyByName(_name);
```
smartcontracts, line 233: Not really related to this PR, but I realized a few weeks ago that it'd probably be better to support standard proxies (or even custom proxies that end-users provide) instead of forcing people to use the `ChugSplashProxy`. One way to do this is to upgrade the contract to a temporary implementation that has the `setStorage` function so we can basically treat any proxy as a "ChugSplashProxy" even if it isn't really one. Anyway, we can talk about that separately, just was reminded about it while reading this.

smartcontracts:
Great work here, left some minor comments

smartcontracts (approved):
Whoops never mind my last comment,  I misread part of the code. Approved!

PR #4, opened 2022-07-19
feat(tasks): hh tasks for creating, approving, and listing projects

sam-goldman:
Adds hardhat tasks for:
- Creating a project
- Listing out all projects
- Creating a proposal
- Approving a proposal
- Listing out all proposals for a given project


packages/plugins/src/hardhat/tasks.ts
```diff
@@ -30,8 +32,16 @@ const TASK_CHUGSPLASH_BUNDLE_LOCAL = 'chugsplash-bundle-local'
...
+const TASK_CHUGSPLASH_INIT = 'chugsplash-init'
+const TASK_CHUGSPLASH_LIST_ALL_PROJECTS = 'chugsplash-list-all-projects'
 const TASK_CHUGSPLASH_VERIFY = 'chugsplash-verify'
 const TASK_CHUGSPLASH_COMMIT = 'chugsplash-commit'
+const TASK_CHUGSPLASH_PROPOSE = 'chugsplash-propose'
+const TASK_CHUGSPLASH_APPROVE = 'chugsplash-approve'
+const TASK_CHUGSPLASH_LIST_PROPOSALS = 'chugsplash-list-proposals'
+
+// This address was generated using Create2 so it shouldn't need to be changed for now
+const CHUGSPLASH_REGISTRY_ADDRESS = '0x236151FE7938Fc16600a894E8c40C2E862503D3b'
```
smartcontracts, line 44: We can probably consider exporting this in `@chugsplash/contracts` so the address automatically updates if we when we change the source during development

packages/plugins/src/hardhat/tasks.ts
```diff
@@ -148,6 +158,177 @@ subtask(TASK_CHUGSPLASH_FETCH)
     }
   )
 
+// TODO: Change provider to a non-local RPC provider after demo
```
smartcontracts, line 161: I think best practice here is to use `hre.provider` (or wherever the provider object lives on the config). Then users can run this task with `npx hardhat --network <my network> chugsplash-init`.

packages/plugins/src/hardhat/tasks.ts
```diff
@@ -148,6 +158,177 @@ subtask(TASK_CHUGSPLASH_FETCH)
...
+        {
+          deployConfig: args.deployConfig,
+        }
+      )
+
+      await ChugSplashRegistry.proposeChugSplashBundle(
+        config.options.name,
+        args.bundleHash,
+        bundle.actions.length,
+        args.deployConfig
```
smartcontracts, line 254: Last argument here should be a configUri, I would recommend just sub-calling into `chugsplash-commit` to generate the bundle hash and config URI, then using those as inputs to here

packages/plugins/src/hardhat/tasks.ts
```diff
@@ -148,6 +158,177 @@ subtask(TASK_CHUGSPLASH_FETCH)
...
+    events.forEach((event) =>
+      console.log(
+        `Project: ${event.args.nonIndexedProjectName}\t\tManager: ${event.args.manager}`
+      )
+    )
+  })
+
+task(TASK_CHUGSPLASH_PROPOSE)
+  .setDescription('Proposes a new ChugSplash bundle')
+  .addParam('bundleHash', 'hash of the bundle')
```
smartcontracts, line 220: Probably easier if we just call into `chugsplash-commit` to automatically generate this. Simplifies the UX a bit and it just becomes a single `chugsplash-propose` step instead of having to run `chugsplash-commit` first.

packages/plugins/src/hardhat/tasks.ts
```diff
@@ -148,6 +158,177 @@ subtask(TASK_CHUGSPLASH_FETCH)
...
+        args.deployConfig
+      )
+      spinner.succeed('Bundle successfully proposed')
+    }
+  )
+
+task(TASK_CHUGSPLASH_APPROVE)
+  .setDescription('Allows a manager to approve a bundle to be executed.')
+  .addParam('bundleHash', 'hash of the bundle')
+  .addParam('deployConfig', 'path to chugsplash deploy config')
```
smartcontracts, line 263: Slightly cleaner if we ask the user to provide `project` as an input instead of using `deployConfig` path to resolve the config and then load the name. Important if we decide to remove the name from the config.

packages/plugins/src/hardhat/tasks.ts
```diff
@@ -148,6 +158,177 @@ subtask(TASK_CHUGSPLASH_FETCH)
...
+
+      await ChugSplashRegistry.approveChugSplashBundle(
+        config.options.name,
+        args.bundleHash
+      )
+      spinner.succeed('Bundle successfully approved')
+    }
+  )
+
+// TODO: Exclude listing bundles that have been executed
```
smartcontracts, line 293: Alternatively, maybe we can do a `--include-executed` boolean flag

packages/plugins/src/hardhat/tasks.ts
```diff
@@ -222,12 +403,14 @@ task(TASK_CHUGSPLASH_COMMIT)
 
 task(TASK_CHUGSPLASH_VERIFY)
   .setDescription('Checks if a deployment config matches a bundle hash')
-  .addParam('configUri', 'location of the config file')
+  // TODO: Should `deployConfig` be changed to `configUri` so that deployConfig
+  // has a consistent meaning across tasks?
+  .addParam('deployConfig', 'location of the config file')
```
smartcontracts, line 408: Yeah `deployConfig` should always be a path to the config file, while `configUri` is the remote config URI

smartcontracts (commented):
Looking good! Left a first round of review

packages/plugins/src/hardhat/tasks.ts
sam-goldman: Yeah, I'll leave it as-is in this PR and one of us (or a new contributor!) can tackle the separate issue that you created

packages/plugins/src/hardhat/tasks.ts
```diff
@@ -148,6 +158,177 @@ subtask(TASK_CHUGSPLASH_FETCH)
...
+      })
+
+      const provider = new providers.JsonRpcProvider()
+      const ChugSplashRegistry = new Contract(
+        CHUGSPLASH_REGISTRY_ADDRESS,
+        ChugSplashRegistryABI,
+        await provider.getSigner()
+      )
+
+      await ChugSplashRegistry.register(
```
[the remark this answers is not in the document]
sam-goldman: I agree that we should separate the config into multiple files. Maybe we should discuss this and leave it for a future PR?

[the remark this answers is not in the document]
smartcontracts: Agree, leave it for a future PR

packages/plugins/src/hardhat/tasks.ts
sam-goldman: I updated this task so that it displays bundles as either proposed, approved, or executed (the latter not visible by default)

packages/plugins/src/hardhat/tasks.ts
sam-goldman: Fixed this; needed to add @nomiclabs/hardhat-ethers as a dependency

packages/plugins/src/hardhat/tasks.ts
smartcontracts: Ok perfect

packages/plugins/src/hardhat/tasks.ts
smartcontracts: I'll approve for now, I might hop in here to figure out how to remove this dependency. I don't wanna have to make downstream projects necessarily use this dep, I assume there are alternatives

smartcontracts (approved):

PR #14, opened 2022-07-24
feat(ct): bring back Manager

smartcontracts:
Brings back the ChugSplashManager contract. I realized that there's a
compelling argument for a separation of ownership. We can still get the
best of both worlds with the announceEvent pattern that makes it easy to
find events on the central registry.

### DIFF
--- a/packages/contracts/contracts/ChugSplashManager.sol
+++ b/packages/contracts/contracts/ChugSplashManager.sol
@@ -1,0 +1,311 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.9;
+
+import { Owned } from "@rari-capital/solmate/src/auth/Owned.sol";
+import { ChugSplashRegistry } from "./ChugSplashRegistry.sol";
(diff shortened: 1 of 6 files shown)


packages/contracts/contracts/ChugSplashManager.sol
```diff
@@ -0,0 +1,312 @@
...
+    /**
+     * @notice Emitted when a ChugSplash bundle is approved.
+     *
+     * @param bundleId ID of the bundle being approved.
+     */
+    event ChugSplashBundleApproved(bytes32 indexed bundleId);
+
+    /**
+     * Emitted when a ChugSplash action is executed.
+     * @notice Emitted when a ChugSplash action is executed.
```
sam-goldman, line 74: Looks like this was commented twice

sam-goldman:
Looking good!!

packages/contracts/contracts/ChugSplashManager.sol
smartcontracts: Oops

sam-goldman (approved):

PR #15, opened 2022-07-24
fix(ci): always run CI

smartcontracts:
For now, always run CI on every PR.


smartcontracts:
Force merging this to fix CI

PR #17, opened 2022-07-25
fix(ci): merge not squash on main

smartcontracts:
Fixes mergify.yml to merge, not squash, when merging into main.


smartcontracts:
Force merging so we can do a release

PR #25, opened 2022-08-09
feat(ct): manager supports adapters and proxy admin

sam-goldman:
This PR removes the `ChugSplashProxy` logic in `ChugSplashManager` and replaces it with logic to support non-standard proxy types.


packages/contracts/contracts/ChugSplashRegistry.sol
```diff
@@ -53,23 +67,48 @@ contract ChugSplashRegistry {
...
+     * @param _proxyUpdater Address of the ProxyUpdater for this project. Projects that do not use a custom
+     *                      proxy should set this to the zero-address.
+     * @param _proxyAdmin   Address of the ProxyAdmin for this project. Projects that do not use a custom
+     *                      proxy should set this to the zero-address.
      */
-    function register(string memory _name, address _owner) public {
+    function register(
+        string memory _name,
+        address _owner,
+        address _proxyUpdater,
```
smartcontracts, line 85: We can use the same `ProxyUpdater` implementation for every contract. I would recommend making this an `immutable` that either gets passed into the constructor of the `ChugSplashRegistry`. We then don't need users to pass the value in here themselves. We can then either also pass it into the constructor of the `ChugSplashManager` or we can simply call back to the registry from inside the manager. Calling back will be slightly more expensive.

packages/contracts/contracts/ChugSplashRegistry.sol
```diff
@@ -53,23 +67,48 @@ contract ChugSplashRegistry {
...
+        address _proxyAdmin
```
smartcontracts, line 86: I think the `ProxyAdmin` contract can just be created within the `register` function to avoid the need for a user to create a second contract.

packages/contracts/contracts/ChugSplashManager.sol
```diff
@@ -241,32 +292,81 @@ contract ChugSplashManager is Owned {
...
+                // TODO: Make sure this cannot happen in any case other than OOG.
+                require(
+                    address(created) != address(proxy),
+                    "ChugSplashManager: Proxy was not created correctly"
+                );
+            }
+
+            // Get the address of the proxy's current implementation in case we need to set the
+            // implementation back to its original value later.
+            address implementation = proxy.implementation();
```
smartcontracts, line 322: Looking at this now, I think we can actually mostly merge the logic here with the logic for non-standard contracts. Have everything be owned by the `ProxyAdmin` and use the `ProxyAdmin` as the standard interface for interacting with these proxies. Slightly more expensive but it simplifies the code. For now I think it's worth trading off gas efficiency for legibility.

smartcontracts (approved):

smartcontracts:
I have a few minor nits but we can address them later, nothing blocking. Just gonna approve. Nice work!

PR #26, opened 2022-08-11
feat(ct): add proxy updater

sam-goldman:
Adds the ProxyUpdater contract, which contains the logic for the setCode/setStorage actions.


packages/contracts/contracts/ProxyUpdater.sol
```diff
@@ -0,0 +1,116 @@
...
+     *         from the standard proxy scheme where one would typically deploy the code separately
+     *         and then set the implementation address. We're doing it this way because it gives
+     *         us a lot more freedom on the client side.
+     *
+     * @param _implementationKey The key of the proxy's implementation address. For EIP-1967
+     *                           proxies, this value is:
+     *                           `bytes32(uint256(keccak256('eip1967.proxy.implementation')) - 1)`.
+     * @param _code              New contract code to deploy.
+     */
+    function setCode(bytes32 _implementationKey, bytes memory _code) external {
```
smartcontracts, line 32: This contract may need to be made non-reentrant or possible `owned` if we add support for constructors where people can run arbitrary code. I think this is fine right now since `DEPLOY_CODE_PREFIX` guarantees that no sub-calls will be executed, but a constructor could in theory call another contract which in turn calls back to `setCode` or `setStorage`.

smartcontracts: We should at least have a TODO here.

smartcontracts (approved):

PR #43, opened 2022-08-18
feat(ct): add `ProxyAdmin`

### LINKED ISSUE #27: Create ProxyAdmin contract
The ProxyAdmin is a contract associated with each ChugSplashManager that owns the various proxies for a given project. The ProxyAdmin delegatecalls into various ProxyAdapter contracts that correspond to different proxy types. Using this pattern, the ProxyAdmin can universally handle all different proxy types as long as the ProxyAdmin is considered the owning address.

We separate out the ProxyAdmin from the ChugSplashManager to simplify the codebase and reduce the mental overhead of understanding the risks associated with ProxyAdapter contracts.

sam-goldman:
fixes #27

### DIFF
--- a/packages/contracts/contracts/ProxyAdmin.sol
+++ b/packages/contracts/contracts/ProxyAdmin.sol
@@ -1,28 +1,130 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.9;
 
+import { ChugSplashRegistry } from "./ChugSplashRegistry.sol";
+import { Proxy } from "@eth-optimism/contracts-bedrock/contracts/universal/Proxy.sol";
+import { ProxyUpdater } from "./ProxyUpdater.sol";
+import { IProxyAdapter } from "./IProxyAdapter.sol";
+
+/**
+ * @title ProxyAdmin
+ * @notice The ProxyAdmin is a contract associated with each ChugSplashManager that owns the various
+ *         proxies for a given project. The ProxyAdmin delegatecalls into various ProxyAdapter
+ *         contracts that correspond to different proxy types. Using this pattern, the ProxyAdmin
+ *         can universally handle all different proxy types as long as the ProxyAdmin is considered
+ *         the owning address of each proxy.
+ */
 contract ProxyAdmin {
-    function getProxyImplementation(address _proxy, bytes32 _proxyType)
-        public
-        returns (address implementation)
-    {}
+    /**
+     * @notice Address of the ChugSplashRegistry.
+     */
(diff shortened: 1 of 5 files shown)


packages/contracts/contracts/ProxyAdmin.sol
```diff
@@ -1,28 +1,169 @@
...
         bytes32 _value
-    ) public {}
+    ) public {
+        // If no proxy type has been set, use the default proxy.
+        if (_proxyType == bytes32(0)) {
+            // Upgrade the proxy's implementation to be the ProxyUpdater, and call `setStorage` on
+            // the proxy.
+            Proxy(_proxy).upgradeToAndCall(
+                proxyUpdater,
+                abi.encodeWithSignature("setStorage(bytes32,bytes32)", _key, _value)
```
smartcontracts, line 105: I think it would be worth adding an interface for `IProxyAdapter`. Will simplify the code quite a bit by making it possible to do `abi.encodeCall(IProxyAdapter.setStorage, _key, _value)` which will also give you type safety.

sam-goldman:
fixes #27

packages/contracts/contracts/ProxyAdmin.sol
```diff
@@ -1,28 +1,172 @@
...
     function setProxyCode(
-        address _proxy,
+        address payable _proxy,
         bytes32 _proxyType,
-        bytes memory _data
-    ) public {}
+        bytes memory _code
+    ) public {
+        // If no proxy type has been set, use the default proxy.
+        if (_proxyType == bytes32(0)) {
```
smartcontracts, line 58: Although it's slightly less efficient, I think it would be worth (from a code readability standpoint) to have everything go through the adapter system, including the standard proxy. We get to remove the `if` statement here and have everything under a single code path. Doing this will also mean that we get an example ProxyAdapter implementation for the standard proxy, which is a nice bonus.

packages/contracts/contracts/ProxyAdmin.sol
```diff
@@ -1,28 +1,172 @@
...
     function setProxyStorage(
-        address _proxy,
+        address payable _proxy,
         bytes32 _proxyType,
         bytes32 _key,
         bytes32 _value
-    ) public {}
+    ) public {
+        // If no proxy type has been set, use the default proxy.
+        if (_proxyType == bytes32(0)) {
```
smartcontracts, line 102: Same comment as above re: combining the two codepaths.

smartcontracts (approved):
Nice!

PR #45, opened 2022-08-20
feat(ct): add executor selection to Manager

### LINKED ISSUE #31: Add executor selection to Manager
The ChugSplashManager should only allow actions to be executed by an address that is selected by the [Executor Selection Strategy](https://github.com/smartcontracts/chugsplash/blob/develop/specs/executor-economics.md#executor-selection-strategies) contract for the given project. The Executor Selection Strategy contract must NOT be allowed to change after a deployment is approved.

sam-goldman:
Fixes #31

### DIFF
--- a/packages/contracts/contracts/ChugSplashManager.sol
+++ b/packages/contracts/contracts/ChugSplashManager.sol
@@ -8,6 +8,7 @@ import { ProxyAdmin } from "./ProxyAdmin.sol";
 import { ProxyUpdater } from "./ProxyUpdater.sol";
 import { Create2 } from "./libraries/Create2.sol";
 import { MerkleTree } from "./libraries/MerkleTree.sol";
+import { IExecutorSelectionStrategy } from "./IExecutorSelectionStrategy.sol";
 
 /**
  * @title ChugSplashManager
@@ -136,6 +137,11 @@ contract ChugSplashManager is Owned {
      * @notice ID of the currently active bundle.
      */
     bytes32 public activeBundleId;
+
+    /**
+     * @notice The address of the Executor Selection Strategy for this project.
+     */
+    IExecutorSelectionStrategy public executorSelectionStrategy;
 
     /**
      * @notice Mapping of bundle IDs to bundle state.
@@ -273,6 +279,12 @@ contract ChugSplashManager is Owned {
             "ChugSplashManager: action has already been executed"
         );
 
+        address executor = getSelectedExecutor(activeBundleId);
+        require(
+            executor == msg.sender,
+            "ChugSplashManager: caller is not approved executor for active bundle ID"
+        );
+
         require(
             MerkleTree.verify(
                 activeBundleId,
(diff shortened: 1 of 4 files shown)


packages/contracts/contracts/ChugSplashManager.sol
```diff
@@ -145,6 +146,11 @@ contract ChugSplashManager is Owned {
      */
     bytes32 public activeBundleId;
 
+    /**
+     * @notice The address of the Executor Selection Strategy for this project.
+     */
+    address internal executorSelectionStrategy;
```
smartcontracts, line 152: Minor modification, prefer that this be `IExecutorSelectionStrategy public ess`. Gives better type safety and means we don't need to cast stuff in other places.

packages/contracts/contracts/IExecutorSelectionStrategy.sol
```diff
@@ -0,0 +1,21 @@
...
+ */
+interface IExecutorSelectionStrategy {
+    /**
+     * @notice Queries the selected executor for a given project/bundle.
+     *
+     * @param _project  Address of the ChugSplashManager that mananges the project.
+     * @param _bundleId ID of the bundle currently being executed.
+     *
+     * @return Address of the selected executor.
+     * @return Time when the executor was selected.
```
smartcontracts, line 15: Actually let's get rid of the time here. It adds unnecessary complexity. We should just pass the time since the bundle was approved.

packages/contracts/contracts/ChugSplashManager.sol
```diff
@@ -364,6 +376,37 @@ contract ChugSplashManager is Owned {
...
+    }
+
+    /**
+     * @notice Get the address of the executor selected by the Executor Selection Strategy to
+     *         execute a given bundle ID.
+     *
+     * @param _bundleId Bundle ID.
+     */
+    function getSelectedExecutor(bytes32 _bundleId) public view returns (address) {
+        address executor = executorSelectionStrategy.getSelectedExecutor(
```
smartcontracts, line 403: Minor simplification, can just return the result of the function directly at this point

packages/contracts/contracts/ChugSplashManager.sol
```diff
@@ -364,6 +376,37 @@ contract ChugSplashManager is Owned {
...
+     */
+    function setExecutorSelectionStategy(IExecutorSelectionStrategy _executorSelectionStrategy)
+        external
+        onlyOwner
+    {
+        require(
+            activeBundleId == bytes32(0),
+            "ChugSplashManager: a bundle has been approved and not yet completed"
+        );
+        // TODO: Add an EIP-165 check for the ESS address here?
```
smartcontracts, line 392: Eh, let's leave it to the UI to handle

smartcontracts:
Looks like this just needs a lint fix and should be ready to go!

smartcontracts (approved):

PR #46, opened 2022-08-20
feat(ct): add simple lock ESS

### LINKED ISSUE #33: Create simple lock-based selection module
Once we add Executor Selection Strategies to the ChugSplashManager, we will need to implement a simple ESS module. I recommend that we implement the [Simple Lock Strategy](https://github.com/smartcontracts/chugsplash/blob/develop/specs/executor-economics.md#ess-1-simple-lock) as described in the specs.

sam-goldman:
Fixes #33


packages/contracts/contracts/IExecutorSelectionStrategy.sol
```diff
@@ -1,10 +1,59 @@
...
+        address indexed executor
+    );
+
+    /**
+     * @notice Allows an executor to claim the sole right to execute actions for an upgrade.
+     *
+     * @param _project  Address of the ChugSplashManager for the project being claimed.
+     * @param _bundleId ID of the bundle being claimed.
+     */
+    function claim(address _project, bytes32 _bundleId) external payable;
```
smartcontracts, line 46: I would make these functions specific to the `SimpleLockESS` instead of adding them to the `IExecutorSelectionStrategy`, since these aren't going to be necessary for every ESS.

sam-goldman: We use `claim` in the executor script [here](https://github.com/smartcontracts/chugsplash/blob/744545dfcfa92de0955b5cd811e87d815ad6486b/packages/executor/src/index.ts#L151), `getSelectedExecutor` in the ChugSplashManager [here](https://github.com/smartcontracts/chugsplash/blob/82af0f4eaacc7f7d61553c19f31fdbd6eb57f395/packages/contracts/contracts/ChugSplashManager.sol#L312), and `returnExecutorBond` is added to the ChugSplashManager in #50 [here](https://github.com/smartcontracts/chugsplash/blob/82af0f4eaacc7f7d61553c19f31fdbd6eb57f395/packages/contracts/contracts/ChugSplashManager.sol#L393) and [here](https://github.com/smartcontracts/chugsplash/blob/82af0f4eaacc7f7d61553c19f31fdbd6eb57f395/packages/contracts/contracts/ChugSplashManager.sol#L416). I think these functions should be kept in the interface unless we remove them from the Manager contract / executor script.

packages/contracts/contracts/SimpleLockESS.sol
```diff
@@ -0,0 +1,144 @@
...
+        address prevSelectedExecutor = upgrade.selectedExecutor;
+
+        upgrade.timeClaimed = block.timestamp;
+        upgrade.selectedExecutor = msg.sender;
+
+        if (prevSelectedExecutor != address(0)) {
+            // If the previously selected executor failed to complete the upgrade, their bond is
+            // sent to the address of the ChugSplashManager.
+            // TODO: Does it make more sense to send the ETH to the owner of the
+            // ChugSplashManager (i.e. the project owner)?
```
sam-goldman, line 97: @smartcontracts What do you think?

packages/contracts/contracts/IExecutorSelectionStrategy.sol
smartcontracts: I need to think about this a little more. I'm considering whether it might be worth scrapping the ESS/EPS adapter system for the alpha release. If we're making ChugSplash upgradeable, we could always add it back in later.

smartcontracts: Unfortunately `SimpleLock` does mean a lot of value goes to miners in a PGA. Maybe not worth worrying about for the alpha release, but we may need a simple staking mechanism eventually.

smartcontracts: Simple staking mechanism:
- Post at least 1 ETH and you get to be an executor
- Chance of being selected to execute a job depends on the % of the ETH pool that staked, time-weighted over the last month or something
- Slots for starting an upgrade are short (on the order of a few minutes) so that an upgrade will start within X amount of time on average assuming that Y% of the staking supply is honest
- Once you start you have Z blocks to finish where Z depends on the number of actions within the upgrade or something
- Since some upgrades might not be executable (let's say the thing returned by the IPFS link is just garbage data), you can't punish people for refusing to execute the upgrade (would make a really cool usecase for optimistic fault proofs down the line though!)

packages/contracts/contracts/SimpleLockESS.sol
smartcontracts: Nah sending to the `ChugSplashManager` itself is good enough. Relevant to my comment in #50 where we should just have the `ChugSplashManager` be treated kind of like a piggy bank that holds the funds for everything.

smartcontracts: I also like the idea of seeing a UI that represents changes in balance of the `ChugSplashManager` with an entry like "Account received a fee because this upgrade wasn't completed on time" :wink:

packages/contracts/contracts/IExecutorSelectionStrategy.sol
sam-goldman: Sounds like a good idea to add a staking mechanism like that in the future. Should we scrap the ESS/EPS system for the alpha release or leave it as is?

smartcontracts: Let's scrap both EPS and ESS in favor of simple lock for executor selection and flat fee (something like 10% above gas usage) for executor payment by default for the Alpha release.

sam-goldman: Sounds good, removed the ESS logic and replaced it with an `ExecutorSelector` contract.

packages/contracts/contracts/ChugSplashManager.sol
```diff
@@ -373,6 +423,73 @@ contract ChugSplashManager is Owned {
...
+            block.timestamp > executionDeadline,
+            "ChugSplashManager: bundle already claimed by an executor"
+        );
+
+        address prevSelectedExecutor = bundle.selectedExecutor;
+
+        bundle.timeClaimed = block.timestamp;
+        bundle.selectedExecutor = msg.sender;
+
+        emit BundleClaimed(_bundleId, msg.sender, prevSelectedExecutor);
```
smartcontracts, line 460: I think we need to propagate this by sending to `announce` on the `ChugSplashRegistry`

packages/contracts/contracts/ChugSplashManager.sol
```diff
@@ -373,6 +423,73 @@ contract ChugSplashManager is Owned {
...
+
+        bundle.executorBondReturned = true;
+        // Set the other parameters back to their default values.
+        bundle.timeClaimed = 0;
+        bundle.selectedExecutor = address(0);
+
+        (bool success, ) = payable(executor).call{ value: executorBondAmount }(new bytes(0));
+        require(success, "ChugSplashManager: call to executor failed");
+
+        emit ExecutorBondReturned(_bundleId, executor);
```
smartcontracts, line 490: Same as above, need to `announce`

packages/contracts/contracts/ChugSplashManager.sol
```diff
@@ -373,6 +423,73 @@ contract ChugSplashManager is Owned {
...
+        );
+        require(
+            !bundle.executorBondReturned,
+            "ChugSplashManager: bond already returned to executor"
+        );
+
+        address executor = bundle.selectedExecutor;
+
+        bundle.executorBondReturned = true;
+        // Set the other parameters back to their default values.
```
smartcontracts, line 483: Hmmm why reset these?

packages/contracts/contracts/ChugSplashManager.sol
```diff
@@ -373,6 +423,73 @@ contract ChugSplashManager is Owned {
...
+        bundle.timeClaimed = 0;
+        bundle.selectedExecutor = address(0);
+
+        (bool success, ) = payable(executor).call{ value: executorBondAmount }(new bytes(0));
```
smartcontracts, line 487: I need to consider whether this could be exploited in any way. I don't think it could, but maybe there's a safer way of doing this transfer without triggering the executor's code.

packages/contracts/contracts/ChugSplashManager.sol
```diff
@@ -373,6 +423,73 @@ contract ChugSplashManager is Owned {
...
+
+    /**
+     * @notice Refunds `executorBondAmount` to the executor if they complete the bundle within
+     *        `executionLockTime`, or if the project owner cancels the bundle.
+     *
+     * @param _bundleId ID of the bundle that was completed or cancelled by the project owner.
+     */
+    function returnExecutorBond(bytes32 _bundleId) external {
+        ClaimedBundle storage bundle = claimedBundles[_bundleId];
+        require(
```
smartcontracts, line 471: We need to require that the bundle is either complete or cancelled

packages/contracts/contracts/ChugSplashManager.sol
```diff
@@ -373,6 +423,73 @@ contract ChugSplashManager is Owned {
...
+            status == ChugSplashBundleStatus.APPROVED,
+            "ChugSplashManager: bundle is not active"
+        );
+
+        ClaimedBundle storage bundle = claimedBundles[_bundleId];
+
+        uint256 executionDeadline = bundle.timeClaimed + executionLockTime;
+        require(
+            block.timestamp > executionDeadline,
```
smartcontracts, line 451: Minor nit but we could just put this on one line instead of defining the variable above

packages/contracts/contracts/ChugSplashManager.sol
```diff
@@ -131,6 +139,28 @@ contract ChugSplashManager is Owned {
...
+     *
+     * @param bundleId     ID of the bundle that was claimed.
+     * @param executor     Address of the executor that claimed the bundle ID for the project.
+     * @param prevExecutor Address of the executor that previously claimed the bundle ID for the
+     *                     project. This is address(0) if there was no previous executor. If there
+     *                     was a previous executor, then this event signifies that the previous
+     *                     executor's bond has been forfeited to the ChugSplashManager, which the
+     *                     project owner can claim with a call to `withdrawETH`.
+     */
+    event BundleClaimed(bytes32 indexed bundleId, address indexed executor, address prevExecutor);
```
smartcontracts, line 153: I think we can get rid of the `prevExecutor` field and just use the existence of more than one `BundleClaimed` event to signify that the bundle wasn't executed in time. Slightly cleans up the code too.

packages/contracts/contracts/ChugSplashManager.sol
```diff
@@ -50,6 +49,15 @@ contract ChugSplashManager is Owned {
         uint256 total;
     }
 
+    /**
+     * @notice Struct representing a bundle claimed by an executor.
+     */
+    struct ClaimedBundle {
```
smartcontracts, line 55: I think this could be merged into the `ChugSplashBundleState` instead of having a new struct.

smartcontracts: Avoids introducing a new mapping too.

packages/contracts/contracts/ChugSplashManager.sol
```diff
@@ -158,9 +188,17 @@ contract ChugSplashManager is Owned {
     bytes32 public activeBundleId;
 
     /**
-     * @notice The address of the Executor Selection Strategy for this project.
+     * @notice Amount in ETH that the executor must send to this contract to claim a bundle for
+     *         `executionLockTime`.
+     */
+    uint256 public immutable executorBondAmount;
```
smartcontracts, line 194: Should be modifiable (but not during an active upgrade).

packages/contracts/contracts/ChugSplashManager.sol
sam-goldman: Yeah actually not necessary

packages/contracts/contracts/ChugSplashManager.sol
sam-goldman: If you think it might be unsafe, we could add a `withdrawExecutorBond` function that can be called by the executor when an upgrade is completed

smartcontracts (approved):
Woot! Let's merge this baby

PR #50, opened 2022-08-28
feat(ct): add logic for owner and executor bonds in Manager

sam-goldman:
Adds logic to the ChugSplashManager to handle executor and project owner bonds. The ChugSplashManager calls the project's ESS contract to refund the executor's bond when a bundle is completed or when a project owner cancels the active  bundle. Project owners are sent their bond upon successful completion of a bundle.


packages/contracts/contracts/ChugSplashManager.sol
```diff
@@ -270,8 +300,8 @@ contract ChugSplashManager is Owned {
         ChugSplashBundleState storage bundle = bundles[activeBundleId];
 
         require(
-            bundle.status != ChugSplashBundleStatus.COMPLETED,
-            "ChugSplashManager: bundle has already been completed"
+            bundle.status == ChugSplashBundleStatus.APPROVED,
```
sam-goldman, line 303: Changed this from `!= ChugSplashBundleStatus.COMPLETED` to `== ChugSplashBundleStatus.APPROVED` because there's now a `CANCELLED` attribute on `ChugSplashBundleStatus`.

packages/contracts/contracts/ChugSplashManager.sol
```diff
@@ -349,13 +379,45 @@ contract ChugSplashManager is Owned {
         // completed and reset the active bundle hash so that a new bundle can be executed.
         if (bundle.total == bundle.executions.length) {
             bundle.status = ChugSplashBundleStatus.COMPLETED;
+            bytes32 completedBundleId = activeBundleId;
             activeBundleId = bytes32(0);
 
-            emit ChugSplashBundleCompleted(activeBundleId, msg.sender, bundle.total);
+            emit ChugSplashBundleCompleted(completedBundleId, msg.sender, bundle.total);
```
sam-goldman, line 385: Changed the event argument from `activeBundleId` to `completedBundleId` because `activeBundleId` is always bytes32(0) here.

packages/contracts/contracts/ChugSplashManager.sol
```diff
@@ -223,12 +248,18 @@ contract ChugSplashManager is Owned {
...
-     *         executed as soon as the bundle is approved.
+     * @notice Allows the owner to approve a bundle to be executed. Owners must submit a bond of
+     *         `ownerBondAmount`, which will be forfeited if they cancel an active bundle. Note that
+     *          the bundle can be executed as soon as the bundle is approved.
      *
      * @param _bundleId ID of the bundle to approve
      */
-    function approveChugSplashBundle(bytes32 _bundleId) public onlyOwner {
+    function approveChugSplashBundle(bytes32 _bundleId) public payable onlyOwner {
+        require(
```
smartcontracts, line 258: We may actually be able to make this slightly cleaner (and separate concerns better) by doing the following:
- Add a function `depositETH` that allows anyone to send funds to the contract
- Add a function `withdrawETH` that allows the owner to withdraw funds to any recipient address
- Funds cannot be withdrawn during an active upgrade
- Funds are used to pay executors
- We can simply check that `balance(this) >= BOND_AMOUNT` when calling `approveChugSplashBundle`, no need for the owner to also hold ETH (some separate account can be responsible for that)

packages/contracts/contracts/ChugSplashManager.sol
```diff
@@ -349,13 +380,45 @@ contract ChugSplashManager is Owned {
...
+        bytes32 cancelledBundleId = activeBundleId;
+        activeBundleId = bytes32(0);
+        bundle.status = ChugSplashBundleStatus.CANCELLED;
+
+        emit ChugSplashBundleCancelled(cancelledBundleId, msg.sender, bundle.total);
+        registry.announce("ChugSplashBundleCancelled");
+
+        executorSelectionStrategy.returnExecutorBond(cancelledBundleId);
+
+        // TODO: Who should receive the project owner's bond? The executor?
```
smartcontracts, line 419: Executor should receive the bond. We mainly want the bond to prevent the project's owner from trolling executors by starting upgrades, waiting for them to start submitting big transactions, then cancelling and forcing the executors to waste gas on reverting transactions where they can't get paid.

packages/contracts/contracts/ChugSplashManager.sol
smartcontracts: Kills two birds with one stone. We'll also later have to put some logic in the executor to simulate the transactions, figure out how much gas the whole upgrade would require, and check if the account has sufficient ETH to be worth claiming the upgrade.

sam-goldman: > We may actually be able to make this slightly cleaner (and separate concerns better) by doing the following:
> 
> * Add a function `depositETH` that allows anyone to send funds to the contract
> * Add a function `withdrawETH` that allows the owner to withdraw funds to any recipient address
> * Funds cannot be withdrawn during an active upgrade
> * Funds are used to pay executors
> * We can simply check that `balance(this) >= BOND_AMOUNT` when calling `approveChugSplashBundle`, no need for the owner to also hold ETH (some separate account can be responsible for that)

Sounds good. Added this except for the second to last bullet point, which seems out of scope for this PR since it involves executor logic in the ChugSplashManager.

packages/contracts/contracts/ChugSplashManager.sol
```diff
@@ -348,13 +383,44 @@ contract ChugSplashManager is Owned {
...
+            bytes32 completedBundleId = activeBundleId;
             activeBundleId = bytes32(0);
 
-            emit ChugSplashBundleCompleted(activeBundleId, msg.sender, bundle.total);
+            emit ChugSplashBundleCompleted(completedBundleId, msg.sender, bundle.total);
             registry.announce("ChugSplashBundleCompleted");
+
+            // Refund the executor's bond
+            executorSelectionStrategy.returnExecutorBond(completedBundleId);
         }
```
sam-goldman, line 394: I left out the call that refunds the owner's bond after a bundle is completed. That way, owners can reuse bonds between bundles.

packages/contracts/contracts/ChugSplashManager.sol
```diff
@@ -450,39 +520,6 @@ contract ChugSplashManager is Owned {
         registry.announce("ChugSplashBundleClaimed");
     }
 
-    /**
-     * @notice Refunds `executorBondAmount` to the executor if they complete the bundle within
-     *        `executionLockTime`, or if the project owner cancels the bundle.
-     *
-     * @param _bundleId ID of the bundle that was completed or cancelled by the project owner.
-     */
-    function returnExecutorBond(bytes32 _bundleId) external {
```
sam-goldman, line 459: Changed this into an internal function since it's only called by the ChugSplashManager

smartcontracts: I think we should do a pull-based model instead of a push-based model, so I did like having this be public. Did you see issues with having it public?

packages/contracts/contracts/ChugSplashManager.sol
smartcontracts: Yep that's perfect. We can also easily represent this in our docs with something like "Because users could grief executors by cancelling upgrades shortly after starting them, we require that you have a minimum balance of X ETH in your Manager contract before you can start an upgrade. This balance acts as a deposit that you will forfeit if you need to cancel an active upgrade."

packages/contracts/contracts/ChugSplashManager.sol
```diff
@@ -408,13 +452,44 @@ contract ChugSplashManager is Owned {
...
+        bundle.status = ChugSplashBundleStatus.CANCELLED;
+
+        emit ChugSplashBundleCancelled(cancelledBundleId, msg.sender, bundle.total);
+        registry.announce("ChugSplashBundleCancelled");
+
+        _returnExecutorBond(cancelledBundleId);
+
+        // Send the project owner's bond to the executor
+        address executor = getSelectedExecutor(activeBundleId);
+        (bool success, ) = payable(executor).call{ value: ownerBondAmount }(new bytes(0));
```
smartcontracts, line 489: An interesting issue here is that the executor could prevent a user from cancelling an active bundle if the executor is a contract address that reverts if called by the `Manager` contract. We'll have to use a pull-based model here instead of push-based.

packages/contracts/contracts/ChugSplashManager.sol
```diff
@@ -408,13 +452,44 @@ contract ChugSplashManager is Owned {
...
+    function cancelActiveChugSplashBundle() public onlyOwner {
+        require(activeBundleId != bytes32(0), "ChugSplashManager: no bundle is currently active");
+
+        ChugSplashBundleState storage bundle = bundles[activeBundleId];
+
+        bytes32 cancelledBundleId = activeBundleId;
+        activeBundleId = bytes32(0);
+        bundle.status = ChugSplashBundleStatus.CANCELLED;
+
+        emit ChugSplashBundleCancelled(cancelledBundleId, msg.sender, bundle.total);
```
smartcontracts, line 482: Probably not necessary to emit the `msg.sender` but emitting `bundle.total` is a great idea.

smartcontracts (approved):
I'm going to approve this, but then I'd like to start fleshing out the `ChugSplashManager` speed spec and get a test suite up so we know exactly what behavior we need before we start implementing anything else: https://github.com/smartcontracts/chugsplash/pull/60/files

PR #51, opened 2022-08-28
fix(ct): allow user to claim proxy ownership

sam-goldman:
This adds logic to the `ChugSplashManager` and `ProxyAdmin` to allow a project owner to claim ownership of the proxies used in a project. This includes non-standard proxies as well as default proxies.

### DIFF
--- a/packages/contracts/contracts/ProxyAdmin.sol
+++ b/packages/contracts/contracts/ProxyAdmin.sol
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.9;
 
+import { Owned } from "@rari-capital/solmate/src/auth/Owned.sol";
 import { ChugSplashRegistry } from "./ChugSplashRegistry.sol";
 import { Proxy } from "@eth-optimism/contracts-bedrock/contracts/universal/Proxy.sol";
 import { ProxyUpdater } from "./ProxyUpdater.sol";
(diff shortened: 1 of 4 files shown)


packages/contracts/contracts/ProxyAdmin.sol
```diff
@@ -35,7 +36,7 @@ contract ProxyAdmin {
      * @param _registry     Address of the ChugSplashRegistry.
      * @param _proxyUpdater Address of the ProxyUpdater.
      */
-    constructor(ChugSplashRegistry _registry, address _proxyUpdater) {
+    constructor(ChugSplashRegistry _registry, address _proxyUpdater) Owned(msg.sender) {
```
smartcontracts, line 39: Owner of this is the `ChugSplashManager`, so I think we need to put a function into the `ChugSplashManager` that calls `transferProxyOwnership`. I'm also thinking now that it might be better to merge the `ChugSplashManager` and `ProxyAdmin` contracts because this ownership stuff starts to get complex (and for good reason). It creates a lot of extra functions in both the `ProxyAdmin` and the `ChugSplashManager`. I'm also coming around to the idea that a malicious `ProxyAdapter` would be able to completely mess with the proxies anyway, so it's not like this separation really gives you any extra security.

sam-goldman: Agreed, made #53 to merge the `ProxyAdmin` into the `ChugSplashManager`. Will close this PR and open a new one that can be merged after #53 is approved.

smartcontracts: Approved #53!

sam-goldman: Sweet! Closed by #55.

PR #54, opened 2022-08-31
feat(ct): add upgradeability to Manager and Registry

sam-goldman:
Adds upgradeability to the `ChugSplashManager` and the `ChugSplashRegistry` to make it easy to upgrade these contracts on a testnet. Each ChugSplashManager shares a single fixed implementation address, which is stored in the (proxied) ChugSplashRegistry as the variable `managerImplementation`. Additionally, there is a root ChugSplashManager, which will have the power to upgrade the ChugSplashRegistry, thereby letting us upgrade all ChugSplashManagers at once (by changing `managerImplementation`).

This PR also adds a `ChugSplashBootLoader`, which is a contract responsible for deploying the contracts mentioned above. Once these contracts are deployed with `ChugSplashBootLoader.ts`, we can use ChugSplash to perform any upgrades we like. (Meta!)


packages/contracts/contracts/ChugSplashRegistry.sol
```diff
@@ -90,12 +102,9 @@ contract ChugSplashRegistry {
             "ChugSplashRegistry: name already registered"
         );
 
-        ChugSplashManager manager = new ChugSplashManager{ salt: bytes32(0) }(
-            this,
-            _name,
-            _owner,
-            proxyUpdater
-        );
+        ChugSplashManager manager = new ChugSplashManager{ salt: bytes32(0) }();
```
smartcontracts, line 105: We need to deploy a `ChugSplashManagerProxy` instead

packages/contracts/contracts/ChugSplashRegistry.sol
```diff
@@ -2,6 +2,7 @@
 pragma solidity ^0.8.9;
 
 import { ChugSplashManager } from "./ChugSplashManager.sol";
+import { Initializable } from "@openzeppelin/contracts/proxy/utils/Initializable.sol";
```
smartcontracts, line 5: For simplicity, I would take `Initializable` from `contracts-upgradeable`

packages/contracts/contracts/ChugSplashManager.sol
```diff
@@ -1,7 +1,10 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.9;
 
-import { Owned } from "@rari-capital/solmate/src/auth/Owned.sol";
+// import { Owned } from "@rari-capital/solmate/src/auth/Owned.sol";
```
smartcontracts, line 4: I say we just get rid of the import entirely. My preference is to stick to either OZ or solmate, since we're using a lot of OZ elsewhere let's do OZ.

packages/contracts/contracts/ChugSplashManager.sol
```diff
@@ -116,12 +119,14 @@ contract ChugSplashManager is Owned {
     /**
      * @notice Address of the ChugSplashRegistry.
      */
-    ChugSplashRegistry public immutable registry;
+    // TODO: Add immutable keyword once this contract is not upgradeable anymore.
```
smartcontracts, line 122: I think we can get rid of the TODOs here, we should just a thorough review when we transition away from upgradeability.

smartcontracts, line 122: Actually now that I'm thinking about it, we can make these immutable if we add a constructor back to this contract since the immutables will then live in the implementation. Just need a constructor that calls the `initialize` function.

packages/contracts/contracts/ChugSplashRegistry.sol
```diff
@@ -69,13 +70,24 @@ contract ChugSplashRegistry {
     /**
      * @notice Address of the ProxyUpdater.
      */
-    address public immutable proxyUpdater;
+    // TODO: Add immutable keyword once this contract is not upgradeable anymore.
```
smartcontracts, line 73: Same note as with `ChugSplashManager`, we can just use a constructor and put the immutables in the implementation contract

packages/contracts/deploy/001-ChugSplashManager.ts
```diff
@@ -0,0 +1,93 @@
...
+    registryProxyArtifact.bytecode
+  }${encodedConstructorArg.slice(2)}`
+  // Get the Proxy address generated by Create2.
+  const registryProxyAddress = hre.ethers.utils.getCreate2Address(
+    '0x4e59b44847b379578588920ca78fbf26c0b4956c', // Deterministic deployment proxy address
+    hre.ethers.utils.solidityKeccak256(['string'], ['Proxy']), // Salt
+    hre.ethers.utils.keccak256(initCode) // Init code hash
+  )
+
+  // Deploy and initialize the ChugSplashManagerProxy:
```
smartcontracts, line 47: Let's deploy the proxy and implementation in two different files. That way when we upgrade the system we aren't re-deploying the proxy.

packages/contracts/deploy/001-ChugSplashManager.ts
```diff
@@ -0,0 +1,93 @@
...
+        managerAddress,
+        deployer,
+        encodedInitializeFunction,
+      ],
+      log: true,
+    }
+  )
+  await deployProxy()
+
+  // Initialize the implementation contract. This won't make any difference during testing, since
```
smartcontracts, line 77: Can remove this if we add that constructor to the implementation

packages/contracts/deploy/001-ChugSplashManager.ts
```diff
@@ -0,0 +1,93 @@
...
+        ['string'],
+        ['ChugSplashManager']
+      ),
+      from: deployer,
+      log: true,
+    }
+  )
+  const { address: managerAddress, abi: managerAbi } = await deployManager()
+
+  // Get the ChugSplashRegistry's proxy address that *will* be deployed using Create2. We'll need
```
smartcontracts, line 24: Lol, funny little bootstrapping problem going on here. Lemme think for a sec...

smartcontracts: Honestly the most fun way to do this would be to use ChugSplash to bootstrap itself, but the `ChugSplashManager` relies on the existence of the `ChugSplashRegistry`. Next best thing might be to have a contract called `ChugSplashBootLoader` that does this deployment instead inside of its constructor. Then our entire initial deployment becomes sending a single transaction, which is very ChugSplashy. 

Once the initial system has been deployed, we can manage ChugSplash with ChugSplash!

packages/contracts/contracts/ChugSplashManagerProxy.sol
```diff
@@ -0,0 +1,33 @@
...
+
+import {
+    TransparentUpgradeableProxy
+} from "@openzeppelin/contracts/proxy/transparent/TransparentUpgradeableProxy.sol";
+import { ChugSplashRegistry } from "./ChugSplashRegistry.sol";
+
+/**
+ * @title ChugSplashManagerProxy
+ */
+contract ChugSplashManagerProxy is TransparentUpgradeableProxy {
```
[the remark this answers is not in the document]
sam-goldman: The standard `Proxy` doesn't have the `virtual` keyword on its `_getImplementation` function, which prevents us from overriding it like we do [here](https://github.com/smartcontracts/chugsplash/blob/026b81ccc19eb98b42f123fc2dc4bba340abad94/packages/contracts/contracts/ChugSplashManagerProxy.sol#L30-L32). That's why I opted for inheriting from `TransparentUpgradeableProxy`

packages/contracts/contracts/ChugSplashManager.sol
sam-goldman: Unfortunately I don't think Solidity lets us do that. I tried to do this same thing when I was making the fixes for the OpenZeppelin audit, since they recommended adding the immutable keyword to variables in some implementation contracts (in L-14). However, this won't work because the variables are assigned outside of the constructor (in an initialize function). Adding the immutable keyword to these variables results in this error from the Solidity compiler: 
```TypeError: Cannot write to immutable here: Immutable variables can only be initialized inline or assigned directly in the constructor```.

File: packages/contracts/contracts/ChugSplashRegistry.sol (hunk 5 above)
Comment (sam-goldman) [reply packages/contracts/contracts/ChugSplashRegistry.sol:73]: Same thing here, it seems like we can't assign immutable variables outside of the constructor.

File: packages/contracts/contracts/ChugSplashManager.sol (hunk 4 above)
Comment (smartcontracts) [reply packages/contracts/contracts/ChugSplashManager.sol:121]: Ah I mean assign the immutables inside the constructor. Only use the constructor for the implementation contract.

File: packages/contracts/contracts/ChugSplashManagerProxy.sol (hunk 9 above)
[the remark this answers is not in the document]
Comment (smartcontracts) [reply packages/contracts/contracts/ChugSplashManagerProxy.sol:12]: Oh erm, we can probably change that upstream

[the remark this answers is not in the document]
Comment (sam-goldman) [reply packages/contracts/contracts/ChugSplashManagerProxy.sol:12]: Ok cool, should we wait for that to be changed in Optimism's repo? Or stick with the TransparentUpgradeableProxy?

File: packages/contracts/contracts/ChugSplashManager.sol (hunk 4 above)
Comment (sam-goldman) [reply packages/contracts/contracts/ChugSplashManager.sol:121]: Yeah, we can do that to initialize the implementation contract, but how will the immutable variables be assigned in the proxy? We can't assign them in the initializer because immutables can only be assigned in the constructor.

File: packages/contracts/deploy/001-ChugSplashManager.ts (hunk 7 above)
Comment (sam-goldman) [reply packages/contracts/deploy/001-ChugSplashManager.ts]: Added a call to the `initialize` function in the constructors. I left the state variables without the `immutable` keyword unless you've got a solution to [this](https://github.com/smartcontracts/chugsplash/pull/54#discussion_r960151869) though

File: packages/contracts/deploy/001-ChugSplashManager.ts (hunk 8 above)
Comment (sam-goldman) [reply packages/contracts/deploy/001-ChugSplashManager.ts]: Good idea! Added the ChugSplashBootLoader and took most of the deploy scripts.

File: packages/contracts/deploy/001-ChugSplashManager.ts (hunk 6 above)
Comment (sam-goldman) [reply packages/contracts/deploy/001-ChugSplashManager.ts]: Removed these deploy scripts in favor of `ChugSplashBootLoader`!

File: packages/contracts/contracts/ChugSplashManager.sol [hunk 10]
```diff
@@ -320,7 +342,7 @@ contract ChugSplashManager is Owned {
                 // standard OOG, then this would halt the entire contract. TODO: Make sure this
                 // cannot happen in any case other than OOG.
                 require(
-                    address(created) != proxy,
+                    address(created) == proxy,
```
sam-goldman, line 345: Changed this check from `!=` to `==` since it seems like the addresses are supposed to match.

packages/contracts/contracts/ChugSplashManager.sol
```diff
@@ -413,7 +435,11 @@ contract ChugSplashManager is Owned {
     function getProxyByName(string memory _name) public view returns (address payable) {
         return (
             payable(
-                Create2.compute(address(this), keccak256(bytes(_name)), type(Proxy).creationCode)
+                Create2.compute(
+                    address(this),
+                    keccak256(bytes(_name)),
+                    abi.encodePacked(type(Proxy).creationCode, abi.encode(address(this)))
```
sam-goldman, line 441: Changed the `_bytecode` argument in `Create2` from `type(Proxy).creationCode` to `abi.encodePacked(type(Proxy).creationCode, abi.encode(address(this)))` since this argument is supposed to be the creation code concatenated with the ABI-encoded constructor arguments ([source](https://docs.soliditylang.org/en/latest/control-structures.html?highlight=create2#salted-contract-creations-create2))

packages/contracts/hardhat.config.ts
```diff
@@ -14,6 +14,9 @@ const config: HardhatUserConfig = {
     version: '0.8.15',
   },
   networks: {
+    hardhat: {
+      allowUnlimitedContractSize: true,
+    },
```
sam-goldman, line 19: I had to set this because the `ChugSplashBootLoader` contract is huge... gonna need to get it to compile somehow.

packages/contracts/contracts/ChugSplashManager.sol
smartcontracts: Ah the proxy gets the immutables from the implementation contract. I believe it should just work.

packages/contracts/contracts/ChugSplashManagerProxy.sol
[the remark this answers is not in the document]
smartcontracts: Stick with TransparenyUpgradeableProxy for now. I'll make the upstream change next week.

packages/contracts/contracts/ChugSplashManager.sol
smartcontracts: I believe this must work or constants wouldn't work inside of implementations. Maybe worth checking briefly in Remix or something but 99% sure this should work out of the box.

sam-goldman: Ahh I see. Fixed!

PR #58, opened 2022-09-07
fix(ct): remove custom proxy type use

smartcontracts:
Removes usage of the custom proxy type system in the ChugSplashManager. We will reintroduce this functionality later, but there are extra steps we need to take to guarantee that custom proxy types will always be executable. To avoid unnecessary complexity at the outset, we are removing this functionality for now.


sam-goldman:
LGTM! Looks like there's a merge conflict though

PR #59, opened 2022-09-09
fix(ct): update `transferProxyOwnership` in `ChugSplashManager`

sam-goldman:
Fixes a couple minor things with the `transferProxyOwnership` function in the Manager contract.

### DIFF
--- a/packages/contracts/contracts/ChugSplashManager.sol
+++ b/packages/contracts/contracts/ChugSplashManager.sol
@@ -117,8 +117,8 @@ contract ChugSplashManager is Owned {
     );
 
     /**
-     * @notice Emitted when ownership of a proxy is transferred from the ProxyAdmin to the project
-     *         owner.
+     * @notice Emitted when the project owner transfers ownership of a proxy away from the
+     *         ChugSplashManager.
      *
      * @param targetNameHash Hash of the target's string name.
      * @param proxy          Address of the proxy that is the subject of the ownership transfer.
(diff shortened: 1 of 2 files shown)


packages/contracts/contracts/ChugSplashManager.sol
```diff
@@ -525,26 +525,38 @@ contract ChugSplashManager is Owned {
...
-     * @param _newOwner  Address of the project owner that is receiving ownership of the proxy.
+     * @param _name     String name of the target that corresponds to the proxy.
+     * @param _newOwner Address of the project owner that is receiving ownership of the proxy.
      */
-    function transferProxyOwnership(
-        address payable _proxy,
-        bytes32 _proxyType,
-        address _newOwner
-    ) public onlyOwner {
+    function transferProxyOwnership(string memory _name, address _newOwner) public onlyOwner {
```
sam-goldman, line 534: The main change is that this function now takes in a target's string name as input instead of the proxy and proxyType.  I think this makes things more consistent with other parts of the contract that do the same.

packages/contracts/contracts/ChugSplashManager.sol
```diff
@@ -525,26 +525,38 @@ contract ChugSplashManager is Owned {
...
 
         // Delegatecall the adapter to change ownership of the proxy.
         (bool success, ) = adapter.delegatecall(
-            abi.encodeCall(IProxyAdapter.changeProxyAdmin, (_proxy, _newOwner))
+            abi.encodeCall(IProxyAdapter.changeProxyAdmin, (proxy, _newOwner))
         );
-        require(success, "ProxyAdmin: delegatecall to change proxy admin failed");
+        require(success, "ChugSplashManager: delegatecall to change proxy admin failed");
+
+        emit ProxyOwnershipTransferred(_name, proxy, proxyType, _newOwner, _name);
```
sam-goldman, line 559: Also, this function didn't emit an event before

smartcontracts: Let's also announce this event to the registry. I think good rule of thumb is to announce everything by default. Makes it much easier for the website to display information.

smartcontracts (approved):

PR #60, opened 2022-09-10
feat(specs): ChugSplashManager speedspec

smartcontracts:
Adds a "speedspec" for the ChugSplashManager. This includes all of the key details without any fluff.

### DIFF
--- a/specs/chugsplash-manager.md
+++ b/specs/chugsplash-manager.md
@@ -3,6 +3,13 @@
 ## Bundle states
 
 ```typescript
+enum ActionType {
+  SET_CODE,
+  SET_STORAGE
+}
+```
+
+```typescript
 enum BundleStatus {
   EMPTY,
   PROPOSED,
@@ -11,6 +18,29 @@ enum BundleStatus {
   COMPLETED
 }
 ```
+
+```typescript
+struct BundleState {
+  BundleStatus status;
+  bool[] executions;
+  uint256 actionsExecuted;
+  uint256 timeClaimed;
+  address selectedExecutor;
+}
+```
+
+```typescript
+struct Action {
+  string target;
+  ActionType actionType;
+  uint256 gasUsed;
+  bytes code;
+}
+```
+
+## State variables
+
+`EXECUTION_COST_MULTIPLIER`: Ensure that enough ETH is deposited in the ChugSplashManager at the time that a bundle is approved. This allows a bundle to be executed even if the gas price spikes while it is active.
 
 ## Computing a unique bundle ID
 
@@ -25,6 +55,31 @@ const computeBundleId = (
       bundleRoot,
       bundleSize,
       configUri,
+      totalGasUsed
+    )
+  )
+}
+```
+
+## Getting the selected executor
+
+```typescript
+const getSelectedExecutor = (
+  bundleId: bytes32
+): address => {
+  return bundles[bundleId].selectedExecutor
+}
+```
+
+## Getting the proxy by name
+
+```typescript
+const getProxyByName = (
+  name: string
+): address payable => {
+  return (
+    payable(
+      Create2.compute(address(this), keccak256(bytes(name)), type(Proxy).creationCode)
     )
   )
 }
@@ -44,8 +99,27 @@ const computeBundleId = (
 * There MUST NOT be any active bundle.
 * Approving a bundle MUST put the bundle in the `APPROVED` state.
 * Approving a bundle MUST make the approved bundle the active bundle.
-* The balance of the ChugSplashManager minus the current `totalDebt` MUST be greater than `BUNDLE_EXECUTION_BOND_AMOUNT` ETH in its balance.
+* The amount of ETH available in the ChugSplashManager to cover the cost of deployment MUST be greater than or equal to the estimated upper bound for the cost of the deployment. In other words, `address(this).balance - totalDebt - BUNDLE_EXECUTION_BOND_AMOUNT >= EXECUTION_COST_MULTIPLIER * totalGasUsed * tx.gasPrice`.
 * The ChugSplashManager MUST emit the event ChugSplashBundleApproved with the ID of the bundle as the only parameter. The ChugSplashManager MUST announce this event to the registry.
+
+## Executing a ChugSplash action
+* There MUST be an `activeBundleId` to execute a ChugSplash action.
+* There MUST be enough reserved ETH in the ChugSplashManager to pay the executor for this action.
+* The `_actionIndex` MUST NOT already have been executed.
+* A ChugSplash action MUST be executed by the `selectedExecutor`.
+* The `_action`, `activeBundleId`, `_actionIndex`, `_gasUsed`, and `_proof` MUST produce a valid Merkle proof.
+* If the proxy corresponding to the `target` has not already been deployed:
+  * The proxy MUST be deployed using Create2.
+  * The proxy MUST have the same address as the the address returned by a call to `getProxyByName`.
+* A `SET_CODE` or `SET_ACTION` MUST be executed on the proxy depending on the `ActionType`.
+* Executing an action MUST increment the `actionsExecuted` by one.
+* Executing an action MUST set the current `_actionIndex` to `true`.
+* Executing an action MUST increase the `totalDebt` and the current executor's `debt` by the basefee times the `gasUsed` times `(100 + executorPaymentPercentage) / 100`.
+* The ChugSplashManager MUST emit the event ChugSplashActionExecuted with the active bundle ID, the executor's address, and the action index. The ChugSplashManager MUST announce this event to the registry.
+* If all actions have been executed for the bundle:
+  * The bundle status MUST be set to `COMPLETED`.
+  * The `activeBundleId` MUST be set to `bytes32(0)`.
+  * The ChugSplashManager MUST emit the event ChugSplashBundleCompleted with the active bundle ID, the executor's address, and the action index. The ChugSplashManager MUST announce this event to the registry.
 
 ## Cancelling a ChugSplash bundle
 
@@ -74,3 +148,31 @@ const computeBundleId = (
 ## Depositing ETH
 
 * Any address should be able to deposit ETH into the `ChugSplashManager` contract at any time.
+
+## Claiming a bundle
+
+* Anyone should be able to claim a bundle.
+* The executor's `debt` plus the `msg.value` MUST be greater than or equal to the `executorBondAmount`.
+* The bundle being claimed MUST have an `APPROVED` status.
+* The current `block.timestamp` MUST be greater than the `timeClaimed` plus the `executionLockTime`.
+* Claiming a bundle MUST set the `timeClaimed` to be the `block.timestamp`.
+* Claiming a bundle MUST set the `selectedExecutor` to be the executor that claimed the bundle.
+* The ChugSplashManager MUST increase the `debt` owed to the current executor by the `executorBondAmount`.
+* The ChugSplashManager MUST increase the `totalDebt` by the `executorBondAmount`.
+* The ChugSplashManager MUST emit the event ChugSplashBundleClaimed with the claimed bundle ID and the address of the executor. The ChugSplashManager MUST announce this event to the registry.
+
+## Setting the executor's bond amount
+
+* The `admin` of the `ChugSplashManager` MUST be the only address that can set the `executorBondAmount`.
+* The executor bond amount MUST NOT be changed while there is an active bundle.
+* MUST set `executorBondAmount` to its new value.
+* The ChugSplashManager MUST emit the ExecutorBondAmountSet event with the new `executorBondAmount`. The ChugSplashManager MUST announce this event to the registry.
+
+## Transferring proxy ownership from the ChugSplashManager
+
+* The `admin` of the `ChugSplashManager` MUST be the only address that can transfer proxy ownership from the ChugSplashManager.
+* The proxy MUST already be deployed.
+* The owner of the proxy MUST be the ChugSplashManager.
+* If there is an active bundle ID, the `totalDebt` and the current executor's `debt` MUST be increased by `BUNDLE_EXECUTION_BOND_AMOUNT`.
+* The delegatecall to transfer proxy ownership MUST succeed.
+* The ChugSplashManager MUST emit the ProxyOwnershipTransferred event with the name of the target corresponding to the proxy (indexed and unindexed), the address of the proxy, and the new owner of the proxy. The ChugSplashManager MUST announce this event to the registry.
--- a/specs/chugsplash-manager.md
+++ b/specs/chugsplash-manager.md
@@ -1,0 +1,76 @@
+# ChugSplashManager Speed Spec
+
+## Bundle states
+
+```typescript
+enum BundleStatus {
+  EMPTY,
+  PROPOSED,
+  APPROVED,
+  CANCELLED,
+  COMPLETED
+}
+```
+
+## Computing a unique bundle ID
+
+```typescript
+const computeBundleId = (
+  bundleRoot: bytes32,
+  bundleSize: uint256,
+  configUri: string
+): bytes => {
+  return keccak256(
+    abi.encode(
+      bundleRoot,
+      bundleSize,
+      configUri,
+    )
+  )
+}
+```
+
+## Proposing a ChugSplash bundle
+
+* A ChugSplash bundle MUST only be proposable by the `admin` of the `ChugSplashManager`.
+* Bundles MUST be in the `EMPTY` state before they can be proposed.
+* Proposing a bundle puts the bundle in the `PROPOSED` state.
+* The ChugSplashManager MUST emit the event ChugSplashBundleProposed with the ID of the bundle, the bundle root, the bundle size, and the config URI as parameters. The ChugSplashManager MUST announce this event to the registry.
+
+## Approving a ChugSplash bundle
+
+* A ChugSplash bundle MUST only be approvable by the `admin` of the `ChugSplashManager`.
+* Bundles MUST have be in the `PROPOSED` state before they can be approved.
+* There MUST NOT be any active bundle.
+* Approving a bundle MUST put the bundle in the `APPROVED` state.
+* Approving a bundle MUST make the approved bundle the active bundle.
+* The balance of the ChugSplashManager minus the current `totalDebt` MUST be greater than `BUNDLE_EXECUTION_BOND_AMOUNT` ETH in its balance.
+* The ChugSplashManager MUST emit the event ChugSplashBundleApproved with the ID of the bundle as the only parameter. The ChugSplashManager MUST announce this event to the registry.
+
+## Cancelling a ChugSplash bundle
+
+* A ChugSplash bundle MUST only be cancellable by the `admin` of the `ChugSplashManager`.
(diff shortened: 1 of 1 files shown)


specs/chugsplash-manager.md
```diff
@@ -0,0 +1,76 @@
...
+      configUri,
+    )
+  )
+}
+```
+
+## Proposing a ChugSplash bundle
+
+* A ChugSplash bundle MUST only be proposable by the `admin` of the `ChugSplashManager`.
+* Bundles MUST be in the `EMPTY` state before they can be proposed.
```
sam-goldman, line 36: This prevents users from being able to roll back a project to a previously executed bundle. I think we can fix this by just changing this to "Bundles MUST NOT be in the `PROPOSED` or `APPROVED` state before they can be proposed."

specs/chugsplash-manager.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Cancelling a ChugSplash bundle
+
+* A ChugSplash bundle MUST only be cancellable by the `admin` of the `ChugSplashManager`.
+* Bundles MUST be in the `APPROVED` state before they can be cancelled.
+* Cancelling a bundle MUST put the bundle in the `CANCELLED` state.
+* Cancelling a bundle MUST remove the active bundle.
+* The ChugSplashManager MUST increase the `debt` owed to the current executor by `BUNDLE_EXECUTION_BOND_AMOUNT`.
+* The ChugSplashManager MUST increase the `totalDebt` by the `BUNDLE_EXECUTION_BOND_AMOUNT`.
+* The ChugSplashManager MUST emit the event ChugSplashBundleCancelled with the ID of the bundle and the total number of actions executed. The ChugSplashManager MUST announce this event to the registry.
```
sam-goldman, line 58: Should the event also include the `admin` of the `ChugSplashManager`, or do you think that's unnecessary?

specs/chugsplash-manager.md
```diff
@@ -0,0 +1,76 @@
...
+* The ChugSplashManager MUST emit the event ChugSplashBundleProposed with the ID of the bundle, the bundle root, the bundle size, and the config URI as parameters. The ChugSplashManager MUST announce this event to the registry.
+
+## Approving a ChugSplash bundle
+
+* A ChugSplash bundle MUST only be approvable by the `admin` of the `ChugSplashManager`.
+* Bundles MUST have be in the `PROPOSED` state before they can be approved.
+* There MUST NOT be any active bundle.
+* Approving a bundle MUST put the bundle in the `APPROVED` state.
+* Approving a bundle MUST make the approved bundle the active bundle.
+* The balance of the ChugSplashManager minus the current `totalDebt` MUST be greater than `BUNDLE_EXECUTION_BOND_AMOUNT` ETH in its balance.
```
sam-goldman, line 47: Shouldn't this be "greater than or equal to" instead of "greater than"?

specs/chugsplash-manager.md
```diff
@@ -0,0 +1,76 @@
...
+* The ChugSplashManager MUST emit the event ChugSplashBundleApproved with the ID of the bundle as the only parameter. The ChugSplashManager MUST announce this event to the registry.
+
+## Cancelling a ChugSplash bundle
+
+* A ChugSplash bundle MUST only be cancellable by the `admin` of the `ChugSplashManager`.
+* Bundles MUST be in the `APPROVED` state before they can be cancelled.
+* Cancelling a bundle MUST put the bundle in the `CANCELLED` state.
+* Cancelling a bundle MUST remove the active bundle.
+* The ChugSplashManager MUST increase the `debt` owed to the current executor by `BUNDLE_EXECUTION_BOND_AMOUNT`.
+* The ChugSplashManager MUST increase the `totalDebt` by the `BUNDLE_EXECUTION_BOND_AMOUNT`.
```
sam-goldman, line 57: I think we should only increase the `debt` and `totalDebt` if there's a `selectedExecutor` that isn't `address(0)`. Otherwise, the ETH is essentially lost.

specs/chugsplash-manager.md
```diff
@@ -0,0 +1,76 @@
...
+
+* ETH stored in the contract MUST only be withdrawable by the `admin` of the `ChugSplashManager`.
+* The admin ETH withdrawal function MUST NOT be callable while a bundle is in the `APPROVED` state.
+* The admin ETH withdrawal function MUST only allow the admin to withdraw the balance of the account minus `totalDebt`.
+* The ChugSplashManager MUST emit the ETHWithdrawn event with the recipient and the amount. The ChugSplashManager MUST announce this event to the registry.
+
+## Claiming executor payments
+
+* Executors MUST be able to claim payments at any time.
+* Executors MUST NOT be able to claim more payments than their `debt` amount.
```
sam-goldman, line 138: We might want to change this line to the following:
* If the caller is the `selectedExecutor`, they MUST NOT be able to claim more payments than their `debt` amount minus the `executorBondAmount`.
* Otherwise, the executor MUST NOT be able to claim more payments than their `debt` amount.

specs/chugsplash-manager.md
```diff
@@ -0,0 +1,178 @@
...
+* The ChugSplashManager MUST emit the event ChugSplashBundleProposed with the ID of the bundle, the bundle root, the bundle size, and the config URI as parameters. The ChugSplashManager MUST announce this event to the registry.
+
+## Approving a ChugSplash bundle
+
+* A ChugSplash bundle MUST only be approvable by the `admin` of the `ChugSplashManager`.
+* Bundles MUST have be in the `PROPOSED` state before they can be approved.
+* There MUST NOT be any active bundle.
+* Approving a bundle MUST put the bundle in the `APPROVED` state.
+* Approving a bundle MUST make the approved bundle the active bundle.
+* The amount of ETH available in the ChugSplashManager to cover the cost of deployment MUST be greater than or equal to the estimated upper bound for the cost of the deployment. In other words, `address(this).balance - totalDebt - BUNDLE_EXECUTION_BOND_AMOUNT >= EXECUTION_COST_MULTIPLIER * totalGasUsed * tx.gasPrice`.
```
sam-goldman, line 102: I felt like this was a reasonable approach since we don't know how much a deployment will cost up-front due to fluctuations in gas price. So, the purpose of this line is to ensure that the owner has deposited more than enough ETH even if the gas price spikes while the bundle is being executed.

specs/chugsplash-manager.md
smartcontracts: Let's ignore rollbacks for the v1, there's a lot to figure out with how to do rollbacks safely

specs/chugsplash-manager.md
smartcontracts: Hmmm this one is interesting. I think we should decide based on the ideal user experience. I do like the idea of estimating ahead of time what the upper-bound cost will be. I'm not entirely convinced that enforcing this at the smart contract level is best though. My preference is still to only require the bond up-front, but have the UI estimate how much you will need to put into the account for someone to pick up the job.

specs/chugsplash-manager.md
smartcontracts: Good point

specs/chugsplash-manager.md
```diff
@@ -0,0 +1,178 @@
...
+* The ChugSplashManager MUST emit the ExecutorPayment event with the recipient and amount. The ChugSplashManager MUST announce this event to the registry.
+
+## Depositing ETH
+
+* Any address should be able to deposit ETH into the `ChugSplashManager` contract at any time.
+
+## Claiming a bundle
+
+* Anyone should be able to claim a bundle.
+* The executor's `debt` plus the `msg.value` MUST be greater than or equal to the `executorBondAmount`.
```
smartcontracts, line 155: Oh interesting, this allows the executor to reuse their bond, which is nice.

smartcontracts: Although I would be very surprised if an executor doesn't just claim the debt immediately after an upgrade

specs/chugsplash-manager.md
smartcontracts: I think we can get around this if we add `executorBondAmount` to `totalDebt` immediately but only to the executor's `debt` when the execution is completed.

specs/chugsplash-manager.md
smartcontracts: For simplicity let's require that the executor provide the bond as msg.value. I don't really want to make the debt management logic more complex than it already is.

specs/chugsplash-manager.md
smartcontracts: Yeah that totally works since you can also just subtract the `executorBondAmount` from the `totalDebt` when a new executor overrides a previous executor instead of doing a payment.

specs/chugsplash-manager.md
```diff
@@ -0,0 +1,178 @@
...
+* The bundle being claimed MUST have an `APPROVED` status.
+* The current `block.timestamp` MUST be greater than the `timeClaimed` plus the `executionLockTime`.
+* Claiming a bundle MUST set the `timeClaimed` to be the `block.timestamp`.
+* Claiming a bundle MUST set the `selectedExecutor` to be the executor that claimed the bundle.
+* The ChugSplashManager MUST increase the `debt` owed to the current executor by the `executorBondAmount`.
+* The ChugSplashManager MUST increase the `totalDebt` by the `executorBondAmount`.
+* The ChugSplashManager MUST emit the event ChugSplashBundleClaimed with the claimed bundle ID and the address of the executor. The ChugSplashManager MUST announce this event to the registry.
+
+## Setting the executor's bond amount
```
smartcontracts, line 164: For simplicity for now let's get rid of the ability to set the executor bond and use a reasonable default.

specs/chugsplash-manager.md
```diff
@@ -0,0 +1,178 @@
...
+## Depositing ETH
+
+* Any address should be able to deposit ETH into the `ChugSplashManager` contract at any time.
+
+## Claiming a bundle
+
+* Anyone should be able to claim a bundle.
+* The executor's `debt` plus the `msg.value` MUST be greater than or equal to the `executorBondAmount`.
+* The bundle being claimed MUST have an `APPROVED` status.
+* The current `block.timestamp` MUST be greater than the `timeClaimed` plus the `executionLockTime`.
```
smartcontracts, line 157: We need to specify the behavior of what happens when one executor overrides a previous executor

specs/chugsplash-manager.md
```diff
@@ -0,0 +1,178 @@
...
+* The executor bond amount MUST NOT be changed while there is an active bundle.
+* MUST set `executorBondAmount` to its new value.
+* The ChugSplashManager MUST emit the ExecutorBondAmountSet event with the new `executorBondAmount`. The ChugSplashManager MUST announce this event to the registry.
+
+## Transferring proxy ownership from the ChugSplashManager
+
+* The `admin` of the `ChugSplashManager` MUST be the only address that can transfer proxy ownership from the ChugSplashManager.
+* The proxy MUST already be deployed.
+* The owner of the proxy MUST be the ChugSplashManager.
+* If there is an active bundle ID, the `totalDebt` and the current executor's `debt` MUST be increased by `BUNDLE_EXECUTION_BOND_AMOUNT`.
```
smartcontracts, line 176: We should not allow transferring proxy ownership during an active upgrade. Cancel or wait for it to be completed instead.

specs/chugsplash-manager.md
```diff
@@ -0,0 +1,178 @@
...
+* The `admin` of the `ChugSplashManager` MUST be the only address that can set the `executorBondAmount`.
+* The executor bond amount MUST NOT be changed while there is an active bundle.
+* MUST set `executorBondAmount` to its new value.
+* The ChugSplashManager MUST emit the ExecutorBondAmountSet event with the new `executorBondAmount`. The ChugSplashManager MUST announce this event to the registry.
+
+## Transferring proxy ownership from the ChugSplashManager
+
+* The `admin` of the `ChugSplashManager` MUST be the only address that can transfer proxy ownership from the ChugSplashManager.
+* The proxy MUST already be deployed.
+* The owner of the proxy MUST be the ChugSplashManager.
```
smartcontracts, line 175: I don't think this line and the line above are necessary to specify -- if the user tries to transfer something that the `ChugSplashManager` doesn't control then it'll just fail, which is fine.

specs/chugsplash-manager.md
```diff
@@ -0,0 +1,178 @@
...
+* If there is an active bundle ID, the `totalDebt` and the current executor's `debt` MUST be increased by `BUNDLE_EXECUTION_BOND_AMOUNT`.
+* The delegatecall to transfer proxy ownership MUST succeed.
```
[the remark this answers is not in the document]
sam-goldman: I was just referring to [this line](https://github.com/smartcontracts/chugsplash/blob/develop/packages/contracts/contracts/ChugSplashManager.sol#L599)

specs/chugsplash-manager.md
sam-goldman: Added this to the spec

specs/chugsplash-manager.md
sam-goldman: Sounds good, I changed this back to what it was before

specs/chugsplash-manager.md
sam-goldman: This function now increments `totalDebt` only if there was no previously selected executor for the bundle. This way, the owner is able to claim any bonds from executors that weren't able to complete the bundle on time.

specs/chugsplash-manager.md
sam-goldman: > I think we can get around this if we add `executorBondAmount` to `totalDebt` immediately but only to the executor's `debt` when the execution is completed.

Made this change

> Yeah that totally works since you can also just subtract the `executorBondAmount` from the `totalDebt` when a new executor overrides a previous executor instead of doing a payment.

I actually don't think we need to subtract the `executorBondAmount` from the `totalDebt` when a new executor overrides a previous executor. I think we can just leave the `totalDebt` as it currently is. See my comment [here](https://github.com/smartcontracts/chugsplash/pull/60#discussion_r974850462)

specs/chugsplash-manager.md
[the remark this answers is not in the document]
smartcontracts: Oh ok that's fine

specs/chugsplash-manager.md
```diff
@@ -0,0 +1,165 @@
...
+  string target;
+  ActionType actionType;
+  bytes code;
+}
+```
+
+## Computing a unique bundle ID
+
+```typescript
+const computeBundleId = (
```
kethcode, line 43: What makes this unique?  Merkle Roots for identical deployemnts can be the same.  Size and Config could be the same as well.  If someone wants to deploy same thing multiple times, will the config have different target addresses or something?

specs/chugsplash-manager.md
```diff
@@ -0,0 +1,165 @@
...
+* The ChugSplashManager MUST emit the event ChugSplashBundleProposed with the ID of the bundle, the bundle root, the bundle size, and the config URI as parameters. The ChugSplashManager MUST announce this event to the registry.
+
+## Approving a ChugSplash bundle
+
+* A ChugSplash bundle MUST only be approvable by the `admin` of the `ChugSplashManager`.
+* Bundles MUST have be in the `PROPOSED` state before they can be approved.
+* There MUST NOT be any active bundle.
+* Approving a bundle MUST put the bundle in the `APPROVED` state.
+* Approving a bundle MUST make the approved bundle the active bundle.
+* The balance of the ChugSplashManager minus the current totalDebt MUST be greater than BUNDLE_EXECUTION_BOND_AMOUNT ETH in its balance.
```
kethcode, line 96: We may want an additional buffer here to account for gas price variability.  Having enough to cover the bond, but not send it back, could be a problem.

specs/chugsplash-manager.md
```diff
@@ -0,0 +1,165 @@
...
+## Executing a ChugSplash action
+* There MUST be an `activeBundleId` to execute a ChugSplash action.
+* There MUST be enough reserved ETH in the ChugSplashManager to pay the executor for this action.
+* The `_actionIndex` MUST NOT already have been executed.
+* A ChugSplash action MUST be executed by the `selectedExecutor`.
+* The `_action`, `activeBundleId`, `_actionIndex`, and `_proof` MUST produce a valid Merkle proof.
+* If the proxy corresponding to the `target` has not already been deployed:
+  * The proxy MUST be deployed using Create2.
+  * The proxy MUST have the same address as the the address returned by a call to `getProxyByName`.
+* A `SET_CODE` or `SET_ACTION` MUST be executed on the proxy depending on the `ActionType`.
```
kethcode, line 108: I see if/else statement for this in the current code (https://github.com/smartcontracts/chugsplash/blob/cd78ceb1cfa4ff61a98dcc247cf15e4f4c308491/packages/contracts/contracts/ChugSplashManager.sol#L393).  Is it possible for the action to not be either of these enum states?  if so, this might default to _setProxyStorage incorrectly

specs/chugsplash-manager.md
```diff
@@ -0,0 +1,165 @@
...
+* Executing an action MUST increment the `actionsExecuted` by one.
+* Executing an action MUST set the current `_actionIndex` to `true`.
+* The ChugSplashManager MUST emit the event ChugSplashActionExecuted with the active bundle ID, the executor's address, and the action index. The ChugSplashManager MUST announce this event to the registry.
+* If all actions have been executed for the bundle:
+  * The bundle status MUST be set to `COMPLETED`.
+  * The ChugSplashManager MUST increase the `debt` owed to the current executor by the `executorBondAmount`.
```
kethcode, line 114: I dont see the debt variables in the existing contract, but I agree this is a critical feature, and the flow of debt and repayment should be carefully mapped out.

specs/chugsplash-manager.md
sam-goldman: I believe the action must be one of those two enum states or else the function would just revert

specs/chugsplash-manager.md
sam-goldman: Agreed, we're going to finalize this spec before updating the contract

specs/chugsplash-manager.md
kethcode: I guess this creates a requirement for future versions of the enum to always support the previous enum list.  It just felt like a potential dangerous failure, and I don't know how impossible it is to create that situation.  It may be a non-issue, as you suggest.

specs/chugsplash-manager.md
```diff
@@ -0,0 +1,165 @@
...
+
+* Executors MUST be able to claim payments at any time.
+* Executors MUST NOT be able to claim more payments than their `debt` amount.
+* The ChugSplashManager MUST update the `debt` and `totalDebt` values to reflect the payment.
+* The ChugSplashManager MUST emit the ExecutorPayment event with the recipient and amount. The ChugSplashManager MUST announce this event to the registry.
+
+## Depositing ETH
+
+* Any address should be able to deposit ETH into the `ChugSplashManager` contract at any time.
+
```
kethcode, line 147: How do people get paid?  I assume executors aren't doing this for free, and I only see bond repayment mechanisms here.

sam-goldman: Executor payments are under the section "Claiming executor payments" on line 137, which basically states that executors are able to withdraw the `debt` owed to them. This debt is aggregated in the "Executing a ChugSplash action" section on line 99, which includes an `executorPaymentPercentage` (i.e. the executor's payment).

PR #61, opened 2022-09-22
feat(pg): add local hardhat deployments

sam-goldman:
1. `cd packages/demo`
2. `npx hardhat test`


sam-goldman:
I'm getting a `CALL_EXCEPTION` when I try to interact with the implementation address of `MyToken` after deploying it locally via ChugSplash.

Steps to reproduce:

In `chugsplash/`
```
yarn build
cd packages/contracts
npx hardhat node
```

In another terminal:
```
cd packages/demo
npx hardhat run --network localhost scripts/test.js
```

This last line will do a local ChugSplash deployment that intentionally halts after the first action so that it just deploys MyToken (without triggering any `setStorage` actions). You can see the deployed implementation address in the Hardhat node terminal:
```
console.log:
  MyToken implementation address:  0xa4039a0aa015b9dc5d5ab37e5bf365d672331475
```

Next:
```
cd packages/demo
npx hardhat run --network localhost error.js
```

This results in a `CALL_EXCEPTION`. Why?



Note that if you change any code in the contracts, you'll likely need to manually change the address of the `ChugSplashRegistry` and/or `TransparentUpgradeableAdapter`, then re-build.

sam-goldman:
Also, it appears that the [Merkle proof verification](https://github.com/smartcontracts/chugsplash/blob/2bb5b0ba001f8ba541e362bf102429f6abebf9f9/packages/contracts/contracts/ChugSplashManager.sol#L393-L402) in the ChugSplashManager isn't working. [Here's the script](https://github.com/smartcontracts/chugsplash/blob/2bb5b0ba001f8ba541e362bf102429f6abebf9f9/packages/plugins/src/hardhat/deployments.ts#L149-L158) that calls `executeAction`. (It's the same code as the executor logic [here](https://github.com/smartcontracts/chugsplash/blob/develop/packages/executor/src/index.ts#L171-L180))

sam-goldman:
> Also, it appears that the [Merkle proof verification](https://github.com/smartcontracts/chugsplash/blob/2bb5b0ba001f8ba541e362bf102429f6abebf9f9/packages/contracts/contracts/ChugSplashManager.sol#L393-L402) in the ChugSplashManager isn't working. [Here's the script](https://github.com/smartcontracts/chugsplash/blob/2bb5b0ba001f8ba541e362bf102429f6abebf9f9/packages/plugins/src/hardhat/deployments.ts#L149-L158) that calls `executeAction`. (It's the same code as the executor logic [here](https://github.com/smartcontracts/chugsplash/blob/develop/packages/executor/src/index.ts#L171-L180))

Fixed this. It turns out that the Merkle tree was using the `activeBundleId` instead of the root hash.

sam-goldman:
> Also, it appears that the [Merkle proof verification](https://github.com/smartcontracts/chugsplash/blob/2bb5b0ba001f8ba541e362bf102429f6abebf9f9/packages/contracts/contracts/ChugSplashManager.sol#L393-L402) in the ChugSplashManager isn't working. [Here's the script](https://github.com/smartcontracts/chugsplash/blob/2bb5b0ba001f8ba541e362bf102429f6abebf9f9/packages/plugins/src/hardhat/deployments.ts#L149-L158) that calls `executeAction`. (It's the same code as the executor logic [here](https://github.com/smartcontracts/chugsplash/blob/develop/packages/executor/src/index.ts#L171-L180))
 
Fixed this. It turns out that the Merkle tree was using the `activeBundleId` as the Merkle root hash when it should've been using `bundle.root`.

sam-goldman:
> I'm getting a `CALL_EXCEPTION` when I try to interact with the implementation address of `MyToken` after deploying it locally via ChugSplash.

Solved this as well. The creation bytecode was being appended to the `DEPLOY_CODE_PREFIX` when it should've been the runtime bytecode.

packages/plugins/src/hardhat/tasks.ts
```diff
@@ -436,10 +514,6 @@ task(TASK_CHUGSPLASH_COMMIT)
       })
       spinner.succeed('Compiled deploy config')
 
-      const ipfs = create({
-        url: args.ipfsUrl || 'https://ipfs.infura.io:5001/api/v0',
-      })
```
sam-goldman, line 441: Infura deprecated their public IPFS gateway ([link](https://docs.infura.io/infura/networks/ipfs/how-to/access-ipfs-content/public-gateway)). Since I couldn't find another public IPFS gateway with the ability to write files, I changed this logic to the following:
* If the user adds the `local` flag to this task, then we get the IPFS hash that *would* be published to IPFS without actually publishing it.
* Alternatively, the user can supply their own IPFS URL like before.
* Lastly, the user can supply an Infura project ID and API key, which are now required to use an Infura node.

sam-goldman:
It looks like CI is failing because the @chugsplash/contracts package is outdated. I think it'd be fine to just merge this PR into develop then bump the version

PR #62, opened 2022-09-26
Update executor-economics.md

kethcode:
brainstormed some ESS options


sam-goldman:
Going to close this for now since we're going to stick with the simple ESS for the time being. Will re-open if we're adding new ESS types!

PR #149, opened 2022-11-05
Add Support for Mapping Types

RPate97:
- Adds support for mappings, including nested mappings and structs
- Adds a basic automated integration test suite to the plugins package

### DIFF
--- a/packages/core/src/languages/solidity/storage.ts
+++ b/packages/core/src/languages/solidity/storage.ts
@@ -300,7 +300,7 @@ const encodeVariable = (
       } else if (variableType.key.startsWith('t_bytes')) {
         // bytes do not need to be encoded, but must be converted from the input string
         // pack type can be pulled straight from input type
-        encodedKey = ethers.utils.toUtf8Bytes(key)
+        encodedKey = fromHexString(key)
       }
 
       // key for nested mappings is computed by packing and hashing the key of the child mapping
--- a/packages/core/src/languages/solidity/storage.ts
+++ b/packages/core/src/languages/solidity/storage.ts
@@ -1,5 +1,5 @@
 import { fromHexString, remove0x } from '@eth-optimism/core-utils'
-import { BigNumber, ethers } from 'ethers'
+import { BigNumber, ethers, utils } from 'ethers'
 
 import { ContractConfig } from '../../config'
 import {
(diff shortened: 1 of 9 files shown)


packages/core/src/languages/solidity/storage.ts
```diff
@@ -266,7 +273,64 @@ const encodeVariable = (
...
+        // all ints must be packed with type int256
+        type = 'int256'
+      } else if (variableType.key.startsWith('t_string')) {
+        // strings do not need to be encoded
+        // pack type can be pulled from input type
+        encodedKey = key
+      } else if (variableType.key.startsWith('t_bytes')) {
+        // bytes do not need to be encoded, but must be converted from the input string
+        // pack type can be pulled straight from input type
+        encodedKey = ethers.utils.toUtf8Bytes(key)
```
sam-goldman, line 303: I think we actually want this line to be:
```
encodedKey = fromHexString(key)
```
This is because we want the original `key` to be of the form `0x...` (see my other comments)

packages/plugins/chugsplash/SimpleStorage.config.ts
```diff
@@ -0,0 +1,84 @@
...
+          1: 'test',
+        },
+        uint128StrTest: {
+          1: 'test',
+        },
+        addressStrTest: {
+          '0x1111111111111111111111111111111111111111': 'test',
+        },
+        bytesStrTest: {
+          abcd: 'test',
```
sam-goldman, line 65: Bytes are usually 0x-prefixed, so I think we'd want the key here to be `0xabcd` instead of `abcd`

packages/plugins/test/SimpleStorage.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+      await FirstSimpleStorage.getAddressStringTestMappingValue(
+        '0x1111111111111111111111111111111111111111'
+      )
+    ).to.equal('test')
+  })
+
+  it('does set bytes mapping to string', async () => {
+    expect(
+      await FirstSimpleStorage.getBytesStringTestMappingValue(
+        ethers.utils.toUtf8Bytes('abcd')
```
sam-goldman, line 89: If we support 0x-prefixed bytes variables (see my other comments), we can replace this line with just `0xabcd` without converting it to utf-8

sam-goldman (approved):

PR #177, opened 2022-11-09
Minimal Executor

RPate97:
- Implements a minimal executor that runs against localhost
- Adds a new CLI parameter `local` to the `node`, `chugsplash-deploy`, and `test` commands which defaults to `true`. This argument determines if the execution command should be run by the CLI or delegated to a separate executor process. This is a stopgap measure to allow testing the executor while continuing to default to fully local deployments. In the future, we'll need to add logic to base this on the network the user is deploying too. 
- Refactors the `TASK_CHUGSPLASH_BUNDLE_REMOTE` task to compile the artifacts necessary to complete the bundle using only the output from running the compiler programmatically. Previously, helper functions pulled artifact data from files that aren't available to the executor.

Note that the `TASK_CHUGSPLASH_BUNDLE_REMOTE` command includes a great deal of logic that is shared with the `getArtifactsFromCanonicalConfig` function which is a dependency of the `verifyChugSplashConfig` function. So, we'll need to refactor these functions to avoid duplicating that logic.

# Testing
- First, in the plugins package run: 
`npx hardhat node --local false`
- Then in the executor package run: 
`yarn start`

### DIFF
--- a/packages/executor/src/index.ts
+++ b/packages/executor/src/index.ts
@@ -74,10 +74,7 @@ export class ChugSplashExecutor extends BaseServiceV2<Options, Metrics, State> {
       )
 
       const activeBundleId = await manager.activeBundleId()
-      if (
-        activeBundleId ===
-        '0x0000000000000000000000000000000000000000000000000000000000000000'
-      ) {
+      if (activeBundleId === ethers.constants.HashZero) {
         console.log('no active bundle')
         continue
       }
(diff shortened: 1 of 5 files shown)


packages/executor/src/index.ts
```diff
@@ -39,90 +33,59 @@ export class ChugSplashExecutor extends BaseServiceV2<Options, Metrics, State> {
...
-      }
-
-      // TODO: Add this to the ChugSplashManager contract
-      const eps = await manager.getExecutorPaymentStrategy()
-      if (!this.state.eps.includes(eps)) {
-        // We don't like this strategy, skip the upgrade.
+      const activeBundleId = await manager.activeBundleId()
+      if (
+        activeBundleId ===
+        '0x0000000000000000000000000000000000000000000000000000000000000000'
```
sam-goldman, line 79: nit: this can be ethers.constants.HashZero

sam-goldman (approved):

PR #189, opened 2022-11-13
Add Docker Configuration for Executor

RPate97:
## Purpose
Adds Docker container configuration for the executor.

You'll need to install [Docker Desktop](https://www.docker.com/products/docker-desktop/) to test this feature locally. 

## Testing
1. In the root directory run `yarn build`.
2. Navigate to the plugins directory and run a local node `npx hardhat node --local false`.
3. Navigate to the `executor` directory, make a copy of .env.example, rename it to .env, and fill in the values (the Amplitude API key is optional).
4. Build the container `yarn build-container`
5. Run the container `yarn run-container`
6. You should see the container start up and output some logs. After ~20 seconds, you'll see logs indicating that the test contract from the plugins package was executed. 

You can also try out deploying other projects with the executor using `npx hardhat chugsplash-deploy --local false`.

## Todo
- [x] Test against Optimism Goerli

### DIFF
--- a/packages/executor/src/index.ts
+++ b/packages/executor/src/index.ts
@@ -8,11 +8,14 @@ import {
 } from '@chugsplash/contracts'
 import { ChugSplashBundleState } from '@chugsplash/core'
 import { getChainId } from '@eth-optimism/core-utils'
+import * as Amplitude from '@amplitude/node'
 
 import { compileRemoteBundle, verifyChugSplashConfig } from './utils'
 
 type Options = {
   network: string
+  privateKey: string
+  amplitudeKey: string
 }
 
 type Metrics = {}
@@ -20,14 +23,18 @@ type Metrics = {}
 type State = {
   registry: ethers.Contract
   wallet: ethers.Wallet
-  ess: string[]
-  eps: string[]
+  lastBlockNumber: number
+  amplitudeClient: Amplitude.NodeClient
 }
+
+// TODO:
+// Add logging agent for docker container and connect to a managed sink such as logz.io
+// Refactor chugsplash commands to decide whether to use the executor based on the target network
 
 export class ChugSplashExecutor extends BaseServiceV2<Options, Metrics, State> {
   constructor(options?: Partial<Options>) {
     super({
-      name: 'executor',
+      name: 'chugsplash-executor',
       // eslint-disable-next-line @typescript-eslint/no-var-requires
       version: require('../package.json').version,
       loop: true,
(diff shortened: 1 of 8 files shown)


packages/executor/src/index.ts
```diff
@@ -37,89 +44,140 @@ export class ChugSplashExecutor extends BaseServiceV2<Options, Metrics, State> {
...
 
+    // If none found, return
+    if (approvalAnnouncementEvents.length === 0) {
+      this.logger.info('no events found')
+      return
+    }
+
+    this.logger.info(`${approvalAnnouncementEvents.length} events found`)
+
+    // store last block hash
```
sam-goldman, line 93: nit: block hash -> block number

packages/executor/tsconfig.json
```diff
@@ -1,11 +1,27 @@
...
+      "node_modules/@types"
+    ],
     "rootDir": "./src",
     "outDir": "./dist"
   },
-
+  "exclude": [
+    "node_modules",
+    "dist"
+  ],
```
sam-goldman, line 24: Why can't we just extend the root tsconfig? Does it have to do with the docker container?

RPate97: Yeah, we need the full config to be copied into the docker container for it to work correctly. 

There probably ways to avoid the duplication here like copying both the root config and the package config or combining them together and copying the result into the container, but I didn't think it was worth worrying about at the moment.

packages/executor/src/index.ts
```diff
@@ -37,89 +44,140 @@ export class ChugSplashExecutor extends BaseServiceV2<Options, Metrics, State> {
...
+        }
+      } catch (e) {
+        this.logger.error(
+          'Error: verification error',
+          e,
+          canonicalConfig.options
+        )
+      }
+
+      this.logger.info('Successfully executed and verified')
```
sam-goldman, line 171: It'd be nice to separate this into two separate logs: a 'Successfully executed' log that's called after `chugsplash-execute` above, and a 'Successfully verified' log that's called here. This way it's easier to distinguish errors caused during execution and errors caused during verification. I expect the latter to fail pretty often since Etherscan's API can be flakey

packages/executor/src/index.ts
```diff
@@ -37,89 +44,138 @@ export class ChugSplashExecutor extends BaseServiceV2<Options, Metrics, State> {
...
+          parsedConfig: canonicalConfig,
+          deployer: signer,
+          hide: false,
+        })
+      } catch (e) {
+        // log error and continue
+        this.logger.error('Error: execution error', e, canonicalConfig.options)
+        continue
+      } finally {
+        this.logger.info('Successfully executed')
```
sam-goldman, line 153: I think this should go on line 147 since code inside of a 'finally' statement will execute when an exception is thrown (i.e. when execution fails)

packages/executor/src/index.ts
```diff
@@ -37,89 +44,138 @@ export class ChugSplashExecutor extends BaseServiceV2<Options, Metrics, State> {
...
+          await verifyChugSplashConfig(hre, proposalEvent.args.configUri)
+        }
+      } catch (e) {
+        this.logger.error(
+          'Error: verification error',
+          e,
+          canonicalConfig.options
+        )
+      } finally {
+        this.logger.info('Successfully verified')
```
sam-goldman, line 168: Ditto ^

sam-goldman (approved):

PR #191, opened 2022-11-14
Parallelize Execution Process

RPate97:
## Purpose
Refactors the execution process so that bundle actions are executed in parallel to improve execution speed. 

Note that transaction queuing does not work when automining is enabled on the hardhat network. So if this PR is merged, then all users will be required to use interval mining when testing against the local hardhat network. If it's possible to detect the mining mode of the hardhat network, then we might want to dynamically choose parallel or synchronous execution based on that. This would allow us to support both modes. For the moment though, I think this is a reasonable restriction that we can revisit if/when we encounter a user that cares a lot about using automining.

### DIFF
--- a/packages/plugins/dev.ts
+++ b/packages/plugins/dev.ts
@@ -57,7 +57,7 @@ const deployConfigForceCommit = async (fileName: string, silent: boolean) => {
   const { bundleId } = await hre.run('chugsplash-commit', {
     configPath,
     ipfsUrl: '',
-    commitToIpfs: false,
+    commitToIpfs: true,
   })
 
   const ChugSplashManager = new Contract(
--- a/.changeset/spicy-geese-roll.md
+++ b/.changeset/spicy-geese-roll.md
@@ -1,0 +1,7 @@
+---
+'@chugsplash/demo': patch
+'@chugsplash/executor': patch
+'@chugsplash/plugins': patch
+---
+
+Improve execution speed with parallelization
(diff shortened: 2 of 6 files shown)


sam-goldman:
This is awesome!!

I think it'd be nice to give users the option to use automining if they'd like. (Fwiw, I'd use automining). This can be done pretty easily:
```
if (
  (await getChainId(hre.ethers.provider)) === 31337 &&
  (await hre.network.provider.request({
    method: 'hardhat_getAutomine',
    params: [],
  }))
) {
  // execute actions in series
} else {
  // execute in parallel
}
```
Checking the chain ID in the first part of the if-statement is important since the call to `hardhat_getAutomine` will error if the network isn't Hardhat. The documentation for `hardhat_getAutomine` is [here](https://hardhat.org/hardhat-network/docs/reference#hardhat_getautomine).

packages/plugins/dev.ts
```diff
@@ -57,7 +57,7 @@ const deployConfigForceCommit = async (fileName: string, silent: boolean) => {
   const { bundleId } = await hre.run('chugsplash-commit', {
     configPath,
     ipfsUrl: '',
-    commitToIpfs: false,
+    commitToIpfs: true,
```
RPate97, line 60: I found the dev script did not actually commit to IPFS until I flipped this to true.

sam-goldman (approved):

PR #193, opened 2022-11-14
fix(ex): run the executor using the hre executable

sam-goldman:
This PR fixes the bug that causes Etherscan verification to fail. It does this by changing the executor's executable from ts-node to hardhat, which injects the correct HRE into the environment.

### DIFF
--- a/packages/executor/.env.example
+++ b/packages/executor/.env.example
@@ -6,4 +6,4 @@ IPFS_API_KEY_SECRET=<ipfs api key to retrieve config file>
 OPT_ETHERSCAN_API_KEY=<Optimism Etherscan API key>
 ETH_ETHERSCAN_API_KEY=<Ethereum Etherscan API key>
 INFURA_API_KEY=<Infura Project API Key>
-NETWORK_NAME=<Target network. Must match a network name in the Hardhat config (e.g. 'optimism-goerli')>
+HARDHAT_NETWORK=<Target hardhat network, used for verification>
--- a/packages/executor/.env.example
+++ b/packages/executor/.env.example
@@ -6,4 +6,4 @@ IPFS_API_KEY_SECRET=<ipfs api key to retrieve config file>
 OPT_ETHERSCAN_API_KEY=<Optimism Etherscan API key>
 ETH_ETHERSCAN_API_KEY=<Ethereum Etherscan API key>
 INFURA_API_KEY=<Infura Project API Key>
-HARDHAT_NETWORK=<Target hardhat network, used for verification>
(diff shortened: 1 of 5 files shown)


packages/executor/.env.example
```diff
@@ -6,4 +6,4 @@ IPFS_API_KEY_SECRET=<ipfs api key to retrieve config file>
 OPT_ETHERSCAN_API_KEY=<Optimism Etherscan API key>
 ETH_ETHERSCAN_API_KEY=<Ethereum Etherscan API key>
 INFURA_API_KEY=<Infura Project API Key>
-HARDHAT_NETWORK=<Target hardhat network, used for verification>
+NETWORK_NAME=<Target network. Must match a network name in the Hardhat config (e.g. 'optimism-goerli')>
```
sam-goldman, line 9: We need to inject this variable into the Dockerfile (see other comment)

packages/executor/Dockerfile
```diff
@@ -6,4 +6,4 @@ RUN npm install
 
 COPY . .
 
-CMD [ "yarn", "start" ]
+CMD [ "yarn", "start", "--network", "TODO"]
```
sam-goldman, line 9: We need to replace "TODO" with a the `NETWORK_NAME` env variable. Do you know how to do this? @RPate97

packages/executor/package.json
```diff
@@ -8,7 +8,7 @@
     "dist/*"
   ],
   "scripts": {
-    "start": "ts-node ./src/index.ts",
+    "start": "hardhat run ./src/index.ts",
```
sam-goldman, line 11: This is the same as running ts-node, except the HRE is also injected into the environment, allowing us to use it for verification

packages/executor/Dockerfile
RPate97: So I thought, this could be acheived just by setting the `HARDHAT_NETWORK` environment variable as described in [this stackoverflow post](https://ethereum.stackexchange.com/questions/83656/where-does-the-line-blur-between-a-task-and-a-script-in-hardhat/93433#93433). I was thinking this would allow us to just specify the `HARDHAT_NETWORK` in the .env file passed to docker run. 

If using `HARDHAT_NETWORK` doesn't work, we can use the [--build-arg](https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg) option. Be aware that --build-arg is passed into the docker build command and the output container will be configured with the input argument. This means we'll need to build and maintain separate images for each network we're supporting which isn't ideal IMO.

sam-goldman: Ah didn't know we could do that with the env variable! Confirmed that we can verify contracts with it. Changed everything back to the way it was.

PR #214, opened 2022-11-16
Fix: Minor Build Info Parsing Bug

RPate97:
## Purpose
Fixes a minor issue where the build info metadata is parsed as if it is JSON when it is not.

### DIFF
--- a/.changeset/four-drinks-peel.md
+++ b/.changeset/four-drinks-peel.md
@@ -1,0 +1,5 @@
+---
+'@chugsplash/plugins': patch
+---
+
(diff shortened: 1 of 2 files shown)


sam-goldman:
It appears `metadata` is sometimes a string and sometimes an object (e.g. it's a string in /demo). I think we can fix this with:
```
const { devdoc, userdoc } = typeof metadata === 'string' ? JSON.parse(metadata).output : metadata.output
```

sam-goldman (approved):

PR #238, opened 2022-11-20
Use executor when deploying against local networks

RPate97:
## Purpose
- Refactors compilation, execution, and verification logic to fully remove dependency on the Hardhat Runtime Environment
- Moves execution logic to the core package
- Updates the executor so that options can be passed in when running it programmatically
- Updates the executor so that the CanonicalChugSplashConfig can be passed in programmatically to avoid needing to upload the config to IPFS during local deployments. 
- Updates the executor so that it can be compiled and run without `ts-node`
- Updates the ChugSplash plugin to use the executor when deploying against local networks

## Notes
- Deploying against remote networks still requires that the executor be run as a standalone process
- This PR introduces some new environment variable requirements for the executor. See `.env.example` in the executor package to ensure you have the correct variables set. 
- You can now define the log level you want for the executor using the `CHUGSPLASH_EXECUTOR__LOG_LEVEL` when running as a standalone process. When the executor is used for local deployments, the log level is set based on the `--silent` flag

## Testing & Review
Unfortunately, this end up being a pretty chunky PR. I recommend focusing mostly on testing various deployment configurations to ensure that it's working properly. 

I had to deal with quite a few merge conflicts while working on this PR. So I also recommend checking parts of this PR that effect things you've worked on recently to ensure your most recent changes were incorporated properly.


packages/executor/src/utils/etherscan.ts
```diff
@@ -48,24 +51,24 @@ export interface EtherscanResponseBody {
...
 export const verifyChugSplashConfig = async (
-  hre: HardhatRuntimeEnvironment,
-  configUri: string
+  configUri: string,
+  provider: ethers.providers.Provider,
+  networkName: string
 ) => {
-  const { etherscanApiKey, etherscanApiEndpoints } = await getEtherscanInfo(hre)
+  const { etherscanApiKey, etherscanApiEndpoints } = await getEtherscanInfo(
+    provider as any, // TODO - figure out how to get the types for this to work correctly
```
sam-goldman, line 59: You should be able to do ethers.providers.JsonRpcProvider here

packages/executor/src/utils/etherscan.ts
```diff
@@ -1,20 +1,16 @@
 import assert from 'assert'
 
-import { Contract } from 'ethers'
+import { Contract, ethers } from 'ethers'
```
sam-goldman, line 3: It's redundant to import `Contract` and `ethers` here, since `Contract` exists as a member on the `ethers` object. (The same applies for basically everything else in the `ethers` package)

packages/executor/src/utils/etherscan.ts
```diff
@@ -48,24 +51,24 @@ export interface EtherscanResponseBody {
...
-  const { etherscanApiKey, etherscanApiEndpoints } = await getEtherscanInfo(hre)
+  const { etherscanApiKey, etherscanApiEndpoints } = await getEtherscanInfo(
+    provider as any, // TODO - figure out how to get the types for this to work correctly
+    networkName
+  )
 
-  const canonicalConfig = await hre.run(TASK_CHUGSPLASH_FETCH, {
-    configUri,
-  })
+  const canonicalConfig = await await chugsplashFetchSubtask({ configUri })
```
sam-goldman, line 63: Looks like there's a redundant `await` here

packages/executor/src/utils/etherscan.ts
```diff
@@ -116,8 +118,8 @@ export const verifyChugSplashConfig = async (
     // Verify the implementation
     try {
       await attemptVerification(
-        hre.network.provider,
-        hre.network.name,
+        provider as any, // TODO - figure out how to get the types for this to work correctly
```
sam-goldman, line 121: > You should be able to do ethers.providers.JsonRpcProvider here

Same here

packages/executor/src/utils/etherscan.ts
RPate97: So that works for the functions we define. However, when we eventually pass the provider into `retrieveContractBytecode` and `getEtherscanEndpoints`, there's a type error because the required provider type is `EthereumProvider`. I'm unsure how to resolve that without casting to any since we don't control the typing of those functions.

packages/plugins/src/hardhat/deployments.ts
```diff
@@ -200,30 +201,41 @@ export const deployChugSplashConfig = async (
...
+
+  // If executing locally, complete post execution actions
+  if (!remoteExecution) {
+    const finalTxnHash = await getFinalDeploymentTxnHash(
+      ChugSplashManager,
+      bundleId
     )
-    await claimExecutorPayment(signer, ChugSplashManager)
+    await postExecutionActions(provider, parsedConfig)
+    await createDeploymentArtifacts(hre, parsedConfig, finalTxnHash)
```
sam-goldman, line 235: We already call `postExecutionActions` and `createDeploymentArtifacts` inside `monitorTask` above (on [these lines](https://github.com/chugsplash/chugsplash/blob/baa98515862f072791e7ee8e688697f96822cadc/packages/plugins/src/hardhat/tasks.ts#L869-L875)

packages/core/src/actions/execute.ts
```diff
@@ -0,0 +1,115 @@
+import { EXECUTOR_BOND_AMOUNT } from '@chugsplash/contracts'
+import { Contract, ethers } from 'ethers'
```
sam-goldman, line 2: Same redundancy here

packages/executor/src/utils/etherscan.ts
sam-goldman: Ah I see. Does it break anything if we set the type to be `EthereumProvider` instead?

RPate97: Yeah we make the type `EthereumProvider` everywhere and that works for running the executor from within the CLI (because the provider that's passed in is of type `EthereumProvider`). However, we run into a problem when running the executor as a standalone process because we need to be able to create the provider from a URL. It seems like hardhat only exports an interface for the `EthereumProvider` which is not sufficient for constructing one from scratch.

packages/plugins/src/hardhat/deployments.ts
```diff
@@ -200,30 +201,41 @@ export const deployChugSplashConfig = async (
...
+  // If executing locally, then startup executor with HRE provider and pass in canonical config
+  if (!remoteExecution) {
+    const executor = new ChugSplashExecutor()
+    executor.init()
+    executor.main(
       {
-        configPath,
-        silent: true,
+        privateKey:
+          '0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80',
```
sam-goldman, line 210: I think we should send ETH from the current signer to the account that corresponds to the `0xac...` private key before calling `executor.main`. This is because users can define their own set of accounts on the Hardhat network, in which case the `0xac` account will hold 0 ETH and won't be able to execute the deployment. I think 1 ETH should be sufficient

packages/executor/src/utils/etherscan.ts
sam-goldman: It looks like the EIP1193Bridge does the trick: https://docs.ethers.io/v5/api/experimental/#experimental-eip1193bridge

(The EthereumProvider inherits from the EIP1193Provider)

packages/executor/.env.example
```diff
@@ -1,6 +1,8 @@
 CHUGSPLASH_EXECUTOR__PRIVATE_KEY=<private key to deploy with>
-CHUGSPLASH_EXECUTOR__NETWORK=<network to monitor, http://host.docker.internal:8545 for local node>
+CHUGSPLASH_EXECUTOR__URL=<url of the network to monitor, http://host.docker.internal:8545 for local node>
+CHUGSPLASH_EXECUTOR__NETWORK=<name of the network to monitor, i.e 'hardhat' or 'optimism-goerli'>
```
sam-goldman, line 3: What's the difference between this and `HARDHAT_NETWORK`?

RPate97: `CHUGSPLASH_EXECUTOR__NETWORK` sets the `network` option in the executor. The network option is used to fetch the etherscan verification endpoint.

sam-goldman (approved):

PR #251, opened 2022-11-24
Basic Support for Upgrades

RPate97:
## Purpose
Adds basic support for upgrading contracts via the `chugsplash-upgrade` task and the `isUpgrade` flag on the `chugsplash-deploy` task. 

## Note
There are still some more complex upgrade cases which we will need to handle to make this a really robust feature. 
- This PR does not handle cases where only some contracts in a config file need to be upgraded. If you attempt to upgrade a contract where the implementation code has not changed, you will encounter an error `ChugSplashManager: implementation was not deployed correctly`. This is due to create2 returning a zero address for the implementation address (I believe) due to the generated address already being in use.
- For the same reason, this PR also does not handle cases where only ChugSplash config variables have changed.
- This PR does not handle cases where only some config files need to be upgraded. I chose to wait on updating the `node`, `test`, and `run` commands to support upgrades due to this. 

We should discuss further how to handle these more complex upgrade cases. I recommend that we take inspiration from Terraform in the design of the upgrade process. This would involve using only the deploy task for both fresh deployments and upgrades. It would also involve dynamically detecting and upgrading only the contracts in a given config that have actually changed since the last deployment.

### DIFF
--- a/packages/core/src/utils.ts
+++ b/packages/core/src/utils.ts
@@ -8,6 +8,7 @@ import {
   ChugSplashManagerABI,
   ChugSplashManagerProxyArtifact,
   CHUGSPLASH_REGISTRY_PROXY_ADDRESS,
+  ProxyABI,
 } from '@chugsplash/contracts'
 
 import { ChugSplashConfig } from './config'
@@ -90,7 +91,9 @@ export const getProxyAddress = (
 
 export const checkValidDeployment = async (
   provider: ethers.providers.Provider,
-  parsedConfig: ChugSplashConfig
+  parsedConfig: ChugSplashConfig,
+  configPath: string,
+  networkName: string
 ) => {
   for (const referenceName of Object.keys(parsedConfig.contracts)) {
     if (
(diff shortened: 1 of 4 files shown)


sam-goldman:
It sounds like there are two questions to answer here:
1. We need to generate a diff of the current config vs what's already deployed, and filter out necessary ChugSplash actions from the bundle before it's committed to IPFS.
2. We may want to merge the chugsplash-upgrade and chugsplash-deploy tasks into a single task.

I'm planning on addressing #1 when i write the chugsplash-dry-run task because the same diffing logic will be applied to both the dry-run task and the upgrade task. I think we can insert this diffing logic [here](https://github.com/chugsplash/chugsplash/blob/develop/packages/plugins/src/hardhat/tasks.ts#L549) for the upgrade task.  A config with no diff (i.e. inside `deployAllChugSplashConfigs`) will just be skipped.

As for #2, I'm personally neutral. Interested to hear how Terraform handles this later.

packages/core/src/utils.ts
```diff
@@ -101,14 +104,84 @@ export const checkValidDeployment = async (
...
-in a previous deployment for this project. You must change this reference name to something other than
-${referenceName} or change the project name to something other than ${parsedConfig.options.projectName}.`
+        `Error: The reference name ${referenceName} inside ${parsedConfig.options.projectName} was already used
+in a previous deployment for this project.
+
+To perform a fresh deployment, you must change this reference name to something other than ${referenceName} or
+change the project name to something other than ${parsedConfig.options.projectName}.
+
+If you wish to upgrade ${parsedConfig.options.projectName} then run to following command:
+npx hardhat chugsplash-upgrade --network ${networkName} ${configPath}`
```
sam-goldman, line 114: I like this!

One nit: I think it's redundant to say `Error: ...` on line 107 since the error message already does automatically. You can see this if you trigger some user-defined error, like calling:
`npx hardhat chugsplash-propose chugsplash/SimpleStorage.config.ts`

packages/plugins/src/hardhat/deployments.ts
```diff
@@ -142,6 +161,22 @@ export const deployChugSplashConfig = async (
...
+    )
+  } else {
+    await checkValidDeployment(
+      hre.ethers.provider,
+      parsedConfig,
+      configPath,
+      hre.network.name
+    )
+  }
+
```
sam-goldman, line 179: Did you mean to duplicate these calls here and a little above on line 112?

packages/plugins/src/hardhat/deployments.ts
```diff
@@ -332,8 +367,6 @@ export const proposeChugSplashBundle = async (
   remoteExecution: boolean,
   ipfsUrl: string
 ) => {
-  await checkValidDeployment(hre.ethers.provider, parsedConfig)
-
```
sam-goldman, line 336: I think we need to put the `checkValidDeployment` / `checkValidUpgrade` calls here instead of in `deployChugSplashConfig` because the latter function is only called in the chugsplash-deploy command, whereas `proposeChugSplashConfig` is called in the chugsplash-propose command as well. I suppose we could also put the logic in `chugsplashCommitTask`, but this way it's only called once per deployment in the proposal step.

packages/plugins/src/hardhat/tasks.ts
```diff
@@ -210,10 +213,37 @@ task(TASK_CHUGSPLASH_DEPLOY)
     'ipfsUrl',
     'Optional IPFS gateway URL for publishing ChugSplash projects to IPFS.'
   )
+  .addFlag('isUpgrade', 'Perform an upgrade on the chosen project')
```
sam-goldman, line 216: I actually don't think this flag should be shown to the user. I.e. ChugSplash should automatically determine if it's a deployment/upgrade

sam-goldman (approved):

PR #254, opened 2022-11-28
Merge deploy and upgrade tasks

RPate97:
## Purpose
Merges the deployment and upgrade tasks. I did this by updating the `checkValidDeployment` function to instead check if an upgrade is happening or not. If an upgrade is happening, then we check if the it is valid and confirm that the user intends to perform an upgrade of their project before continuing with the deployment.

### DIFF
--- a/packages/core/src/utils.ts
+++ b/packages/core/src/utils.ts
@@ -89,12 +89,10 @@ export const getProxyAddress = (
   )
 }
 
-export const checkValidDeployment = async (
+export const checkIsUpgrade = async (
   provider: ethers.providers.Provider,
-  parsedConfig: ChugSplashConfig,
-  configPath: string,
-  networkName: string
-) => {
+  parsedConfig: ChugSplashConfig
+): Promise<boolean | string> => {
   for (const referenceName of Object.keys(parsedConfig.contracts)) {
     if (
       await isProxyDeployed(
@@ -103,18 +101,10 @@ export const checkValidDeployment = async (
         referenceName
       )
     ) {
-      throw new Error(
-        `The reference name ${referenceName} inside ${parsedConfig.options.projectName} was already used
-in a previous deployment for this project.
-
-To perform a fresh deployment, you must change this reference name to something other than ${referenceName} or
-change the project name to something other than ${parsedConfig.options.projectName}.
-
-If you wish to upgrade ${parsedConfig.options.projectName} then run to following command:
-npx hardhat chugsplash-upgrade --network ${networkName} ${configPath}`
-      )
+      return referenceName
     }
   }
+  return false
 }
 
 export const checkValidUpgrade = async (
--- a/packages/plugins/package.json
+++ b/packages/plugins/package.json
@@ -41,7 +41,7 @@
     "node-fetch": "^2.6.7",
     "ora": "^5.4.1",
     "semver": "^7.3.7",
-    "solc": "^0.8.17"
+    "yesno": "^0.4.0"
   },
   "devDependencies": {
     "@nomiclabs/hardhat-ethers": "^2.2.1",
--- a/packages/plugins/src/hardhat/deployments.ts
+++ b/packages/plugins/src/hardhat/deployments.ts
@@ -1,7 +1,8 @@
-import * as path from 'path'
-import * as fs from 'fs'
-
 import '@nomiclabs/hardhat-ethers'
+import * as fs from 'fs'
+import * as path from 'path'
+
+import yesno from 'yesno'
 import { ethers } from 'ethers'
 import {
   ChugSplashConfig,
@@ -16,7 +17,7 @@ import {
   computeBundleId,
   getChugSplashManager,
   getExecutionAmountToSendPlusBuffer,
-  checkValidDeployment,
+  checkIsUpgrade,
   checkValidUpgrade,
 } from '@chugsplash/core'
 import { getChainId } from '@eth-optimism/core-utils'
@@ -49,6 +50,7 @@ export const deployAllChugSplashConfigs = async (
   silent: boolean,
   ipfsUrl: string,
   noCompile: boolean,
+  confirm: boolean,
   spinner: ora.Ora = ora({ isSilent: true })
 ) => {
   const remoteExecution = (await getChainId(hre.ethers.provider)) !== 31337
@@ -73,7 +75,7 @@ export const deployAllChugSplashConfigs = async (
       remoteExecution,
       ipfsUrl,
       noCompile,
-      false,
+      confirm,
       executor,
       spinner
     )
@@ -87,7 +89,7 @@ export const deployChugSplashConfig = async (
   remoteExecution: boolean,
   ipfsUrl: string,
   noCompile: boolean,
-  isUpgrade: boolean,
+  confirm: boolean,
   executor?: ChugSplashExecutor,
   spinner: ora.Ora = ora({ isSilent: true })
 ) => {
@@ -138,7 +140,7 @@ export const deployChugSplashConfig = async (
       hre
     )
 
-  spinner.start(`Committing ${parsedConfig.options.projectName}...`)
+  spinner.start(`Checking the status of ${parsedConfig.options.projectName}...`)
 
   const ChugSplashManager = getChugSplashManager(
     signer,
@@ -171,6 +173,9 @@ export const deployChugSplashConfig = async (
   }
 
   if (currBundleStatus === ChugSplashBundleStatus.EMPTY) {
+    spinner.succeed(
+      `${parsedConfig.options.projectName} has not been proposed before.`
+    )
     await proposeChugSplashBundle(
       hre,
       parsedConfig,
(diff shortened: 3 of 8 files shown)


packages/plugins/package.json
```diff
@@ -41,7 +41,7 @@
     "node-fetch": "^2.6.7",
     "ora": "^5.4.1",
     "semver": "^7.3.7",
-    "solc": "^0.8.17"
+    "yesno": "^0.4.0"
```
RPate97, line 44: Added a convenient utility package for handling confirmations. 
https://www.npmjs.com/package/yesno

Also Solc should not be in the plugins package so I've removed it.

packages/core/src/utils.ts
```diff
@@ -89,11 +89,9 @@ export const getProxyAddress = (
...
 }
 
-export const checkValidDeployment = async (
+export const checkIsUpgrade = async (
   provider: ethers.providers.Provider,
-  parsedConfig: ChugSplashConfig,
-  configPath: string,
-  networkName: string
+  parsedConfig: ChugSplashConfig
 ) => {
```
sam-goldman, line 95: nit: Can you include the return type as `Promise<boolean>`?

packages/plugins/src/hardhat/deployments.ts
```diff
@@ -336,23 +335,32 @@ export const proposeChugSplashBundle = async (
...
-      hre.ethers.provider,
-      parsedConfig,
-      configPath,
-      hre.network.name
-    )
+
+    // Confirm upgrade with user
+    const confirm = await yesno({
+      question: `Prior deployments detected for project ${parsedConfig.options.projectName}, would you like to perform an upgrade? (y/n)`,
+    })
```
sam-goldman, line 356: I like the idea of confirming with the user before upgrading, but this confirmation could be troublesome if performing an upgrade in a CI process. Perhaps we add a `--confirm` flag that allows the user to skip this?

packages/plugins/src/hardhat/deployments.ts
```diff
@@ -336,23 +335,32 @@ export const proposeChugSplashBundle = async (
...
+
+    if (!confirm) {
+      throw new Error(
+        `The project name ${parsedConfig.options.projectName} was already used in a previous deployment. To perform a fresh deployment
+of a new project, you must change the project name to something other than ${parsedConfig.options.projectName}.`
+      )
```
sam-goldman, line 362: Could you add something to the beginning of the error message to affirm that the user denied the upgrade? E.g: `User denied upgrade. The project name ${projectName} was already used...`

packages/plugins/src/hardhat/deployments.ts
RPate97: Yeah, I can add that. Though I do think we'll ultimately want a separate task specifically intended for use in CI processes.

packages/plugins/src/hardhat/deployments.ts
```diff
@@ -336,23 +339,34 @@ export const proposeChugSplashBundle = async (
...
-      hre.ethers.provider,
-      parsedConfig,
-      configPath,
-      hre.network.name
-    )
+
+    if (!confirm) {
+      // Confirm upgrade with user
+      const userConfirmed = await yesno({
+        question: `Prior deployments detected for project ${parsedConfig.options.projectName}, would you like to perform an upgrade? (y/n)`,
```
sam-goldman, line 361: nit: Can you say `Prior deployment(s)` instead? In case there's just one contract that's being upgraded

packages/plugins/src/hardhat/deployments.ts
```diff
@@ -336,23 +339,34 @@ export const proposeChugSplashBundle = async (
...
+      })
+      if (!userConfirmed) {
+        throw new Error(
+          `User denied upgrade. The project name ${parsedConfig.options.projectName} was already used in a previous deployment. To perform a
+  fresh deployment of a new project, you must change the project name to something other than ${parsedConfig.options.projectName}.`
+        )
```
sam-goldman, line 367: The original error message inside `checkValidDeployment` [here](https://github.com/chugsplash/chugsplash/blob/develop/packages/core/src/utils.ts#L106-L114) also suggested that the user could change the reference name to something other than ${referenceName}. Would mind adding that in?

packages/plugins/src/hardhat/tasks.ts
```diff
@@ -1059,9 +1044,14 @@ task(TASK_NODE)
 
 task(TASK_TEST)
   .addFlag('show', 'Show ChugSplash deployment information')
+  .addFlag('confirm', 'Automatically confirm contract upgrades')
```
sam-goldman, line 1047: Imo, this flag should only be necessary if the user is on a live network. I'd personally find it annoying to include a `--confirm` flag whenever I'm using this command to test an upgrade on the Hardhat network. A solution to this would be:
1. Change this flag description to be something like: `Automatically confirm contract upgrades. Only applicable if upgrading on a live network.`
2. Inside this task, set `confirm` to `true` if we're on the Hardhat network.

In the future, we could use the CI deployment task instead of this solution, but I think this is sufficient for now.

packages/plugins/src/hardhat/tasks.ts
```diff
@@ -1100,16 +1090,18 @@ task(TASK_RUN)
     'deployAll',
     'Deploy all ChugSplash configs before executing your script.'
   )
+  .addFlag('confirm', 'Automatically confirm contract upgrades')
```
sam-goldman, line 1093: > Imo, this flag should only be necessary if the user is on a live network. I'd personally find it annoying to include a `--confirm` flag whenever I'm using this command to test an upgrade on the Hardhat network. A solution to this would be:
> 
> 1. Change this flag description to be something like: `Automatically confirm contract upgrades. Only applicable if upgrading on a live network.`
> 2. Inside this task, set `confirm` to `true` if we're on the Hardhat network.
> 
> In the future, we could use the CI deployment task instead of this solution, but I think this is sufficient for now.

Same for this task

packages/plugins/src/hardhat/tasks.ts
```diff
@@ -1020,6 +990,7 @@ task(TASK_NODE)
     "Completely disable all of ChugSplash's activity."
   )
   .addFlag('hide', "Hide all of ChugSplash's output")
+  .addFlag('confirm', 'Automatically confirm contract upgrades')
```
sam-goldman, line 993: Since the `node` task always runs on the Hardhat network, we can remove the 'confirm' flag and let `confirm` be `true` in the call to `deployAllChugSplashConfigs`.

packages/plugins/src/hardhat/deployments.ts
```diff
@@ -336,23 +339,34 @@ export const proposeChugSplashBundle = async (
   configUri: string,
   remoteExecution: boolean,
   ipfsUrl: string,
-  isUpgrade: boolean,
-  configPath: string
+  configPath: string,
+  spinner: ora.Ora,
```
sam-goldman, line 343: Including this parameter requires us to create a `spinner` instance each time we call this `proposeChugSplashBundle`, which isn't ideal if we aren't interested in logging information to the user. I got around this issue in `deployAllChugSplashConfigs` by setting this parameter to `spinner: ora.Ora = ora({ isSilent: true })`, which I think we should do here too.

packages/plugins/src/hardhat/deployments.ts
```diff
@@ -336,23 +339,34 @@ export const proposeChugSplashBundle = async (
...
+  confirm: boolean
 ) => {
+  // Determine if the deployment is an upgrade
+  const isUpgrade = await checkIsUpgrade(hre.ethers.provider, parsedConfig)
   if (isUpgrade) {
+    spinner.succeed('Upgrade detected')
```
sam-goldman, line 349: I like that you use the spinner to inform the user whether the project is a deployment or upgrade, but there's a slight issue. In general, if you call `spinner.succeed` without calling `spinner.start` shortly after, there will be a pause in the spinner output until the next `spinner.start`. In this particular case, the pause might be pretty long since we commit the bundle to IPFS after this line. From the user's perspective, this could make it seem that the process is frozen when in fact it's not. In general, you can avoid this issue by calling `spinner.start` very shortly after each `spinner.succeed`. To demonstrate this, I made a PR [here]() that fixes this issue, since it was easier than leaving comments.

packages/plugins/src/hardhat/tasks.ts
```diff
@@ -1086,7 +1064,7 @@ task(TASK_TEST)
           if (!noCompile) {
             await cleanThenCompile(hre)
           }
-          await deployAllChugSplashConfigs(hre, !show, '', true)
+          await deployAllChugSplashConfigs(hre, !show, '', true, true)
```
RPate97, line 1067: Since the test task only ever deploys contracts when run against the hardhat network, I just removed the confirm flag like the `node` task.

packages/plugins/src/hardhat/tasks.ts
RPate97: Since the test task only ever deploys contracts when run against the hardhat network, I ended up just removing the confirm flag like the node task.

packages/plugins/src/hardhat/tasks.ts
sam-goldman: Looks like you might've forgotten to do #1

RPate97: Huh, seems like the fix for #1 might have been reversed when I rebased didn't realize that could happen.

sam-goldman (approved):

PR #255, opened 2022-11-28
spinner logic for upgrades

sam-goldman:
Adds spinner output for upgrades as per [this comment](https://github.com/chugsplash/chugsplash/pull/254/files#r1033981244).


sam-goldman:
As you can see, it's kind of a hassle working with the spinner. i like that we use it to inform the user whether the project is a deployment or upgrade, but i'm personally going to use it pretty sparingly from now on, especially since we already output a lot of info to the user.

PR #289, opened 2022-12-05
Transfer Ownership and Claim Ownership Tasks

RPate97:
## Purpose
- Adds a task to transfer ownership of a proxy to ChugSplash
- Adds a task to claim ownership of a proxy back from ChugSplash. I initially added this to help with testing, but then decided to keep it because it seems nice to have. 

## Note
- I primarily tested this by deploying a contract with ChugSplash, then claiming ownership of the proxy with my personal account, then transferring it back to ChugSplash. 
- I tested the proxy validation logic by running this against implementation contracts which I know to be non compliant. 
- It would be good to also check this against real non standard proxy contracts.


packages/plugins/src/hardhat/tasks.ts
```diff
@@ -1352,3 +1355,171 @@ export const listProjectsTask = async ({}, hre: HardhatRuntimeEnvironment) => {
...
+    false
+  ) {
+    errorProjectNotRegistered(
+      await getChainId(hre.ethers.provider),
+      hre.network.name,
+      configPath
+    )
+  }
+
+  spinner.succeed('Process registration detected')
```
sam-goldman, line 1444: Do you mean `Project registration detected`?

packages/plugins/src/hardhat/tasks.ts
```diff
@@ -1352,3 +1355,171 @@ export const listProjectsTask = async ({}, hre: HardhatRuntimeEnvironment) => {
...
+  Currently ChugSplash only supports proxies that implement EIP-1967 which yours does not appear to do.
+  If you believe this is a mistake, please reach out to the developers or open an issue on GitHub.`
+
+  // Fetch proxy bytecode and check if it contains the expected EIP-1967 function definitions
+  const iface = new ethers.utils.Interface([
+    'function implementation() external view returns (address)',
+    'function admin() returns (address)',
+    'function upgradeTo(address)',
+    'function changeAdmin(address newAdmin)',
+  ])
```
sam-goldman, line 1457: Instead of hard-coding the interface here, you can just import `ProxyABI` from `@chugsplash/contracts` and do:
```
const iface = new ethers.utils.Interface(ProxyABI)
```

packages/plugins/src/hardhat/tasks.ts
```diff
@@ -1352,3 +1355,171 @@ export const listProjectsTask = async ({}, hre: HardhatRuntimeEnvironment) => {
 task(TASK_CHUGSPLASH_LIST_PROJECTS)
   .setDescription('Lists all projects that are owned by the caller.')
   .setAction(listProjectsTask)
+
+const claimOwnershipTask = async (
```
sam-goldman, line 1359: Could you export this function and `transferOwnershipTask`?

packages/plugins/src/hardhat/tasks.ts
```diff
@@ -1352,3 +1355,171 @@ export const listProjectsTask = async ({}, hre: HardhatRuntimeEnvironment) => {
...
+    }
+  }
+
+  // Fetch proxy owner address from storage slot defined by EIP-1967
+  const ownerAddress = BigNumber.from(
+    await provider.getStorageAt(
+      proxyAddress,
+      '0xb53127684a568b3173ae13b9f8a6016e243e63b6e8ee1178d6a717850b5d6103'
+    )
+  ).toHexString()
```
sam-goldman, line 1473: The `toHexString` function truncates leading zeros, which can cause valid owner addresses to error on the subsequent call to `ethers.utils.isAddress`. For example, say the address `0x0011111111111111111111111111111111111111` is a valid owner address. Calling `toHexString` will truncate the address and return `0x11111111111111111111111111111111111111` (which is one byte shorter than an address). Instead, what we should do is use ethers' [decode](https://docs.ethers.io/v5/api/utils/abi/coder/#AbiCoder-decode) function.

packages/plugins/src/hardhat/tasks.ts
```diff
@@ -1352,3 +1355,171 @@ export const listProjectsTask = async ({}, hre: HardhatRuntimeEnvironment) => {
...
+
+  // Fetch ChugSplashManager address for this project
+  const managerAddress = await getChugSplashManagerProxyAddress(
+    parsedConfig.options.projectName
+  )
```
sam-goldman, line 1478: The `getChugSplashManagerProxyAddress` is synchronous, so no need for `await` here

packages/plugins/src/hardhat/tasks.ts
```diff
@@ -1352,3 +1355,171 @@ export const listProjectsTask = async ({}, hre: HardhatRuntimeEnvironment) => {
...
+  if (ownerAddress.toLowerCase() !== signerAddress.toLowerCase()) {
+    throw new Error(`Target proxy is owned by: ${ownerAddress}.
+You attempted to transfer ownership of the proxy using the address: ${signerAddress}`)
+  }
+
+  // Check that the proxy implementation function returns an address
+  const contract = new ethers.Contract(proxyAddress, iface, signer)
+  const implementationAddress = BigNumber.from(
+    await contract.callStatic.implementation()
+  ).toHexString()
```
sam-goldman, line 1501: This can just be:
```
const implementationAddress =  await contract.callStatic.implementation()
```
This'll return the 0x-prefixed address without potentially truncating leading zeros.

packages/plugins/src/hardhat/tasks.ts
```diff
@@ -1352,3 +1355,171 @@ export const listProjectsTask = async ({}, hre: HardhatRuntimeEnvironment) => {
...
+  if (!ethers.utils.isAddress(implementationAddress)) {
+    throw new Error(incompatibleProxyError)
+  }
+
+  spinner.succeed('Proxy compatibility verified')
+  spinner.start('Transferring proxy ownership to ChugSplash...')
+
+  // Transfer ownership of the proxy to the ChugSplashManager.
+  await contract.changeAdmin(managerAddress)
```
sam-goldman, line 1510: Any state-changing transaction should be wrapped in `await (...).wait()`. So, this should be:
```
await (await contract.changeAdmin(managerAddress)).wait()
```

packages/plugins/src/hardhat/tasks.ts
```diff
@@ -90,6 +90,9 @@ export const TASK_CHUGSPLASH_MONITOR = 'chugsplash-monitor'
 export const TASK_CHUGSPLASH_CANCEL = 'chugsplash-cancel'
 export const TASK_CHUGSPLASH_WITHDRAW = 'chugsplash-withdraw'
 export const TASK_CHUGSPLASH_LIST_PROJECTS = 'chugsplash-list-projects'
+export const TASK_CHUGSPLASH_TRANSFER_OWNERSHIP =
+  'chugsplash-transfer-ownership'
+export const TASK_CHUGSPLASH_CLAIM_OWNERSHIP = 'chugsplash-claim-ownership'
```
sam-goldman, line 95: What do you think about changing this task name to `chugsplash-claim-proxies`? I feel like "claim ownership" could either mean claiming ownership of the ChugSplashManager or claiming ownership of the proxies.

packages/plugins/src/hardhat/tasks.ts
```diff
@@ -1352,3 +1355,171 @@ export const listProjectsTask = async ({}, hre: HardhatRuntimeEnvironment) => {
...
+    errorProjectNotRegistered(
+      await getChainId(hre.ethers.provider),
+      hre.network.name,
+      configPath
+    )
+  }
+
+  spinner.start('Claiming proxy ownership...')
+
+  const manager = await getChugSplashManager(
```
sam-goldman, line 1390: No need for `await` here since `getChugSplashManager` is synchronous

packages/plugins/src/hardhat/tasks.ts
```diff
@@ -1352,3 +1355,171 @@ export const listProjectsTask = async ({}, hre: HardhatRuntimeEnvironment) => {
...
+    (await isProjectRegistered(signer, parsedConfig.options.projectName)) ===
+    false
+  ) {
+    errorProjectNotRegistered(
+      await getChainId(hre.ethers.provider),
+      hre.network.name,
+      configPath
+    )
+  }
+
```
sam-goldman, line 1387: We should check for a couple more things here:
1. We should throw an error if the caller is not the project owner (via `getProjectOwnerAddress` in the core package)
2. We should throw an error if there is currently an active bundle (via `manager.activeBundleId()`)

packages/plugins/src/hardhat/tasks.ts
```diff
@@ -1352,3 +1355,171 @@ export const listProjectsTask = async ({}, hre: HardhatRuntimeEnvironment) => {
 task(TASK_CHUGSPLASH_LIST_PROJECTS)
   .setDescription('Lists all projects that are owned by the caller.')
   .setAction(listProjectsTask)
+
+const claimOwnershipTask = async (
+  args: {
+    configPath: string
+    referenceName: string
```
sam-goldman, line 1362: Instead of passing in a `referenceName`, what do you think about allowing the owner to claim ownership of all proxies in a single call by default? Imo this is preferable since I can't really imagine a case where an owner would want them transferred one at a time.

packages/plugins/src/hardhat/tasks.ts
```diff
@@ -1352,3 +1355,171 @@ export const listProjectsTask = async ({}, hre: HardhatRuntimeEnvironment) => {
...
+    silent: boolean
+  },
+  hre
```
sam-goldman, line 1365: We can set this type to be `HardhatRuntimeEnvironment`. Same with the transferOwnershipTask

packages/plugins/src/hardhat/tasks.ts
```diff
@@ -1352,3 +1355,171 @@ export const listProjectsTask = async ({}, hre: HardhatRuntimeEnvironment) => {
...
+    false
+  ) {
+    errorProjectNotRegistered(
+      await getChainId(hre.ethers.provider),
+      hre.network.name,
+      configPath
+    )
+  }
+
+  spinner.start('Claiming proxy ownership...')
```
sam-goldman, line 1388: I think you're missing a `spinner.succeed` call before this `spinner.start`

sam-goldman:
Looks like you can resolve [this TODO](https://github.com/chugsplash/chugsplash/blob/develop/packages/core/src/utils.ts#L164) from a couple weeks ago

sam-goldman:
What do you think about following [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) for commit messages?

packages/plugins/src/hardhat/tasks.ts
RPate97: I think that's a good quality of life improvement. However, given that this task is already somewhat outside the scope of the original spec for this feature, I think we should avoid adding additional scope to it. 

I'd say we should put that idea on the to-do list and come back to it if we find that this is a commonly used feature. For now, I've just updated the task name to more clearly reflect what it does.

packages/plugins/src/hardhat/tasks.ts
sam-goldman: Looks like it wasn't added to the `transferOwnershipTask`

packages/core/src/utils.ts
```diff
@@ -161,7 +161,6 @@ npx hardhat chugsplash-deploy --network ${networkName} ${configPath}
   }
 
   if (requiresOwnershipTransfer.length > 0) {
-    // TODO update this once the transfer ownership task is implemented
     throw new Error(
       `Error: Detected proxy contracts which are not managed by ChugSplash.
       ${requiresOwnershipTransfer.map(
```
sam-goldman, line 166: Could you fill out the task on line 171 of this file? It's currently `npx hardhat chugsplash-transfer-ownership>`. Something like:
```
npx hardhat chugsplash-transfer-ownership --network ${hre.network.name} --proxy-address <address> ${configPath}
```

packages/plugins/src/hardhat/tasks.ts
```diff
@@ -1375,3 +1378,182 @@ export const listProjectsTask = async ({}, hre: HardhatRuntimeEnvironment) => {
...
+
+  spinner.succeed(`Proxy ownership claimed by address ${signerAddress}`)
+}
+
+task(TASK_CHUGSPLASH_CLAIM_PROXY)
+  .addPositionalParam(
+    'configPath',
+    'Path to the ChugSplash config file for the project that owns the target contract'
+  )
+  .addPositionalParam(
```
sam-goldman, line 1448: Was just messing around with these tasks and realized that having two positional params is pretty unintuitive. E.g. running `npx hardhat chugsplash-claim-proxy <configPath> <referenceName>` works, but switching their order (i.e. `<configPath> <referenceName>`) doesn't. In all of the Hardhat tasks I've used, there's only one positional param to avoid this. What do you think about changing the reference name to `addFlag`?

packages/plugins/src/hardhat/tasks.ts
```diff
@@ -1375,3 +1378,182 @@ export const listProjectsTask = async ({}, hre: HardhatRuntimeEnvironment) => {
...
+
+  spinner.succeed('Proxy ownership successfully transferred to ChugSplash')
+}
+
+task(TASK_CHUGSPLASH_TRANSFER_OWNERSHIP)
+  .addPositionalParam(
+    'configPath',
+    'Path to the ChugSplash config file for the project that you would like to own the target contract'
+  )
+  .addPositionalParam(
```
sam-goldman, line 1554: > Was just messing around with these tasks and realized that having two positional params is pretty unintuitive. E.g. running `npx hardhat chugsplash-claim-proxy <configPath> <referenceName>` works, but switching their order (i.e. `<referenceName> <configPath>`) doesn't. In all of the Hardhat tasks I've used, there's only one positional param to avoid this. What do you think about changing the reference name to `addFlag`?

Same thing here. We could change the flag to be `proxy` or keep it as `proxyAddress` if you'd prefer

packages/core/src/utils.ts
```diff
@@ -161,15 +161,14 @@ npx hardhat chugsplash-deploy --network ${networkName} ${configPath}
...
         ({ name, address }) => `${name}, ${address}\n`
       )}
 
 To upgrade these contracts, you must first transfer ownership of them to ChugSplash using the following command:
-npx hardhat chugsplash-transfer-ownership>
+npx hardhat chugsplash-transfer-ownership --network ${networkName} ${configPath} <proxy address>
```
sam-goldman, line 171: If you agree with the other comments, this'll need to be updated to reflect the latest change

packages/plugins/src/hardhat/tasks.ts
```diff
@@ -1375,3 +1378,182 @@ export const listProjectsTask = async ({}, hre: HardhatRuntimeEnvironment) => {
...
+
+  await (
+    await manager.transferProxyOwnership(referenceName, signerAddress)
+  ).wait()
+
+  spinner.succeed(`Proxy ownership claimed by address ${signerAddress}`)
+}
+
+task(TASK_CHUGSPLASH_CLAIM_PROXY)
+  .addPositionalParam(
```
sam-goldman, line 1444: Can you add a short description for this task via `setDescription`?

packages/plugins/src/hardhat/tasks.ts
```diff
@@ -1375,3 +1378,182 @@ export const listProjectsTask = async ({}, hre: HardhatRuntimeEnvironment) => {
...
+  spinner.start('Transferring proxy ownership to ChugSplash...')
+
+  // Transfer ownership of the proxy to the ChugSplashManager.
+  await (await contract.changeAdmin(managerAddress)).wait()
+
+  spinner.succeed('Proxy ownership successfully transferred to ChugSplash')
+}
+
+task(TASK_CHUGSPLASH_TRANSFER_OWNERSHIP)
+  .addPositionalParam(
```
sam-goldman, line 1550: > Can you add a short description for this task via `setDescription`?

Same here

packages/core/src/utils.ts
```diff
@@ -161,15 +161,14 @@ npx hardhat chugsplash-deploy --network ${networkName} ${configPath}
...
-    // TODO update this once the transfer ownership task is implemented
     throw new Error(
       `Error: Detected proxy contracts which are not managed by ChugSplash.
       ${requiresOwnershipTransfer.map(
         ({ name, address }) => `${name}, ${address}\n`
       )}
 
 To upgrade these contracts, you must first transfer ownership of them to ChugSplash using the following command:
-npx hardhat chugsplash-transfer-ownership>
+npx hardhat chugsplash-transfer-ownership --network ${networkName} ${configPath} --proxy <proxy address>
```
sam-goldman, line 171: This'll need to be updated to `--config ${configPath}` since it's now a regular param

packages/plugins/src/hardhat/tasks.ts
```diff
@@ -1375,3 +1378,186 @@ export const listProjectsTask = async ({}, hre: HardhatRuntimeEnvironment) => {
...
+
+  spinner.succeed(`Proxy ownership claimed by address ${signerAddress}`)
+}
+
+task(TASK_CHUGSPLASH_CLAIM_PROXY)
+  .setDescription(
+    'Transfers ownership of a proxy from ChugSplash to the caller'
+  )
+  .addPositionalParam(
+    'configPath',
```
sam-goldman, line 1448: We decided to change `addPositionalParam` to `addParam` and `configPath` to `config` for both tasks

sam-goldman (approved):

PR #292, opened 2022-12-08
feat(pg) Add list-proposers and add-proposers tasks

RPate97:
## Purpose
Adds `chugsplash-list-proposers` and `chugsplash-add-proposer`

### DIFF
--- a/packages/plugins/src/hardhat/tasks.ts
+++ b/packages/plugins/src/hardhat/tasks.ts
@@ -18,6 +18,7 @@ import {
   ChugSplashActionBundle,
   ChugSplashBundleState,
   ChugSplashBundleStatus,
+  displayProposerTable,
   registerChugSplashProject,
   getChugSplashRegistry,
   displayDeploymentTable,
@@ -92,6 +93,8 @@ export const TASK_CHUGSPLASH_MONITOR = 'chugsplash-monitor'
 export const TASK_CHUGSPLASH_CANCEL = 'chugsplash-cancel'
 export const TASK_CHUGSPLASH_WITHDRAW = 'chugsplash-withdraw'
 export const TASK_CHUGSPLASH_LIST_PROJECTS = 'chugsplash-list-projects'
+export const TASK_CHUGSPLASH_LIST_PROPOSERS = 'chugsplash-list-proposers'
+export const TASK_CHUGSPLASH_ADD_PROPOSER = 'chugsplash-add-proposers'
 
 subtask(TASK_CHUGSPLASH_FETCH)
   .addParam('configUri', undefined, undefined, types.string)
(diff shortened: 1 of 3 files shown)


sam-goldman:
nit: There should be a colon after `feat(pg)` in the commit message

packages/plugins/src/hardhat/tasks.ts
```diff
@@ -1375,3 +1378,140 @@ export const listProjectsTask = async ({}, hre: HardhatRuntimeEnvironment) => {
...
+
+  if (newProposers.length === 0) {
+    throw new Error('You must specify at least one proposer to add.')
+  }
+
+  const parsedConfig = loadParsedChugSplashConfig(configPath)
+  const provider = hre.ethers.provider
+  const signer = provider.getSigner()
+
+  const spinner = ora()
```
sam-goldman, line 1460: I think there should be a `spinner.start` here

packages/plugins/src/hardhat/tasks.ts
```diff
@@ -1375,3 +1378,140 @@ export const listProjectsTask = async ({}, hre: HardhatRuntimeEnvironment) => {
...
+  // Fetch current owner
+  const projectOwnerAddress = await getProjectOwnerAddress(
+    provider,
+    parsedConfig.options.projectName
+  )
+  if (projectOwnerAddress !== (await signer.getAddress())) {
+    throw new Error(`Project is owned by: ${projectOwnerAddress}.
+You attempted to add a proposer using address: ${await signer.getAddress()}`)
+  }
+
```
sam-goldman, line 1487: If you add a `spinner.start` above, this would be a good place to have a `spinner.succeed`

packages/plugins/src/hardhat/tasks.ts
```diff
@@ -1375,3 +1378,140 @@ export const listProjectsTask = async ({}, hre: HardhatRuntimeEnvironment) => {
...
+    spinner.succeed(`Proposer ${newProposer} successfully added!`)
+  }
+
+  await listProposersTask({ configPath }, hre)
+}
+
+task(TASK_CHUGSPLASH_ADD_PROPOSER)
+  .setDescription('Adds a new proposer to the list of approved proposers')
+  .addPositionalParam(
```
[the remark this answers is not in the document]
RPate97: So I agree with you that it would be good for there to be only one positional parameter per task. However, I don't think it's a good idea for us to include `configPath` and not make it the positional param simply because it is a positional param in all of our other tasks. So it's very unintuitive that it would not be positional in this specific task. 

Unfortunately, I don't think we can make a non-positional and variadic param. So I believe our choices are to either keep it as is or only allow the user to input a single proposer address per task run.

packages/plugins/src/hardhat/tasks.ts
```diff
@@ -1375,3 +1378,140 @@ export const listProjectsTask = async ({}, hre: HardhatRuntimeEnvironment) => {
...
+
+    spinner.succeed(`Proposer ${newProposer} successfully added!`)
+  }
+
+  await listProposersTask({ configPath }, hre)
+}
+
+task(TASK_CHUGSPLASH_ADD_PROPOSER)
+  .setDescription('Adds a new proposer to the list of approved proposers')
+  .addParam('configPath', 'Path to the ChugSplash config file to propose')
```
sam-goldman, line 1511: We decided to change this from `configPath` to `config`

packages/plugins/src/hardhat/tasks.ts
```diff
@@ -1375,3 +1378,140 @@ export const listProjectsTask = async ({}, hre: HardhatRuntimeEnvironment) => {
...
+  }
+
+  // Display the list of proposers
+  displayProposerTable(proposers)
+}
+
+task(TASK_CHUGSPLASH_LIST_PROPOSERS)
+  .setDescription('Lists all of the approved proposers for this project')
+  .addPositionalParam(
+    'configPath',
```
sam-goldman, line 1438: > We decided to change this from `configPath` to `config`

Same here

packages/plugins/src/hardhat/tasks.ts
```diff
@@ -1375,3 +1378,140 @@ export const listProjectsTask = async ({}, hre: HardhatRuntimeEnvironment) => {
...
+    'Path to the ChugSplash config file to propose'
+  )
```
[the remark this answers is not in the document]
RPate97: yes

sam-goldman (approved):

PR #294, opened 2022-12-09
feat(core): smarter batch action execution

smartcontracts:
Updates batch action execution with helper functions that figure stuff out automatically to reduce the number of things that need to be hard-coded. Now also batch executes implementation deployments.

### DIFF
--- a/packages/core/src/actions/execute.ts
+++ b/packages/core/src/actions/execute.ts
@@ -8,6 +8,7 @@ import {
   isSetImplementationAction,
 } from './bundle'
 import {
+  BundledChugSplashAction,
   ChugSplashActionBundle,
   ChugSplashBundleState,
   ChugSplashBundleStatus,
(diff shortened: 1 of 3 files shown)


sam-goldman:
Unfortunately, this new approach actually slows down the time that it takes to execute the SetStorage actions by quite a bit if there are a lot of SetStorage actions in the bundle. I timed the execution of the SetStorage actions before and after this PR for Storage.config.ts in the plugins package (which has 59 SetStorage actions) on the Hardhat network:

before: 10.175s
after: 46.702s

However, for Optimism's devnetL1 deployment, which has 21 SetStorage actions, this PR kept the execution time the same:

Optimism (devnetL1 network):
before: 16.284s
after: 16.465s

On the bright side, there was a significant speed-up for the DeployImplementation actions. I only recorded this for Optimism's deployment since Storage.config.ts only contains one contract.

before: 1:28.793 (m:ss.mmm)
after: 44.630s

So, I think it'd be best to keep the naive solution for the SetStorage actions unless there's some way to modify the binary search algorithm so that it performs less `estimateGas` calls (which I think is the issue). We should def use this PR for the DeployImplementation actions since it speeds things up by ~2x.

packages/core/src/actions/execute.ts
```diff
@@ -73,73 +153,31 @@ export const executeTask = async (args: {
...
-        )
-      ).wait()
-      currActionsExecuted += 1
-      logger.info(
-        `Deployed implementation contract: ${
-          currActionsExecuted - firstDeployImplIndex
-        }/${firstSetImplIndex - firstDeployImplIndex}`
+    logger.info(`executed DeployImplementation actions`)
+
+    // Execute SetImplementation actions in batches.
```
sam-goldman, line 168: nit: We execute the SetImplementation actions in a single transaction (not in batches)

packages/core/src/actions/execute.ts
```diff
@@ -58,10 +58,90 @@ export const executeTask = async (args: {
...
+        throw new Error(
+          'unable to find a batch size that does not exceed the block gas limit'
+        )
+      }
+
+      return min
+    }
 
-    // Execute actions that have not been executed yet.
-    let currActionsExecuted = bundleState.actionsExecuted.toNumber()
```
sam-goldman, line 64: I think removing `currActionsExecuted` will be problematic since it allows us to skip actions that have already been executed. For example, say the executor gets halfway through a deployment then a transaction fails for some reason. Previously, the executor would be able to resume where it left off because of `currActionsExecuted`, but now it'll attempt to execute all of the actions again, including those that have previously been executed.

packages/core/src/actions/execute.ts
```diff
@@ -58,10 +58,90 @@ export const executeTask = async (args: {
...
+    /**
+     * Helper function for executing a list of actions in batches.
+     *
+     * @param actions List of actions to execute.
+     */
+    const executeBatchActions = async (actions: any[]) => {
+      let executed = 0
+      while (executed < actions.length) {
+        // Figure out the maximum number of actions that can be executed in a single batch.
+        const batchSize = await findMaxBatchSize(actions)
```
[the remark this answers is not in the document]
smartcontracts: Yep! Nice catch.

packages/core/src/actions/execute.ts
```diff
@@ -58,88 +59,172 @@ export const executeTask = async (args: {
...
-          action.proof.siblings
-        )
-      ).wait()
-      currActionsExecuted += 1
-      logger.info(
-        `Deployed implementation contract: ${
-          currActionsExecuted - firstDeployImplIndex
-        }/${firstSetImplIndex - firstDeployImplIndex}`
+    // Execute SetStorage actions in batches.
+    logger.info(`executing SetStorage actions`)
```
sam-goldman, line 206: nit: Missing `...` at the end. (To keep things consistent with the next couple log messages

sam-goldman:
This is sick! One minor comment then I'll go ahead and merge it

packages/core/src/actions/execute.ts
smartcontracts: Done!

sam-goldman (approved):

PR #320, opened 2022-12-17
docs: add chugsplash file

sam-goldman:


README.md
```diff
@@ -28,6 +27,15 @@ ChugSplash makes it easy to deploy and upgrade smart contracts **securely**. It'
...
 * Generates deployment artifacts in the same format as hardhat-deploy.
 
+## Documentation
+
+1. [ChugSplash File](https://github.com/chugsplash/chugsplash/blob/develop/docs/chugsplash-file.md): Detailed explanation of the file where you define a deployment or upgrade. Start here.
+2. [Defining Variables in a ChugSplash File](https://github.com/chugsplash/chugsplash/blob/develop/docs/variables.md): Comprehensive reference that explains how to assign values to every variable type in a ChugSplash file.
+
+## Reach out
+
+If you have questions or want to request a feature, join our [Discord channel](https://discord.com/channels/1053048300565188729/1053048301143986219)!
```
RPate97, line 37: This needs to be an actual Discord invite link. You can get one by clicking the server name at the top left and then selected "Invite People".

docs/chugsplash-file.md
```diff
@@ -0,0 +1,62 @@
...
+        },
+        // Other contract definitions:
+        ...
+    }
+}
+```
+
+### Configuration settings
+
+The `options` property in your ChugSplash file contains only a single field, `projectName`. The project name can be any name you choose.
```
RPate97, line 48: Might want to specify that the project name needs to be unique. This is actually a limitation we should probably deal with b/c it will be quite annoying for users.

sam-goldman: We need project names to be unique because it determines the unique address of the ChugSplashManager (which in turn determines the unique addresses of all contracts deployed by it). I don't think this is too much of an issue. E.g. github repos have unique names

sam-goldman: Gonna add this to the techncial overview

PR #322, opened 2022-12-18
feat(core): support bytes/strings (length >31)

RPate97:
## Purpose
Adds support for bytes/strings of length > 31

### DIFF
--- a/packages/core/src/languages/solidity/storage.ts
+++ b/packages/core/src/languages/solidity/storage.ts
@@ -249,7 +249,19 @@ export const encodeVariable = (
         },
       ]
     } else {
-      throw new Error('large strings (>31 bytes) not supported')
+      let slots = [
+        {
+          key: slotKey,
+          val: padHexSlotValue((bytes.length * 2 + 1).toString(16), 0),
+        },
+      ]
+      slots = slots.concat(
+        encodeBytesArrayElements(
+          bytes,
+          utils.keccak256(slotKey) // The slot key of the array elements begins at the hash of the `slotKey`.
+        )
+      )
+      return slots
     }
   } else if (variableType.encoding === 'mapping') {
     // Iterate over every key/value in the mapping to get the storage slot pair for each one.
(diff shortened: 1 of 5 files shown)


packages/plugins/chugsplash/Storage.config.ts
```diff
@@ -10,11 +10,15 @@ const config: UserChugSplashConfig = {
...
+      variables: {
+        ...variables,
+        myStorage: '0x1111111111111111111111111111111111111111',
+      },
     },
     MySimpleStorage: {
-      contract: 'SimpleStorage',
+      contract: 'Storage',
       variables: {
+        ...variables,
```
sam-goldman, line 21: Is there a reason you changed this from `SimpleStorage`? If you don't want to use `SimpleStorage.t.sol`, could you remove it?

RPate97: I actually changed this because `SimpleStorage.t.sol` has been removed already. I'm not sure why though, I was going to ask you about that actually haha.

sam-goldman: Whoops, my bad! Added that back

sam-goldman:
Do you think you could refactor this so that it uses `encodeArrayElements`? I think it'd be possible since long bytes/strings are encoded identically to `bytes1[]`. This encoding logic is probably the most important TypeScript code in ChugSplash, so it'd be great if we could avoid duplication as much as possible

RPate97:
> long bytes/strings are encoded identically to `bytes1[]`

This is actually not true. ["...you can distinguish a short array from a long array by checking if the lowest bit is set: short (not set) and long (set)"](https://docs.soliditylang.org/en/latest/internals/layout_in_storage.html#bytes-and-string). While this can be dealt with, it adds a lot of complexity to the `encodeVariable` function which feels unnecessary to me. 

> Do you think you could refactor this so that it uses `encodeArrayElements`?

So normally I would be very pro-removing duplication and in fact, I actually implemented this feature that way in the first place. However, I was not happy with that implementation because it added a lot of opaque complexity related to the above issue and also to the fact that the `encodeArrayElements` function is designed to work with dynamic arrays which Solidity generates different info on. 

So, I chose to refactor it to use a separate function because I believe the differences in handling this specific case are large enough to warrant it. While this method does introduce some duplication, I feel that it is significantly cleaner and will result in less error-prone code.

packages/core/src/languages/solidity/storage.ts
```diff
@@ -407,6 +419,46 @@ export const encodeArrayElements = (
   return slots
 }
 
+/**
+ * Encodes a bytes/string value of length > 31 bytes as a series of key/value storage slot pairs
+ * using the Solidity storage layout.
+ *
+ * TODO - Refactor long bytes/string encoding to instead use the encodeArrayElements and treat the
+ * variable as a dynamic array of bytes1.
```
sam-goldman, line 427: Minor correction: we'd treat the variable as a fixed-size array of bytes1

sam-goldman (approved):

PR #323, opened 2022-12-18
feat(pg): add chugsplash-init hardhat task

sam-goldman:


packages/plugins/src/sample-project/chugsplash-files.ts
```diff
@@ -0,0 +1,41 @@
...
+        stored: true,
+        storageName: 'First',
+        otherStorage: '0x1111111111111111111111111111111111111111',
+      },
+    },
+  },
+}
+
+export default config
+`
```
RPate97, line 21: I think this feature will be easier to maintain if we move these into their own files so they'll be covered by the linter and can be tested more easily.

sam-goldman: Did this for the sample ChugSplash files. Decided not to do it for the sample contract because we inject the solidity version into the source code, so it's easier to work with a string. I tried to do this for the sample test files but couldn't make it work to because we import `@chugsplash/plugins` in the test files, which isn't possible to do from within the plugins package. So, I think those are best left as strings.

RPate97 (approved):
Looks good, I see you're reasoning on the contract and test files.

PR #325, opened 2022-12-18
docs: add project setup guide

sam-goldman:


README.md
```diff
@@ -29,8 +29,9 @@ ChugSplash makes it easy to deploy and upgrade smart contracts **securely**. It'
 
 ## Documentation
 
-1. [ChugSplash File](https://github.com/chugsplash/chugsplash/blob/develop/docs/chugsplash-file.md): Detailed explanation of the file where you define a deployment or upgrade. Start here.
-2. [Defining Variables in a ChugSplash File](https://github.com/chugsplash/chugsplash/blob/develop/docs/variables.md): Comprehensive reference that explains how to assign values to every variable type in a ChugSplash file.
+1. [Setting up a ChugSplash project](https://github.com/chugsplash/chugsplash/blob/develop/docs/setup-project.md): Take your first steps with ChugSplash.
```
RPate97, line 32: Would be good to also update the Tutorial section to link to this guide as well.

docs/setup-project.md
```diff
@@ -0,0 +1,176 @@
...
+With Yarn:
+```
+yarn add --dev @chugsplash/plugins @chugsplash/core
+```
+With npm:
+```
+npm install --save-dev @chugsplash/plugins @chugsplash/core
+```
+
+## Setup ChugSplash using JavaScript
```
RPate97, line 62: I'm conflicted about duplicating this info for both Javascript and Typescript. I feel like we could get away with just including typescript and then note that the `chugsplash-init` command will also generate javascript depending on your project setup.

README.md
sam-goldman: Was actually going to delete the tutorial since this serves the same purpose with a better UX. There are only two things included in the tutorial that aren't covered in this guide: immutables and a technical overview. Was planning on adding docs for both of these

docs/setup-project.md
sam-goldman: Yeah it's not super elegant but I think it's the best solution for now. We could probably get away with just having a single TS or JS guide, but I think this provides a subpar experience for users who are trying to setup with the language that we haven't explicitly given instructions for. Fwiw Hardhat has a [JS guide](https://hardhat.org/hardhat-runner/docs/guides/project-setup) and a [TS guide](https://hardhat.org/hardhat-runner/docs/guides/typescript), and so does [smock](https://smock.readthedocs.io/en/latest/getting-started.html)

README.md
RPate97: That sounds good to me

PR #329, opened 2022-12-19
docs: remove tutorial

sam-goldman:
Most of the content in the tutorial is now in [Setting up a ChugSplash project](https://github.com/chugsplash/chugsplash/blob/develop/docs/setup-project.md). The rest will go in the technical overview doc, which isn't written yet.


packages/plugins/README.md
```diff
@@ -1,157 +0,0 @@
-# ChugSplash Tutorial
```
RPate97, line 1: Just FYI, removing this file means there will be no readme on NPM. I'm not sure exactly how, but I'm guessing there's probably a way to share the top level readme with the individual packages.

sam-goldman: Great, thanks for catching that!

PR #343, opened 2023-01-04
fix(ex): remove infura as rpc service

sam-goldman:


packages/executor/.env.example
```diff
@@ -5,7 +5,5 @@ CHUGSPLASH_EXECUTOR__AMPLITUDE_KEY=<amplitude api key for analytics (optional)>
 CHUGSPLASH_EXECUTOR__LOG_LEVEL=<filter logs based on this level, i.e 'error', 'warning', 'info'>
 IPFS_PROJECT_ID=<ipfs project id to retrieve config file>
 IPFS_API_KEY_SECRET=<ipfs api key to retrieve config file>
-OPT_ETHERSCAN_API_KEY=<Optimism Etherscan API key>
-ETH_ETHERSCAN_API_KEY=<Ethereum Etherscan API key>
-INFURA_API_KEY=<Infura Project API Key>
```
sam-goldman, line 10: Is it okay to remove `INFURA_API_KEY` here? It doesn't look like it's used in the executor package.

RPate97: Yep good catch

packages/plugins/hardhat.config.ts
```diff
@@ -22,17 +22,14 @@ const config: HardhatUserConfig = {
     },
   },
   networks: {
-    localhost: {
-      url: 'http://localhost:8545',
-    },
```
RPate97, line 27: Why remove this?

RPate97 (approved):
Looks good

packages/plugins/hardhat.config.ts
sam-goldman: `http://localhost:8545` is the default port for the localhost Hardhat network, so there's no need to explicitly define this

PR #344, opened 2023-01-04
fix(ex): fix etherscan verification bug that resulted from incorrect implementation address calculation

sam-goldman:


packages/executor/src/utils/constants.ts
```diff
@@ -3,23 +3,9 @@ export enum ExecutorSelectionStrategy {
...
-    urls: {
-      apiURL: 'https://api-goerli.arbiscan.io/api',
-      browserURL: 'https://goerli.arbiscan.io/',
-    },
-  },
-]
+// Array of custom chains for Etherscan verification. Each array element must be in the following
+// format:
+// https://hardhat.org/hardhat-runner/plugins/nomiclabs-hardhat-etherscan#adding-support-for-other-networks
+export const customChains = []
```
sam-goldman, line 9: The `optimisticGoerli` and `arbitrumGoerli` chains are already defined in `chainConfig` (imported from @nomiclabs/hardhat-etherscan), so I removed them from this array.

PR #374, opened 2023-01-12
fix(pg): write deployment artifacts for proxy and implementation

sam-goldman:
Previously, we were only generating a deployment artifact for each implementation contract. This PR adds deployment artifacts for the proxies.

@RPate97 This can wait until after the foundry refactor if it's not convenient to integrate rn


RPate97:
> Previously, we were only generating a deployment artifact for each implementation contract. This PR adds deployment artifacts for the proxies.
> 
> @RPate97 This can wait until after the foundry refactor if it's not convenient to integrate rn

This will generate a lot of merge conflicts for the `pate/foundry-library` branch if merged now. I think it would be best to rebase this onto the `pate/foundry-library` branch and then merge it into that instead. I can assist with resolving issues related to the rebase.

Any other fixes, changes, improvements, etc we want to make prior to merging the `pate/foundry-library` branch should also be developed by branching off of `pate/foundry-library` to avoid similar headaches.

sam-goldman:
> Any other fixes, changes, improvements, etc we want to make prior to merging the `pate/foundry-library` branch should also be developed by branching off of `pate/foundry-library` to avoid similar headaches.

Got it. Let's just hold off till later then to merge this then. I would've pushed it weeks ago but it was hiding in a random branch.

sam-goldman:
Closed by #585

PR #420, opened 2023-01-26
feat(pg): Merge chugsplash-foundry into monorepo

RPate97:
## Purpose
- Moves the ChugSplash Foundry library into the monorepo
- Updates the main readme to reflect the updates made in the foundry repo
- Updates the documentation to include Foundry documentation

Note that when this PR is merged, we should make sure to do a release shortly afterward to ensure the documented setup process works as expected.

## Regression
There is one regression as a result of this change. It is no longer possible to use a typescript file for your ChugSplash config file when performing deployments using the Foundry repo. Given that we encourage people to use JSON when deploying via foundry though, I think this is acceptable. 

## Testing
I, of course, recommend testing this by running through the updated getting started guide to verify that you are able to successfully install the library and perform a deployment/run scripts. Note that you'll need to link the ChugSplash packages locally to make it work.

I also encountered quite a few development environment-related issues when setting this up. I believe I resolved these issues, but I recommend doing quite a bit of testing within the plugins package as well to verify that you are able to do test deployments with both the hardhat plugin and the foundry library. 

As part of this change, I added quite a few more ChugSplash config files to the plugins package. These config files are used to run integration tests on the Foundry library. However, they also make running the hardhat tests quite slow because we automatically deploy all config files detected. I've added an additional param to the hardhat test command to allow the user to specify a single config to deploy if desired:
```
npx hardhat test --config-path ./chugsplash/Storage.config.ts
```

### Running foundry tests: 
1. Add the following variables to your .env file: 
```
DEV_FILE_PATH=./dist/foundry/index.js
NETWORK=localhost
```
2. Start anvil in a separate process: `anvil`
3. Run the foundry tests against it: `forge test`


sam-goldman:
> ## Regression
> It is no longer possible to use a typescript file for your ChugSplash config file when performing deployments using the Foundry repo.

Just wondering, why is this the case?

RPate97:
@sam-goldman 
> Just wondering, why is this the case?

Because we compile the typescript code to js for distribution and the user runs it with node. We'd need to have the user also install ts-node in their project (there may be other ways to add support for it though) and have them run it with that for typescript config files to work. Currently, the standalone foundry library does exactly that, but the ts-node dependency is installed in the library so it's less overt.

sam-goldman:
> * Updates the main readme to reflect the updates made in the foundry repo
> * Updates the documentation to include Foundry documentation

It doesn't look like any documentation/readme was changed in this pr?

RPate97:
> > * Updates the main readme to reflect the updates made in the foundry repo
> > * Updates the documentation to include Foundry documentation
> 
> It doesn't look like any documentation/readme was changed in this pr?

Oof, seems I must have lost those changes when rebasing.

README.md
```diff
@@ -1,41 +1,38 @@
-# ChugSplash
+# ChugSplash Foundry
```
sam-goldman, line 1: Shouldn't this just be `ChugSplash`?

README.md
```diff
@@ -1,41 +1,38 @@
...
 
-A Hardhat and Foundry plugin for deploying and managing upgradeable smart contracts.
-
-If you're using Foundry, check out out our Foundry plugin [here](https://github.com/chugsplash/chugsplash-foundry). Otherwise, keep reading for the Hardhat plugin.
+A modern, declarative, and deterministic smart contract deployment and upgrade framework for Foundry and Hardhat.
```
sam-goldman, line 3: Thoughts on leaving 'modern' out? It isn't very descriptive and sounds a little buzzwordy IMO

Also, I think a slightly better way to rephrase this would be "A declarative and deterministic framework for deploying and upgrading smart contracts. Available as a Foundry and Hardhat plugin."

sam-goldman:
When I `yarn install` from the root of the monorepo, the `yarn.lock` gets updated. Not sure if this is just on my end

sam-goldman:
Could you post instructions for running the forge tests in the monorepo?

README.md
RPate97: Yeah, I'm fine with that description. I don't think modern is a superfluous word though. I think it implies this is a tool that is on par with the quality of modern web2 developer tools.

RPate97:
> Could you post instructions for running the forge tests in the monorepo?

I've added this to the PR description, did you want it in the README?

sam-goldman:
The old chugsplash-foundry repo is referenced twice in the file `import-openzeppelin.md` and once in `storage-checker.md`

docs/foundry/getting-started.md
```diff
@@ -0,0 +1,317 @@
...
+To create an Anvil node in a stand-alone process, run:
+
+```
+anvil
+```
+
+In another terminal window, run the following command to deploy your upgradeable contract:
+
+```
+forge script --rpc-url http://localhost:8545 script/MyFirstProject.s.sol
```
sam-goldman, line 188: We should either do `--rpc-url localhost` here or remove this from `foundry.toml`:
```
[rpc_endpoints]
localhost = "http://127.0.0.1:8545"
```

This is because the setting in foundry.toml is unused if we don't do `--rpc-url localhost`

docs/foundry/getting-started.md
```diff
@@ -0,0 +1,317 @@
...
+build_info = true
+extra_output = ['storageLayout']
+force = true
+fs_permissions = [{ access = "read", path = "./"}]
+
+remappings = [
+  'forge-std/=node_modules/forge-std/src/',
+  'ds-test/=node_modules/ds-test/src/',
+  'chugsplash/=node_modules/@chugsplash/plugins/dist/contracts/'
+]
```
sam-goldman, line 74: It's redundant to have both a `remappings` field in `foundry.toml` as well as a `remappings.txt` file, so we can just remove this parameter here

docs/foundry/getting-started.md
RPate97: > the setting in foundry.toml is unused if we don't do --rpc-url localhost

Our library resolves the network url from the `foundry.toml` file using the network name fetched from the environment variable `NETWORK` or the default name `localhost`, so we need this value regardless. I suppose we could eliminate it by defaulting to `http://127.0.0.1:8545` if nothing is defined, but we'd still need to have the entry for all other networks so IMO we should maintain the pattern for localhost as well. 

I initially updated this because I don't think forge resolves `--rpc-url localhost` to the url defined in `foundry.toml`. I could be wrong about that though as it seems you disagree.

However, I just realized the `--rpc-url` parameter is actually not used at all because our library connects to the target network using the process described above. The only reason `--rpc-url` is needed is if the user wants to connect to the network from the forge environment in which case we recommend using `chugsplash.refresh()` to fork the target network which ensures that the transactions performed by our library outside the forge environment are replicated within it. 

So I think the correct answer here is actually to remove the `--rpc-url` param entirely.

sam-goldman: If we remove the `--rpc-url` flag and the user adds any transactions to the script that don't involve the `chugsplash` contract, the transactions won't go through. This is because foundry will use the in-process network for those transactions. That's why I originally added the `--rpc-url` flag

How about we just leave as is?

sam-goldman (approved):

PR #424, opened 2023-01-28
fix(core): Throw error if immutable variable value is defined in both the contract and config file

RPate97:
## Purpose
Fixes an issue where defining the value of an immutable variable in both the contract and config file could lead to unanticipated results.


sam-goldman:
This looks great! Super clean

PR #425, opened 2023-01-30
fix(core): allow user to explicitly define proxy type in ChugSplash file

sam-goldman:


packages/core/src/config/parse.ts
```diff
@@ -146,62 +179,70 @@ export const assertStorageSlotCheck = async (
...
+
+    contracts[referenceName] = proxy
   }
 
-  return parsed
+  return JSON.parse(
+    Handlebars.compile(JSON.stringify(parsedConfig))({
+      ...contracts,
+    })
+  )
```
RPate97, line 245: Just curious, what is the purpose of using handlebars here?

PR #459, opened 2023-02-03
docs: how chugsplash works

sam-goldman:


docs/how-chugsplash-works.md
```diff
@@ -0,0 +1,252 @@
...
+
+ChugSplash's declarative format significantly reduces the amount of code necessary to deploy a set of contracts. ChugSplash transforms Optimism's L1 deployment scripts from [1500+ lines](https://github.com/ethereum-optimism/optimism/tree/sc/ctb-chugsplash/packages/contracts-bedrock/deploy) with hardhat-deploy, or [630 lines](https://gist.github.com/brockelmore/76f445bab01c2479f011e66332df7f82) using Forge scripts, to just over [100 lines](https://github.com/ethereum-optimism/optimism/blob/sc/ctb-chugsplash/packages/contracts-bedrock/chugsplash/mainnet.ts).
+
+ChugSplash lets you:
+* Define deployments and upgrades [declaratively](https://github.com/chugsplash/chugsplash/blob/develop/docs/chugsplash-file.md#layout-of-a-chugsplash-file) in a single file, inspired by [Terraform](https://www.terraform.io/).
+* Guarantee that your deployments and upgrades are fully **deterministic**, meaning they can't be halted for any reason.
+* Guarantee that your deployments and upgrades are **atomic**, meaning they're finalized in a single transaction.
+* Approve upgrades of *any* size with a single tiny transaction from a multisig or governance.
```
RPate97, line 28:
```suggestion
* Securely propose deployments and upgrades from a CI process.
* Approve upgrades of *any* size with a single tiny transaction from a multisig or governance.
```

docs/how-chugsplash-works.md
```diff
@@ -0,0 +1,252 @@
...
+
+The managed service is currently in a closed beta. If you'd like to join Optimism and other top protocols in our beta, you can apply [here](https://o7n5gakt66b.typeform.com/to/UoSoli9r). We're accepting only a handful of beta users to ensure that each of them has a fantastic experience.
+
+## How ChugSplash Works
+
+Note: This is about to get technical. You should be proficient with Solidity and have an understanding of proxies and Merkle trees.
+
+### The Brief Version
+
+ChugSplash deployments and upgrades begin when the user defines a [ChugSplash file](https://github.com/chugsplash/chugsplash/blob/develop/docs/chugsplash-file.md). The state variables in the ChugSplash file are converted into 32-byte storage slot key/value pairs using the contract's storage layout. These storage slot key/value pairs are set via `SSTORE` in your proxy. This approach allows us to guarantee that the deployment or upgrade is deterministic, meaning it can't halt for any reason. Since ChugSplash relies on `SSTORE`, all contracts deployed by ChugSplash are proxied. By default, these proxies are upgradeable, but they can always be made non-upgradeable by setting the `admin` to `address(0)`.
```
RPate97, line 53: IMO this is confusing because it focuses on only one specific part of the system and ignores everything else. I think we should either rewrite this to cover more of the system at a higher level or remove it entirely. FWIW, I don't think the intended audience is here for the short version.

docs/how-chugsplash-works.md
```diff
@@ -0,0 +1,252 @@
...
+function setStorage(bytes32 _key, bytes32 _value) external {
+    assembly {
+        sstore(_key, _value)
+    }
+}
+```
+2. Trigger the `setStorage` function which is now available on the proxy.
+3. Set the proxy's implementation back to `address(0)`.
+
+Since these steps all occur over the course of a *single* transaction, the proxy's implementation is never set to anything other than `address(0)` *between* transactions. This ensures a malicious attacker can't trigger a `setStorage` call on the proxy.
```
RPate97, line 173: This will need to be updated.

docs/how-chugsplash-works.md
```diff
@@ -0,0 +1,252 @@
...
+
+This allows us to guarantee that your deployment is deterministic. We wouldn't be able to make this guarantee if we called functions because they can contain arbitrary logic. This means we wouldn't be able to reliably determine the end-state of the upgrade or even which state variables are changing.
+
+Our approach also removes the need for initializer functions, which removes the threat of uninitialized proxies getting hijacked by attackers. Uninitialized proxies have been the cause of some notable vulnerabilities, including the Nomad bridge hack, the $10 million bug bounty paid by Wormhole, and the $2 million bug bounty paid by Arbitrum.
+
+### If a multisig or governance is approving the deployment from a UI, and anybody could be executing it, how can you ensure that the contracts are verified on Etherscan?
+
+There are two solutions to this problem.
+1. You can execute your own deployment, which is as easy as spinning up a Docker container.
+2. You can sign up for the ChugSplash Managed Service (described [here](#wrapping-up)), which will verify the contracts on your behalf for free.
```
RPate97, line 193: You said previously that the managed service is invite-only. We should probably reconcile that with what you're saying here. 

I recommend just removing this faq entry for now. We can come back and update it to add more info about Etherscan verification once we've nailed down how the managed service is going to work.

docs/how-chugsplash-works.md
```diff
@@ -0,0 +1,252 @@
...
+```
+[0x000...000, 0x000...04D2]
+```
+
+This encoding occurs off-chain using the contract's storage layout.
+
+> At this point, you might be thinking, "*why would they compile the variables into individual storage slot values instead of just calling existing functions on the contract?*"\
+The short answer is that it allows us to guarantee your deployment is deterministic, meaning it can't halt for any reason. Head to the [FAQ section](#why-does-chugsplash-encode-variables-into-storage-slot-values-instead-of-just-calling-existing-functions-on-the-contract) for further explanation.
+
+It's feasible to approve a small deployment in a single transaction from a multisig or governance, but this quickly becomes unrealistic for larger deployments. One solution would be to use multicall from a multisig, which batches smaller transactions into a single larger one. However, multicall transactions are capped at the block gas limit, so this doesn't solve the original problem of *click* *click* *clicking* on a Ledger for 30 minutes during a large deployment. Additionally, it becomes difficult to verify the correctness of these enormous transactions on a tiny Ledger screen or a multisig UI.
```
RPate97, line 104: It's pretty confusing to jump to talking about approving with a multisig in the section on proposals. I get why you're talking about it though b/c it's important to explain the reasoning behind the architecture in order to explain how the proposal step works. However, I wouldn't include that explanation in this section. I would move it up to the beginning where we talk about the three steps. I think that section could be a lot longer and include all the details about why those three steps exist, what the properties of them are, and how they relate to each other.

RPate97 (approved):
Looks pretty good, I left some comments with my thoughts. 

I think the biggest area for improvement is the explanation of the three steps. They're kind of confusing because you try to explain why the architecture is the way that it is and what each of the steps is actually doing at the same time. I think it would be better to first focus on explaining why the architecture is set up the way that it is and what the goals are. Then we can explain how we practically deliver on those goals.

docs/how-chugsplash-works.md
```diff
@@ -0,0 +1,171 @@
...
+[0x000...000, 0x000...04D2]
+```
+
+This encoding occurs off-chain using the contract's storage layout.
+
+These key/value pairs, called `SetStorage` actions, are encoded as leafs in a Merkle tree. Additionally, each contract's creation bytecode appended with its constructor arguments is encoded as a leaf in the Merkle tree. These leafs are called `DeployImplementation` actions. The remote executor must supply the Merkle proof of each `SetStorage` and `DeployImplementation` action during the deployment, or else the transaction will revert.
+
+If you're wondering why we use a Merkle tree instead of a simple hash of the deployment, the answer is that using a simple hash would make it impossible to support arbitrarily large deployments. This is because the executor would need to supply the entire deployment's data in a single transaction, which is capped at the block gas limit.
+
+In order for the executor to complete the deployment remotely, it must be able to fetch the deployment info and re-create the Merkle tree. Additionally, it must have the information necessary to verify the contracts on a block explorer.
```
RPate97, line 111:
```suggestion
In order for the executor to complete the deployment remotely, it must be able to fetch the deployment info and re-create the Merkle tree.
```

docs/how-chugsplash-works.md
```diff
@@ -0,0 +1,171 @@
...
+
+In addition to committing this info to IPFS, the user also submits a single `propose` transaction on the project's `ChugSplashManager` contract. There are three inputs to this transaction:
+
+1. The IPFS URI. This allows the executor to fetch the data from IPFS during the execution phase.
+2. The root of the Merkle tree. Each Merkle proof supplied by the executor must yield this Merkle root.
+3. The number of leafs in the Merkle tree (i.e. the number of `SetStorage` and `DeployImplementation` actions). It's necessary to explicitly specify this to ensure that each action is only executed exactly once.
+
+These three inputs are hashed to yield a 32-byte **bundle ID**, which is the unique identifier for the entire deployment. This bundle will be approved by the project's owner during the next step. It's called a bundle ID because the set of `DeployImplementation` and `SetStorage` actions are referred to as a bundle internally.
+
+Anyone can view a git-style diff of a proposed upgrade, including state variables, before any action is taken on it. We recommend that teams integrate proposals into their CI process.
```
RPate97, line 125: Nit: This isn't supported yet, so we might want to remove it.

PR #479, opened 2023-02-08
feat(ct): add support for uups proxies

RPate97:
## Purpose
- Adds support for UUPS Proxies
- Updates the IProxyAdapter interface to include four functions `initiateExecution`, `setStorage`, `completeExecution`, and `changeProxyAdmin`. Also updates the `DefaultAdapter` to conform to the new interface.
- Adds new adapters `OZUUPSAdapter` and `OZTransparentAdapter`
- Removes the Reverter/Updater contracts and replaces them with a single generic IProxyUpdater interface which must be implemented for each proxy type.
- Adds updater contracts `DefaultUpdater`, `OZUUPSUpdater`, and `OZTransparentUpdater`
- Removes redundant contract implementation updates due to `updateToAndCall` except for open zeppelin transparent proxies where the pattern is necessary to circumvent their restrictions on what functions admins are allowed to call.
- Adds E2E tests for OpenZeppelin Transparent and UUPS proxies deployed using the open zeppelin upgrades plugin. 

## Note
This PR does not include docs or unit tests. Both of those will come in follow-up PRs.


sam-goldman:
nit: should be feat(ct), not feat(contracts)

packages/contracts/contracts/ChugSplashManager.sol
```diff
@@ -674,7 +668,14 @@ contract ChugSplashManager is OwnableUpgradeable, ReentrancyGuardUpgradeable {
...
             // Upgrade the proxy's implementation contract.
-            _upgradeProxyTo(proxy, adapter, implementation);
+            (bool success, bytes memory result) = adapter.delegatecall(
+                abi.encodeCall(IProxyAdapter.completeExecution, (proxy, implementation))
+            );
+            if (success == false) {
+                assembly {
+                    revert(add(result, 32), mload(result))
+                }
+            }
```
sam-goldman, line 678: Did you intend to leave this in? Only necessary for testing purposes right?

packages/contracts/contracts/ChugSplashRegistry.sol
```diff
@@ -248,18 +237,24 @@ contract ChugSplashRegistry is Initializable, OwnableUpgradeable, IChugSplashReg
...
-    function addProxyType(bytes32 _proxyType, address _adapter) external {
+    function addProxyType(bytes32 _proxyType, address _adapter, address _updater) external {
         require(
             adapters[_proxyType] == address(0),
             "ChugSplashRegistry: proxy type has an existing adapter"
         );
+        require(
+            updaters[_proxyType] == address(0),
+            "ChugSplashRegistry: proxy type has an existing updater"
+        );
```
sam-goldman, line 255: FYI I'm gonna be removing this logic soon so no need to do unit tests for it

packages/contracts/contracts/updaters/DefaultUpdater.sol
```diff
@@ -0,0 +1,64 @@
...
+     */
+    function setStorage(bytes32 _key, bytes32 _value) external ifAdmin {
+        assembly {
+            sstore(_key, _value)
+        }
+    }
+
+    fallback() external {
+        revert();
+    }
```
sam-goldman, line 63: I think we should follow the same pattern in `Proxy.sol` here:
```
    receive() external payable {
        revert();
    }

    fallback() external payable {
        revert();
    }
```

Note that both functions are `payable`

packages/contracts/contracts/updaters/DefaultUpdater.sol
```diff
@@ -0,0 +1,64 @@
...
+     * @param _value New value for the storage key.
+     */
+    function setStorage(bytes32 _key, bytes32 _value) external ifAdmin {
+        assembly {
+            sstore(_key, _value)
+        }
+    }
+
+    fallback() external {
+        revert();
```
sam-goldman, line 62: It'd be nice if there was an error message here and all the other `revert` statements. That way, front-ends and our tests can have some info to display. Particularly, I can see it being difficult to debug failing tests that don't return any error messages.

Something like: `DefaultUpdater: caller is not admin`

packages/core/src/utils.ts
```diff
@@ -648,6 +749,9 @@ export const assertValidUpgrade = async (
 
 To perform an upgrade, you must first transfer ownership of the contract(s) to ChugSplash using the following command:
 npx hardhat chugsplash-transfer-ownership --network <network> --config-path <path> --proxy <proxyAddress>
+
+If you are using an external UUPS proxy, you'll need to give your ChugSplashManager address ${chugSplashManagerAddress}
+permission to upgrade them.
```
sam-goldman, line 754: Could we be more specific and say:
```
If you are using any Transparent proxies, you must transfer ownership of each to ChugSplash using the following command:
npx hardhat chugsplash-transfer-ownership --network <network> --config-path <path> --proxy <proxyAddress>

If you are using any UUPS proxies, you must give your ChugSplashManager contract {chugSplashManagerAddress}
permission to call the 'upgradeTo' function on each of them.
```

packages/plugins/src/hardhat/deployments.ts
```diff
@@ -56,7 +55,7 @@ export const deployAllChugSplashConfigs = async (
 
   let executor: ChugSplashExecutorType | undefined
   if (!remoteExecution) {
-    executor = await initializeExecutor(hre.ethers.provider)
+    executor = hre.chugsplash.executor
```
sam-goldman, line 58: Nice, I like this

packages/contracts/contracts/updaters/OZUUPSUpdater.sol
```diff
@@ -0,0 +1,133 @@
...
+contract OZUUPSUpdater is IProxyUpdater {
+    /**
+     * @notice The storage slot that holds the address of the implementation.
+     *         bytes32(uint256(keccak256('eip1967.proxy.implementation')) - 1)
+     */
+    bytes32 internal constant IMPLEMENTATION_KEY =
+        0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc;
+
+    /**
+     * @notice The storage slot that holds the address of the implementation.
```
sam-goldman, line 22: address of the owner*

packages/contracts/contracts/updaters/OZUUPSUpdater.sol
```diff
@@ -0,0 +1,133 @@
...
+    function setStorage(bytes32 _key, bytes32 _value) external ifChugSplashAdmin {
+        assembly {
+            sstore(_key, _value)
+        }
+    }
+
+    /**
+     * @notice Sets up the proxy updater. In this case, there is no setup required.
+     */
+    function setup() public {
```
sam-goldman, line 116: Thoughts on just having a `changeAdmin(address)` instead of a `setup` function? Then we can simplify this logic a bit by just copying the existing `changeAdmin` function in `Proxy.sol`

sam-goldman: Actually, I suppose we want to make a standard interface that inherits from `IProxyUpdater`. Thoughts on changing `setup` to `initiateExecution` and `teardown` to `completeExecution` to match the adapter?

packages/contracts/contracts/updaters/OZUUPSUpdater.sol
```diff
@@ -0,0 +1,133 @@
...
+        if (_getChugSplashAdmin() == address(0)) {
```
sam-goldman, line 117: Thoughts on changing `setup` to `initiateExecution` and `teardown` to `completeExecution` to match the adapter?

sam-goldman, line 117: nit: I think we can remove `if (_getChugSplashAdmin() == address(0))`. Will remove the need to do one unit test :)

packages/core/src/utils.ts
```diff
@@ -664,6 +768,19 @@ npx hardhat chugsplash-transfer-ownership --network <network> --config-path <pat
...
     }
 
+    // Check new implementations include the UUPS interface if the proxy type is marked as UUPS
+    for (const [referenceName] of Object.entries(parsedConfig.contracts)) {
+      const contractConfig = parsedConfig.contracts[referenceName]
+      if (contractConfig.proxyType === 'oz-uups') {
+        if (!bytecodeContainsUUPSInterface(contractConfig.contract)) {
+          // throw error
+          throw new Error(
+            `Contract ${referenceName} proxy type is marked as UUPS, but the new implementation no longer implements the UUPS interface. You must include the expected interface or you will no longer be able to upgrade this contract.`
```
sam-goldman, line 778: How about we say something more expliclt here like:
```
...no longer has a public `upgradeTo(address)` function.
You must include this function or...
```

sam-goldman:
One thing we need to resolve. This isn't necessary to do in this PR, but we'll need to address it before we release the UUPS feature.

Currently, users have no way to reclaim ownership of their UUPS proxy via `transferProxyOwnership`. This isn't an issue if the new implementation uses `AccessControl` and the user is the "root admin" (i.e. their role is `DEFAULT_ADMIN_ROLE`), because the user can simply revoke the ChugSplashManager's role. This becomes an issue in every other case though:
1. If the implementation uses `Ownable`, the ChugSplashManager will be the only address that has power to transfer ownership of the proxy. So, it's necessary for us to make a mechanism to transfer ownership via its `transferOwnership` function.
2. Likewise if the implementation uses `AccessControl` and the ChugSplashManager has the root admin privilege. This would be the case if the root admin is the only account that can call the `upgradeTo` function, which means the user would need to transfer the root admin privilege to the ChugSplashManager initially to perform the upgrade
3. Any arbitrary user-defined "transfer ownership" function that doesn't follow the `Ownable` or `AccessControl` interface. E.g. `updateProxyAdmin`

We can solve 1 and 2 by creating two distinct adapters for these specific cases. These two adapters would replace the existing `OZUUPSAdapter`. The existing UUPS adapter's `changeProxyAdmin` doesn't do anything in this context anyways unless the new implementation exposes the `changeAdmin` function, which isn't very likely since this function is usually only used for Transparent proxies.

Since these two new adapters only need to change the logic inside their `changeProxyAdmin` function, I think the cleanest structure would be the following. The `IOZUUPSAdapter` contract, which is marked `abstract`, inherits from `IProxyAdapter`. `IOZUUPSAdapter` keeps its existing logic for every function other than `changeProxyAdmin`. We make `changeProxyAdmin` an unimplemented function by doing:
```
function changeProxyAdmin(address payable _proxy, address _newAdmin) external virtual;
```

The two inheriting adapters will be `OZOwnableUUPSAdapter` and `OZAccessControlUUPSAdapter`. Their implementations for `changeProxyAdmin` function will look something like:
```
// OZOwnableUUPSAdapter
function changeProxyAdmin(address payable _proxy, address _newAdmin) external {
    Ownable(_proxy).transferOwnership(_newAdmin);
}
```

```
// OZAccessControlUUPSAdapter
function changeProxyAdmin(address payable _proxy, address _newAdmin) external {
    // `DEFAULT_ADMIN_ROLE` always equals bytes32(0)
    AccessControl(_proxy).grantRole(bytes32(0), _newAdmin);
}
```

Supporting these two adapters will cover the vast majority of cases.

Lastly, to cover case 3 above, it'd suffice to just say something like this in our docs:
```
If you'd like to use an access control mechanism that isn't OpenZeppelin's `AccessControlUpgradeable` or `OwnableUpgradeable`, please reach out to us in our Discord, and we'll support it for you.
```

packages/core/src/utils.ts
```diff
@@ -554,14 +558,95 @@ export const isTransparentProxy = async (
...
+  provider: providers.Provider,
+  proxyAddress: string
+): Promise<boolean> => {
+  const implementationAddress = await getEIP1967ProxyImplementationAddress(
+    provider,
+    proxyAddress
+  )
+
+  // We don't consider default proxies to be ERC proxies, even though they share the same
+  // interface.
```
sam-goldman, line 579: This comment isn't true for UUPS proxies

packages/core/src/utils.ts
```diff
@@ -554,14 +558,95 @@ export const isTransparentProxy = async (
...
+
+  // If proxy owner is not a valid address, then proxy type is incompatible
+  if (!ethers.utils.isAddress(ownerAddress)) {
+    return false
+  }
+
+  return true
+}
+
+const bytecodeContainsUUPSInterface = async (
```
sam-goldman, line 604: This function doesn't need to be async, and all calls to this function can remove the `await` keyword

packages/core/src/utils.ts
```diff
@@ -554,14 +558,95 @@ export const isTransparentProxy = async (
...
+  const implementationAddress = await getEIP1967ProxyImplementationAddress(
+    provider,
+    proxyAddress
+  )
+
+  // We don't consider default proxies to be ERC proxies, even though they share the same
+  // interface.
+  if ((await isDefaultProxy(provider, implementationAddress)) === true) {
+    return false
+  }
```
sam-goldman, line 582: We should remove this entire if-statement since it's relevant only for Transparent proxies, not for UUPS proxies

packages/contracts/contracts/updaters/OZUUPSUpdater.sol
sam-goldman: This contract actually can't be publicly accessible across multiple blocks because the `initiate` call in `OZUUPSAdapter.initiateExecution` always occurs immediately after the `upgradeTo` call. Even if it did occur across multiple blocks though, the if-statement doesn't add any extra security, because `_getChugSplashAdmin()` always returns `address(0)`

PR #480, opened 2023-02-08
refactor some uups logic

sam-goldman:


packages/core/src/utils.ts
```diff
@@ -709,34 +713,51 @@ export const assertValidUpgrade = async (
...
+          ProxyABI,
+          chugsplashManager
+        )
+        try {
+          // Attempt to staticcall the `upgradeTo` function on the proxy from the ChugSplashManager's
+          // address. Note that we upgrade to the address of the ChugSplashRegistry's proxy to ensure
+          // that the call doesn't fail due to this check:
+          // https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/proxy/ERC1967/ERC1967Upgrade.sol#L44
+          await UUPSProxy.callStatic.upgradeTo(
+            CHUGSPLASH_REGISTRY_PROXY_ADDRESS
```
RPate97, line 737: This doesn't work b/c the OpenZeppelin does not consider the interface of the ChugSplashRegistry to be UUPS compatible. So if you run the e2e tests, the UUPS test will fail: 
`yarn test:hardhat`

If you print the error you'll see it's message is: 
`ERC1967Upgrade: new implementation is not UUPS`

This error occurs b/c the target does not include the `proxiableUUID()` function: 
https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/dd8ca8adc47624c5c5e2f4d412f5f421951dcc25/contracts/proxy/ERC1967/ERC1967UpgradeUpgradeable.sol#L91

To be fair, this issue also existed in my implementation, your refactor just uncovered it. The solution is to use an address here that conforms to the full UUPS expected spec. So we'd either need to add the `proxiableUUID()` function to the ChugSplashRegistry or use the address of a different contract that implements it.

PR #514, opened 2023-02-17
feat(exec): Update deployment status from executor

RPate97:
## Purpose
- Adds logic integrating the executor with ChugSplash managed
- Fixes a couple of bugs with the propose tasks in Foundry and Hardhat


packages/plugins/src/hardhat/tasks.ts
```diff
@@ -350,6 +350,7 @@ task(TASK_CHUGSPLASH_PROPOSE)
     'skipStorageCheck',
     "Upgrade your contract(s) without checking for storage layout compatibility. Only use this when confident that the upgrade won't lead to storage layout issues."
   )
+  .addFlag('remoteExecution', 'Upload config file to IPFS for remote execution')
```
sam-goldman, line 353: We actually shouldn't have `remoteExecution` as a parameter to the `chugsplashProposeTask`, since the user doesn't need to think about this. Instead of having `remoteExecution` as a parameter, we can do this:
```
const remoteExecution = (await getChainId(provider)) !== hre.config.networks.hardhat.chainId
```
We do this everywhere else in our codebase to get the value of `remoteExecution`

RPate97: So the problem with this is that it entirely prevents me from testing this feature on a local network. I see why you don't want it visible to the user, but I do think we should try to make it possible to test things locally.

packages/plugins/src/hardhat/tasks.ts
```diff
@@ -317,6 +309,11 @@ export const chugsplashProposeTask = async (
     'hardhat'
   )
 
+  const remoteExecution =
+    process.env.FORCE_REMOTE_EXECUTION === 'true'
```
sam-goldman, line 313: What's the context for this part?

RPate97: This is necessary to make it possible to test performing deployments against a local node with an executor running in a separate process as if the deployment was being done against a live network. This is required to test the deployment process using the UI. 

I'm using an environment variable here because I agree that this functionality should not be visible to the user. 

We should actually expand this pattern to cover all other cases where we use `const remoteExecution = (await getChainId(provider)) !== hre.config.networks.hardhat.chainId` to determine whether the execution is being performed by a remote executor. Currently, it's quite difficult to run realistic tests when I make changes to the executor b/c of this limitation.

sam-goldman: Ok got it, we should probably refactor this into a function `isRemoteExecution`. Want to make a ticket for that? I'll go ahead and approve this PR.

sam-goldman (approved):

PR #524, opened 2023-02-21
feat(exec): add exponential backoff retry policy

RPate97:
## Purpose
**Updates the executor to improve reliability**
- Fixes an issue where all events are retried continuously due to an issue where events were not popped from the queue if there was no active bundle id found. 
- Improves the retry policy by implementing exponential backoff this prevents some events from being infinitely retried which hurts the performance of the executor
- Adds a promise timeout utility and uses the utility in IPFS calls to fetch ChugSplash config files. This handles a bug where the executor can hang if it cannot retrieve a config file from IPFS.


packages/executor/src/index.ts
```diff
@@ -453,10 +537,10 @@ export class ChugSplashExecutor extends BaseServiceV2<
 
         // If we make it to this point, we know that the executor has executed the bundle (or that it
         // has been cancelled by the owner), and that the executor has claimed its payment.
-
-        // Remove the current event from the events queue.
-        this.state.eventsQueue.pop()
       }
+
+      // Remove the current event from the events queue.
+      this.state.eventsQueue.pop()
```
RPate97, line 543: This is what was causing the infinite retry loop I noticed. If no active bundle is found in the project then the event would not be removed from the queue, and we'd get an infinite retry loop.

sam-goldman:
Looks good!

Since we know the exact time that a bundle can be re-claimed (`executionLockTime + timeClaimed`), we might want to do retries at this exact time instead of doing exponential backoff in the future. We can make X attempts before discarding the bundle. We can discuss this in the future though; this PR looks good

sam-goldman (approved):

PR #551, opened 2023-03-07
feat(core): Add support for async config files

RPate97:
## Purpose
Adds support for asynchronously resolved ChugSplash configs. Users will now be able to export an asynchronous function from their config file and our tooling will take care of resolving it internally. 

## Context
I was integrating [Gelato Relay](https://github.com/gelatodigital/relay-contracts) (which appears to be their primary focus right now, though it is not their core protocol) on the plane flight back and encountered a use case for this feature so I went ahead and implemented it. We may want to consider updating the docs to make this the recommended method of creating your config file b/c it is significantly more powerful. 

## Example
The Gelato Relay codebase includes some mock contracts used for testing. This is the config file I wrote to deploy them using the asynchronous config file feature. 

On a side note, this is also just an interesting use case for ChugSplash. These contracts will never be deployed to mainnet, but it still makes sense to use ChugSplash to deploy them IMO. BTW doing this deployment actually also required that I disable our check preventing all non immutable constructor args. [I've created a ticket related to that.](https://linear.app/chugsplash/issue/CHU-128/allow-forcing-non-immutable-constructor-args)
```js
import { UserChugSplashConfig } from "@chugsplash/core";
import { getNamedAccounts } from "hardhat";
import { INIT_TOKEN_BALANCE } from "../test/constants";

export default async (): Promise<UserChugSplashConfig> => {
  const { deployer } = await getNamedAccounts();
  return {
    options: {
      projectName: "Gelato Relay Mocks",
    },
    contracts: {
      MockERC20: {
        contract: "MockERC20",
        constructorArgs: {
          name: "MockERC20",
          symbol: "ME2",
        },
        variables: {
          // eslint-disable-next-line @typescript-eslint/naming-convention
          _name: "MockERC20",
          // eslint-disable-next-line @typescript-eslint/naming-convention
          _symbol: "ME2",
          // eslint-disable-next-line @typescript-eslint/naming-convention
          _totalSupply: INIT_TOKEN_BALANCE.toString(),
          // eslint-disable-next-line @typescript-eslint/naming-convention
          _balances: {
            [deployer]: INIT_TOKEN_BALANCE.toString(),
          },
          // eslint-disable-next-line @typescript-eslint/naming-convention
          _allowances: {},
        },
      },
      MockGelatoRelayContext: {
        contract: "MockGelatoRelayContext",
        constructorArgs: {},
        variables: {},
      },
      MockGelatoRelayContextERC2771: {
        contract: "MockGelatoRelayContextERC2771",
        constructorArgs: {},
        variables: {},
      },
      MockGelatoRelayFeeCollector: {
        contract: "MockGelatoRelayFeeCollector",
        constructorArgs: {},
        variables: {},
      },
      MockGelatoRelayFeeCollectorERC2771: {
        contract: "MockGelatoRelayFeeCollectorERC2771",
        constructorArgs: {},
        variables: {},
      }
    },
  };
};
```

### DIFF
--- a/.changeset/brave-clocks-act.md
+++ b/.changeset/brave-clocks-act.md
@@ -1,0 +1,6 @@
+---
+'@chugsplash/plugins': patch
+'@chugsplash/core': patch
+---
+
+Add support for async config files
--- a/packages/core/src/config/parse.ts
+++ b/packages/core/src/config/parse.ts
@@ -31,18 +31,29 @@ export const readParsedChugSplashConfig = async (
   artifactPaths: ArtifactPaths,
   integration: Integration
 ): Promise<ParsedChugSplashConfig> => {
-  const userConfig = readUserChugSplashConfig(configPath)
+  const userConfig = await readUserChugSplashConfig(configPath)
   return parseChugSplashConfig(provider, userConfig, artifactPaths, integration)
 }
 
-export const readUserChugSplashConfig = (
+export const readUserChugSplashConfig = async (
   configPath: string
-): UserChugSplashConfig => {
+): Promise<UserChugSplashConfig> => {
   delete require.cache[require.resolve(path.resolve(configPath))]
 
+  let config
   // eslint-disable-next-line @typescript-eslint/no-var-requires
-  let config = require(path.resolve(configPath))
-  config = config.default || config
+  let exported = require(path.resolve(configPath))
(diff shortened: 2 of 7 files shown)


sam-goldman:
This is great! Can you make a linear ticket for docs on this feature?

PR #553, opened 2023-03-09
feat(core): Add support for user defined types

RPate97:
## Purpose
Adds support for user-defined types

## Implementation Details
- Since the storage layout doesn't include information on the underlying type of user-defined types by default, I had to fetch the underlying type from the compiler output and inject it into the storage layout. 
- Note that while this feature adds support for user-defined types, we are limited to treating them like their underlying types.

### DIFF
--- a/docs/variables.md
+++ b/docs/variables.md
@@ -18,6 +18,7 @@ This is a reference that explains how to assign values to every variable type in
 - [Arrays](#arrays)
 - [Structs](#structs)
 - [Mappings](#mappings)
+- [User-Defined Value Types](#user-defined-types)
 
 
 ## Booleans
@@ -312,4 +313,18 @@ myMultiNestedMapping: {
   },
   ...
 },
+```
+
+## [User-Defined Value Types](https://docs.soliditylang.org/en/latest/types.html#user-defined-value-types)
+ChugSplash treats your user defined types as if they were their underlying types.
+
+Define your type in Solidity
+```solidity
+type UserDefinedType is uint256;
+UserDefinedType public userDefined;
+```
+
+In your ChugSplash config file:
+```ts
+userDefined: 1
 ```
--- a/.changeset/fuzzy-emus-stare.md
+++ b/.changeset/fuzzy-emus-stare.md
@@ -1,0 +1,6 @@
+---
+'@chugsplash/plugins': patch
+'@chugsplash/core': patch
+---
+
+Add support for user defined types
--- a/packages/core/src/actions/bundle.ts
+++ b/packages/core/src/actions/bundle.ts
@@ -1,6 +1,7 @@
 import { fromHexString, toHexString } from '@eth-optimism/core-utils'
 import { ethers, providers } from 'ethers'
 import MerkleTree from 'merkletreejs'
+import { astDereferencer } from 'solidity-ast/utils'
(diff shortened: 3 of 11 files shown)


sam-goldman:
tl;dr I think there are a couple issues with this PR so I opened a new PR to be merged into this one.

The issues:

1. I don't think it's a good idea to modify the `StorageLayout` in-place, which is what we currently do in `addUserDefinedTypesToStorageLayout`. The storage layout object is attached to the build info, which could be used later in a totally different part of the codebase (e.g. when displaying the diff).

2. [Here](https://github.com/chugsplash/chugsplash/blob/bca4ecdf3a49e5a416d2966e1a5eeab6ec9a36ac/packages/core/src/actions/bundle.ts#L359), you only iterate over the nodes for the current contract, and not any of its parents. To illustrate why this is an issue, the following example will result in an error:

a. Go to `Storage.sol` in the plugins package

b. Move the type definitions from the `Storage` contract to a new parent contract, `ParentStorage`:
```
// New contract
contract ParentStorage {
    // User-defined types are in this parent contract
    type UserDefinedType is uint256;
    type UserDefinedBytes32 is bytes32;
    type UserDefinedInt is int;
    type UserDefinedInt8 is int8;
    type UserDefinedUint8 is uint8;
    type UserDefinedBool is bool;
}

// Storage now inherits from ParentStorage
contract Storage is ParentStorage {
    // Everything stays the same below
    enum TestEnum { A, B, C }
    ...
```
c. Run `yarn test:hardhat`

To fix this bug, we actually need to iterate over every source in the compiler output, since the user-defined type node can be in any one of them. Unfortunately, this gets error prone because it's tricky to guarantee you've found the right node, and it's slow to iterate over every node in the compiler output many times. Fortunately, there's a handy library made by OpenZeppelin called [solidity-ast](https://github.com/OpenZeppelin/solidity-ast) that's built exactly for this.

To solve these issues, I opened a new PR to be merged into this one. It uses the `solidity-ast` library and gets the underlying storage type on the fly instead of modifying the existing storage layout.

docs/variables.md
```diff
@@ -18,6 +18,7 @@ This is a reference that explains how to assign values to every variable type in
 - [Arrays](#arrays)
 - [Structs](#structs)
 - [Mappings](#mappings)
+- [User Defined Types](#user-defined-types)
```
sam-goldman, line 21: nit: User Defined Types -> User-Defined Value Types

docs/variables.md
```diff
@@ -313,3 +314,17 @@ myMultiNestedMapping: {
   ...
 },
 ```
+
+## User defined types
```
sam-goldman, line 318: nit: User Defined Types -> [User-Defined Value Types](https://docs.soliditylang.org/en/latest/types.html#user-defined-value-types) (including the embedded link to the Solidity docs)

docs/variables.md
```diff
@@ -313,3 +314,17 @@ myMultiNestedMapping: {
...
+ChugSplash treats your user defined types as if they were their underlying types.
```
sam-goldman, line 319: > nit: User Defined Types -> User-Defined Value Types

sam-goldman:
Looks good! Just a couple nits

PR #557, opened 2023-03-11
docs: Add specification for ChugSplash config input types

RPate97:
## Purpose 
- Adds a spec document for a valid ChugSplash config and it's variable inputs
- Adds a test for a nested dynamic array

Does not cover constructor arguments or other config options. I [created a ticket to address those](https://linear.app/chugsplash/issue/CHU-135/add-constructor-args-and-other-config-options-to-chugsplash-config) in this document as well.

### DIFF
--- a/specs/chugsplash-config.md
+++ b/specs/chugsplash-config.md
@@ -1,0 +1,360 @@
+# ChugSplash Config Specification
+ChugSplash config files may be written in Javascript, Typescript, or JSON. For Javascript and Typescript users, their config file must have a default export that is either a valid ChugSplash config object or an asynchronous function which resolves to a valid ChugSplash config object. JSON users may only export a valid ChugSplash config object. The rest of this document outlines what a valid config file is.
+
+> Note that when using the ChugSplash Foundry library, users are limited to either Javascript or JSON.
+
+# Table of Contents
+- [Contract Config Definitions](#contract-config-definition)
+- [Contract Constructor Arguments](#contract-constructor-args)
+- [Signed & Unsigned Integers](#signed--unsigned-integers)
+- [Booleans](#booleans)
+- [Addresses](#addresses)
+- [Contracts](#contracts)
+- [Strings](#strings)
+- [Enums](#enums)
+- [User-defined Types](#user-defined-types)
+- [Fixed-size bytes](#fixed-size-bytes-arrays)
+- [Dynamically-sized bytes](#dynamically-sized-bytes)
+- [Fixed Size Arrays](#fixed-size-arrays)
+- [Dynamically-sized Arrays](#dynamically-sized-arrays)
+- [Mappings](#mappings)
+- [Structs](#structs)
+- [Arbitary Nested Types](#arbitary-nested-types)
+
+## Options
+- Project name
+- todo spec out other project options
+
+## Contracts Config Field
+JSON or Javascript object defining the contracts that should be deployed. Each key is a contracts reference name and its value is a contract definition.
+
+> Reference names are used internally by ChugSplash to identify specific instances of contracts. Each combination of a project name and contract reference name should be unique within the ChugSplash system.
+
+## Contract Config Definition
+JSON or Javascript object defining a contract to be deployed, it's constructor args, and variable definitions. Contract definitions must conform to the following format:
+
+```js
+ReferenceName: {
+  contract: 'LiteralContractName',
+  constructorArgs: {},
+  variables: {}
+}
+```
+
+## Contract Constructor Args
+A JSON or Javascript object defining a set of constructor args. All constructor arguements are required to be immutable variables...
+
+TODO finish defining constructor args spec
+
+## Contract Variables
+A JSON or Javascript object defining a set of ChugSplash input variables. ChugSplash is designed to remove the need for constructors or initializers by allowing the user to define the value of their variables directly during the deployment process. To ensure this process works reliably we perform input validation during the bundling process. Below we define the expected valid types. All other input types should be caught and rejected.
+
+Note that ChugSplash config files may be written in Javascript and therefore users are able to take advantage of various Javscript features and libraries to fetch values to be used in their ChugSplash config files.
+
+## Preserve
+Any variable may use the '{ preserve }' keyword which maintains the currently value of the variables storage slot. For more information on preserve, see the [Special Variable Definitions documentation](https://github.com/chugsplash/chugsplash/blob/develop/docs/special-var-defs.md#preserve-keyword).
+
+## Signed & Unsigned integers
+For both signed and unsigned integers, we allow three valid input types:
+- Base 10 string numbers: "1", "10", "1000000000000000000", "-100"
+- Base 10 literal numbers: 1, 10, 100, -100
+- [BigNumbers](https://docs.ethers.org/v5/api/utils/bignumber/)
+
+We enforce valid input ranges based on the specific byte size and sign of the target solidity variable. For example, if a variable is defined as uint8, we expect that values outside the range 0:255 to be caught and rejected during our bundling process. Likewise for int8, we expect that values outside the range -127:128 would be caught and rejected.
+
+Note that we explicitly do not support alternative base inputs such as hexadecimal for uint and int.
+
+### Examples
+```js
+// Solidity
+uint public x;
+
+// Javascript
+x: "1",
+```
+
+```js
+// Solidity
+uint public x;
+
+// Javascript
+x: 1,
+```
+
+```js
+// Solidity
+int public x;
+
+// Javascript
+x: -1,
+```
+
+```js
+// Solidity
+int public x;
+
+// Javascript
+x: "-1",
+```
+
+## Booleans
+For booleans, we only allow literal Javascript or JSON booleans.
+
+### Examples
+```js
+// Solidity
+boolean public x;
+boolean public y;
+
+// Javascript
+x: true,
+y: false,
+```
+
+## Addresses
+For addresses, we only accept 20 byte [DataHexStrings](https://docs.ethers.org/v5/api/utils/bytes/#DataHexString).
+
+### Example
+```js
+// Solidity
+address public x;
+
+// Javascript
+x: '0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266',
+```
+
+## Contracts
+For contracts, we accept two possible input types:
+- 20 byte [DataHexStrings](https://docs.ethers.org/v5/api/utils/bytes/#DataHexString)
+- ChugSplash contract references
+
+ChugSplash contract references are a format that allows users to reference other contracts deployed using ChugSplash. Our offchain tooling converts these references into the underlying contract address using Create2.
+
+Valid contract references must reference contracts in the same ChugSplash project and use the following format:
+```
+{{ <Contract Reference Name> }}
+```
+
+### Example
+```js
+// Solidity
+Contract public contractReference;
+
+// Javascript
+const config: UserChugSplashConfig = {
+  // Configuration options for the project:
+  options: {
+    projectName: 'My Project',
+  },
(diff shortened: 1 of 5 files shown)


specs/chugsplash-config.md
```diff
@@ -0,0 +1,276 @@
+# ChugSplash Config Specification
+ChugSplash config files may be written in Javascript, Typescript, or JSON. For Javascript/Typescript users, their config file must have a default export that is either a valid ChugSplash config object or an asynchronous function which resolves to a valid ChugSplash config object. The rest of this document outlines what a valid config file is.
```
sam-goldman, line 2: nit: JavaScript, TypeScript

specs/chugsplash-config.md
```diff
@@ -0,0 +1,276 @@
...
+
+> Note that when using the ChugSplash Foundry library, users are limited to either Javascript or JSON.
+
+## Options
+- Project name
+- todo spec out other project options
+
+## Contracts
+JSON or Javascript object defining the contracts that should be deployed. Each key is a contracts reference name and its value is a contract definition.
```
sam-goldman, line 11: nit: I think it's technically a JavaScript object?

specs/chugsplash-config.md
```diff
@@ -0,0 +1,276 @@
...
+
+### Examples
+```js
+// Solidity
+bytes32 public x;
+bytes32 public y;
+
+// Javascript
+x: '0x1111111111111111111111111111111111111111111111111111111111111111',
+y: ethers.utils.hexZeroPad(ethers.utils.hexlify(1), 32),
```
sam-goldman, line 89: Instead of writing this call to `hexZeroPad`, can you just write its output? I.e. `0x111...`. Makes it so that I don't need to go to my editor

specs/chugsplash-config.md
```diff
@@ -0,0 +1,276 @@
...
+
+```js
+// Solidity
+int public x;
+
+// Javascript
+x: "-1",
+```
+
+## Bytes32
```
sam-goldman, line 78: I think this should be "Fixed-size byte arrays" instead of "bytes32". In the Solidity docs: https://docs.soliditylang.org/en/latest/types.html

specs/chugsplash-config.md
```diff
@@ -0,0 +1,276 @@
...
+For bytes32, we allow only exactly 32 byte hex strings.
```
sam-goldman, line 79: We should be explicit about whether the string needs to be 0x-prefixed or not. Imo, we should require that it's 0x-prefixed

specs/chugsplash-config.md
sam-goldman, line 78: > For bytes32, we allow only exactly 32 byte hex strings.

This actually isn't true. In our `encodeVariable` function, we call [`ethers.utils.isBytesLike`](https://docs.ethers.org/v5/api/utils/bytes/#utils-isBytesLike), which allows for `Bytes` arrays, e.g. [0, 255]. Check out the 'bytes' section of `encodeVariable` to see what we currently allow

specs/chugsplash-config.md
```diff
@@ -0,0 +1,276 @@
...
+```js
+// Solidity
+string public x;
+
+// Javascript
+x: 'literalString',
+```
+
+## Booleans
+For booleans, we allow literal Javascript or JSON booleans or the literal strings "true" and "false".
```
sam-goldman, line 106: nit: just JavaScript, not JSON. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean

specs/chugsplash-config.md
```diff
@@ -0,0 +1,276 @@
...
+boolean public x;
+boolean public y;
+
+// Javascript
+x: "true",
+y: "false",
+```
+
+## User Defined Types
+We treat user defined types as if they were their underlying primitive type. We expect the input format to match the expected format of the underlying primitive.
```
sam-goldman, line 130: nit: they're called value types in Solidity, not primitives

specs/chugsplash-config.md
```diff
@@ -0,0 +1,276 @@
...
+// Solidity
+type UserDefinedType is uint256;
+UserDefinedType public x;
+
+// Javascript
+x: 1,
+```
+
+## Dynamic Bytes
+For dynamic bytes, we accept hex strings of an abritrary length as long as they have an even number of [nibbles](https://en.wikipedia.org/wiki/Nibble).
```
sam-goldman, line 143: > [Ethers.js] has a more explicit definition of what we allow, called [DataHexString](https://docs.ethers.org/v5/api/utils/bytes/#DataHexString)

Same here

specs/chugsplash-config.md
sam-goldman, line 106: Thoughts on just allowing `true` and `false` instead of their string equivalents? Less if-statements down the road during testing and writing the decoding logic

specs/chugsplash-config.md
RPate97: The goal of this example was to demonstrate that you could use libraries that generate valid 32 byte strings. TBH though this is probably unnecessary so I'm going to remove it.

specs/chugsplash-config.md
RPate97: Agreed. 

My goal with this spec was to define what we intend to support, not necessarily precisely what currently works. I plan to disallow things like this when implementing input validation in a follow-up PR.

specs/chugsplash-config.md
RPate97: Hmm, good question. In my experience, it's pretty standard for libraries to support the literal strings "true" and "false". However, in our case, I can see an argument that it could introduce longer-term bugs. I'm cool with removing this feature.

specs/chugsplash-config.md
RPate97: nice

specs/chugsplash-config.md
```diff
@@ -0,0 +1,359 @@
...
+
+## Contract Variables
+A JSON or Javascript object defining a set of ChugSplash input variables. ChugSplash is designed to remove the need for constructors or initializers by allowing the user to define the value of their variables directly during the deployment process. To ensure this process works reliably we perform input validation during the bundling process. Below we define the expected valid types. All other input types should be caught and rejected.
+
+Note that ChugSplash config files may be written in Javascript and therefore users are able to take advantage of various Javscript features and libraries to fetch values to be used in their ChugSplash config files.
+
+## Signed & Unsigned integers
+For both signed and unsigned integers, we allow two valid input types:
+- Base 10 string numbers: "1", "10", "1000000000000000000", "-100"
+- Base 10 literal numbers: 1, 10, 100, -100
```
sam-goldman, line 57: Add ethers.BigNumber?

specs/chugsplash-config.md
```diff
@@ -0,0 +1,359 @@
...
+boolean public x;
+boolean public y;
+
+// Javascript
+x: true,
+y: false,
+```
+
+## Addresses
+For addresses, we only accept hexadecimal literals that pass the address checksum test.
```
sam-goldman, line 111: Thoughts on changing this to:
> We only accept [`DataHexstrings`](https://docs.ethers.org/v5/api/utils/bytes/#DataHexString) that are 20 bytes long

specs/chugsplash-config.md
```diff
@@ -0,0 +1,359 @@
...
+// Solidity
+address public x;
+
+// Javascript
+x: '0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266',
+```
+
+## Contracts
+For contracts, we accept two possible input types:
+- Valid checksum addresses as defined above
```
sam-goldman, line 124: Should update this alongside the change to address types

specs/chugsplash-config.md
```diff
@@ -0,0 +1,359 @@
...
+UserDefinedType public x;
+
+// Javascript
+x: 1,
+```
+
+## Enums
+Similar to user defined types, we treat enums like their underlying value type uint8. We accept the following possible input formats:
+- Base 10 integers between 0-255 where the integer corresponds to a valid Solidity enum value.
+- Base 10 integer strings between 0-255 where the integer corresponds to a valid Solidity enum value.
```
sam-goldman, line 189: IMO enums should only be defined as integers, not strings. I don't really see a good reason to let them be strings

specs/chugsplash-config.md
```diff
@@ -0,0 +1,359 @@
...
+- Typescript enums where the underlying number falls between 0-255 and corresponds to a valid Solidity enum value.
```
sam-goldman, line 190: I don't think this is necessary to include given that TypeScript enums don't exist at runtime. Any variable in this spec could be a TypeScript type defined by the user, so it feels unnecessary to include this

specs/chugsplash-config.md
```diff
@@ -0,0 +1,359 @@
...
+type UserDefinedType is uint256;
+UserDefinedType public x;
+
+// Javascript
+x: 1,
+```
+
+## Enums
+Similar to user defined types, we treat enums like their underlying value type uint8. We accept the following possible input formats:
+- Base 10 integers between 0-255 where the integer corresponds to a valid Solidity enum value.
```
sam-goldman, line 188: Instead of 0-255, it should actually be something like "must be an integer N for which:
```
0 <= N < type(NameOfEnum).max
```
In other words, if there are three members in an enum, our validation logic should throw an error if the user enters `255`.

sam-goldman:
This spec should also have a section on the preserve keyword. Instead of explaining what it does, I think it'd be best to point to the [section in the docs](https://github.com/chugsplash/chugsplash/blob/develop/docs/special-var-defs.md#preserve-keyword) where we already do that

specs/chugsplash-config.md
```diff
@@ -0,0 +1,359 @@
...
+```js
+// Solidity
+bytes public x;
+
+// Javascript
+x: '0xabcd1234',
+```
+
+## Fixed-size Arrays
+For fixed length arrays, we accept JSON or JS arrays of exactly the defined length. The underlying array element types are validated individually and are expected to match the same input format as the corresponding value types.
```
sam-goldman, line 238: We currently allow users to define an empty array `[]` for fixed-size arrays. This is particularly useful for `__gap` variables, which are fixed-size arrays of length ~50 that appear in basically every upgradeable contract.
We have two choices:
1. Allow users to define `[]` for any fixed-size array. This is a footgun IMO because users should explicitly define each element in a fixed-size array the vast majority of the time.
2. Prevent users from defining `[]` for any fixed-size array. This is bad UX because users would need to write `__gap: [0, 0, ... 0]` for each gap variable. Also, since we currently create a SetStorage action for each config variable, we'd be creating roughly 50 SetStorage actions for each __gap variable.
3. Prevent users from defining `[]` for any fixed-size array _except_ `__gap` variables. Users simply cannot define `__gap` variables in their config.

I think 3 is the obvious choice. If you agree, we can leave this spec as is.

RPate97: We'll use a special keyword to resolve this which is an alias for a full fixed sized array full of zeros.

specs/chugsplash-config.md
RPate97: I disagree, I think typescript enums are a good thing for UX and we should explicitly state that we expect to support them here. However, you are correct that they do not represent a different input format at runtime so I do think this should be updated to make that clear.

sam-goldman: I suppose it depends on the goal of this spec. IMO this isn't meant to be a user-facing document; it should be to define precisely what we allow / don't allow for auditors and for us to test this code later. Adding something like a TypeScript enum just adds an unnecessary layer of indirection. Something like that should definitely be included in user-facing docs; not here IMO.

In other words, why would we include a bullet point about TypeScript enums and not general user-defined TypeScript types?

PR #560, opened 2023-03-15
feat(ct) add more gas optimization for loop

marc-aurele-besner:
### Purpose

Sorry for this second PR, looking at the contracts with my colleagues, one of them pointed out there was more optimization to be made on my original PR

- Defining memory outside of loop and so using warm memory in the loop instead of cold
- Removing i = 0 since it's the default value
- Applying this loop logic to new loop

### DIFF
--- a/packages/contracts/contracts/ChugSplashManager.sol
+++ b/packages/contracts/contracts/ChugSplashManager.sol
@@ -494,9 +494,13 @@ contract ChugSplashManager is OwnableUpgradeable, ReentrancyGuardUpgradeable {
             "ChugSplashManager: incorrect number of targets"
         );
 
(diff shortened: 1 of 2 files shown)


sam-goldman:
Hey @marc-aurele-besner, I want to let you know that we decided to close this PR. The reason is that we're optimizing for readability instead of gas efficiency in this version of ChugSplash. This is because the system we're building is already quite complex, and we want it to be as easy as possible for anyone to read through the code that handles their deployments. For this reason, I the `unchecked` blocks in your previous PR as well. In #565, I included the parts of your PRs that define memory outside of the for-loops, since that improves gas efficiency without compromising readability. Hope you understand!

PR #561, opened 2023-03-15
feat(core): Split encode variable into parsing and encoding processes

RPate97:
## Purpose
- Splits the encode variable function into separate parsing and encoding processes using a new abstraction to avoid duplicating the recursive structure. 
- Adds a framework for catching and displaying all errors that occur during the parsing process

## Note
This PR explicitly does not address adding additional input validation. That will come in a follow-up PR. 

Blocked by #557


packages/core/src/config/parse.ts
```diff
@@ -171,14 +198,507 @@ export const assertValidUserConfigFields = (config: UserChugSplashConfig) => {
...
+  return parsedArray
+}
+
+/**
+ * Handles parsing and validating fixed-size arrays
+ *
+ * @param props standard VariableHandler props. See ./iterator.ts for more information.
+ * @returns
+ */
+export const parseInplaceArray: VariableHandler<ParsedConfigVariable> = (
```
sam-goldman, line 265: I think this should either be `parseAndValidateInplaceArray` or the call below should be `parseArrayElements` for consistency

packages/core/src/config/parse.ts
```diff
@@ -171,14 +198,507 @@ export const assertValidUserConfigFields = (config: UserChugSplashConfig) => {
...
+
 /**
  * Parses a ChugSplash config file from the config file given by the user.
  *
  * @param userConfig Unparsed config file to parse.
  * @param env Environment variables to inject into the file.
  * @return Parsed config file with template variables replaced.
  */
-export const parseChugSplashConfig = async (
+const parseAndValidateChugSplashConfig = async (
```
sam-goldman, line 701: add `export`?

RPate97: This function is not supposed to be used outside this library. Instead, we should use `readParsedChugSplashConfig` which includes more validation and calls `parseAndValidateChugSplashConfig`.

packages/core/src/config/parse.ts
RPate97: What other functions are you referring too?

sam-goldman: Did we decide whether to make everything `parseAndValidate`, `parse`, or keep as is?

sam-goldman (approved):

PR #567, opened 2023-03-17
feat(core): Add complete config variable input validation

RPate97:
## Purpose
- Adds complete input validation for config variables
- Adds support for BigNumbers
- Adds integration tests for Immutables, BigNumbers, and input validation errors

## Note
This PR does not include comprehensive input validation for constructor arguments. I've created a separate [ticket](https://linear.app/chugsplash/issue/CHU-146/add-constructor-argument-parsing-and-validation-to-chugsplash-config) for that.


packages/core/src/config/parse.ts
```diff
@@ -198,6 +209,10 @@ export const assertValidUserConfigFields = (config: UserChugSplashConfig) => {
   }
 }
 
+const stringifyVariableType = (variable: UserConfigVariable) => {
```
sam-goldman, line 212: `export`?

packages/core/src/config/parse.ts
```diff
@@ -262,11 +277,47 @@ export const parseArrayElements = (
...
+  }
+
+  // array object types come in the format: t_array(t_<type>)<size>_storage)
+  // when nested, the format is repeated: t_array(t_array(t_<type>)<size>_storage)<size>_storage)
+  // So to get the size of the array, we split on the ')' character, remove the first element (which is the type),
+  // remove the _storage suffix, and parse the remaining element as an integer.
+  let stringSizes = storageObj.type.split(')')
+  stringSizes = stringSizes.map((el) => el.replace('_storage', ''))
+  stringSizes.shift()
+  const sizes = stringSizes.map((el) => parseInt(el, 10))
```
sam-goldman, line 302: Probably cleaner to use a regex here. [This is how OZ does a similar thing](https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/master/packages/core/src/scripts/migrate-oz-cli-project.ts#L330). Not necessary to do now

packages/core/src/config/parse.ts
```diff
@@ -360,16 +425,40 @@ export const parseInplaceBytes: VariableHandler<ParsedConfigVariable> = (
...
+  const { variable, variableType, storageObj } = props
+
+  if (
+    typeof variable !== 'number' &&
+    typeof variable !== 'string' &&
+    !(
+      typeof variable === 'object' &&
+      'type' in variable &&
+      variable.type === 'BigNumber'
+    )
```
sam-goldman, line 440: Can use [`ethers.isBigNumber`](https://docs.ethers.org/v5/api/utils/bignumber/#BigNumber--BigNumber--methods--inspection)

packages/core/src/config/parse.ts
```diff
@@ -381,10 +470,28 @@ export const parseInplaceUint: VariableHandler<ParsedConfigVariable> = (
...
```
sam-goldman, line 485: > Can use [`ethers.isBigNumber`](https://docs.ethers.org/v5/api/utils/bignumber/#BigNumber--BigNumber--methods--inspection)

packages/core/src/config/parse.ts
RPate97: Believe it or not, this actually doesn't work. I think it's because of how we handle importing the config object.

packages/core/src/config/parse.ts
```diff
@@ -676,21 +832,119 @@ const parseContractVariables = (
...
+        `${incorrectConstructorArgNames.map((argName) => `\n${argName}`)}`
+    )
+  }
+
+  constructorFragment.inputs.forEach((input) => {
+    const constructorArgValue = userConstructorArgs[input.name]
+    if (constructorArgValue === undefined) {
+      throw new Error(
+        `User did not define the constructor argument '${input.name}' in the ChugSplash config file\n` +
+          `for ${referenceName}. Please include it in the 'constructorArgs' field in your ChugSplash config file.`
```
sam-goldman, line 933: Would be great to have the list of missing constructor args instead of throwing an error for the first missing one

sam-goldman:
Overall looks great. Left some minor comments, none of them blocking. I think we should make an effort to search for edge cases that pass the validation logic in a future PR. For example, [ICAP addresses](https://docs.ethers.org/v5/api/utils/address/#utils-getIcapAddress) pass the validation logic currently. I'd rather do that later though and get this PR through.

PR #570, opened 2023-03-18
maint(core): Bring rest of config validation into the parse config interface

RPate97:
## Purpose
- Brings the rest of the config related validation into the parse config interface
- Replaces `readUserChugSpalshConfig` and `readParsedChugSplashConfig` with `readValidationChugSplashConfig` and `readUnvalidatedChugSplashConfig`
- Introduces a new global object pattern

## Global Object Pattern
In this PR, I introduce a new pattern where we initialize a set of global variables at the beginning of our tasks to avoid passing the variables through a long chain of functions. This pattern allows me to avoid doing a large refactor to make sure the `readParsedChugSplashConfig` has the necessary variables to perform the full validation. In the future, we can simplify our existing logic by adding additional variables to the global object and removing them as parameters from our functions.

This pattern does have an important tradeoff though. It is significantly less observable than our current pattern of passing everything around as function parameters. Our code often fails to compile if we make a change that breaks this web of dependencies. 

If we perpetuate the global object pattern, we will be introducing a hidden dependency and it will be on us to remember that the global object always needs to be properly defined. So this may make debugging somewhat more difficult because we will have to remember that bugs could be related to the global object. However, using the global object will allow us to significantly reduce the complexity of our functions and make our code more maintainable.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -34,6 +34,7 @@
     "@eth-optimism/contracts-bedrock": "^0.8.0",
     "@eth-optimism/core-utils": "^0.9.1",
     "@ethersproject/abstract-provider": "^5.7.0",
+    "@openzeppelin/hardhat-upgrades": "^1.22.1",
     "@openzeppelin/upgrades-core": "^1.20.6",
     "core-js": "^3.27.1",
     "ethers": "^5.6.9",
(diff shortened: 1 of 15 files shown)


sam-goldman:
> ## Global Object Pattern

> If we perpetuate the global object pattern, we will be introducing a hidden dependency and it will be on us to remember that the global object always needs to be properly defined. So this may make debugging somewhat more difficult because we will have to remember that bugs could be related to the global object.

This concerns me. It sounds like this'll make it more difficult to audit and debug our code. It also sounds like our code may break in unexpected ways, which seems like it'd make it less maintainable over time. Why not use some sort of object that contains these variables instead? Hardhat uses the HRE for this purpose.

sam-goldman (approved):

PR #572, opened 2023-03-23
feat(core): Add support for other openzeppelin safety check options

RPate97:
## Purpose
- Adds support for other openzeppelin safety check options
- Moves `silent` and `stream` parameters onto the CRE (ChugSplashRuntimeEnvironment) and uses refactors all uses of them to reference the CRE.

I also updated the `chugsplashLog` function to handle more complex outputs and used that function to output some warnings when users enable the unsafe flags. In a followup PR, I'll be updating the rest of the parsing logic to output errors using the `chugsplashLog` function to make the overall UX better.

### DIFF
--- a/packages/contracts/contracts/interfaces/IChugSplashManager.sol
+++ b/packages/contracts/contracts/interfaces/IChugSplashManager.sol
@@ -1,0 +1,12 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.9;
+
+/**
+ * @title ChugSplashManager
+ * @notice Interface that must be inherited the ChugSplash manager.
+ */
+interface IChugSplashManager {
+    function activeBundleId() external returns (bytes32);
+
+    function proposers(address) external returns (bool);
+}
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -36,6 +36,7 @@
     "@ethersproject/abstract-provider": "^5.7.0",
     "@openzeppelin/hardhat-upgrades": "^1.22.1",
     "@openzeppelin/upgrades-core": "^1.20.6",
+    "chalk": "^4.1.2",
     "core-js": "^3.27.1",
     "ethers": "^5.6.9",
     "fs": "^0.0.1-security",
--- a/packages/core/src/config/parse.ts
+++ b/packages/core/src/config/parse.ts
@@ -10,6 +10,7 @@ import { Fragment } from 'ethers/lib/utils'
 import {
   assertStorageUpgradeSafe,
   StorageLayout,
+  ValidationErrors,
 } from '@openzeppelin/upgrades-core'
 import { OZ_UUPS_UPDATER_ADDRESS, ProxyABI } from '@chugsplash/contracts'
 import { getDetailedLayout } from '@openzeppelin/upgrades-core/dist/storage/layout'
@@ -31,10 +32,11 @@ import {
   getChugSplashManagerProxyAddress,
   getEIP1967ProxyAdminAddress,
   getPreviousStorageLayoutOZFormat,
-  getOpenZeppelinStorageLayout,
+  getOpenZeppelinUpgradableContract,
   isEqualType,
   getOpenZeppelinValidationOpts,
   getParentContractASTNodes,
+  chugsplashLog,
 } from '../utils'
 import {
   UserChugSplashConfig,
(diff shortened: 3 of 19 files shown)


packages/contracts/contracts/interfaces/IChugSplashManager.sol
```diff
@@ -0,0 +1,12 @@
...
+
+/**
+ * @title ChugSplashManager
+ * @notice Interface that must be inherited the ChugSplash manager.
+ */
+interface IChugSplashManager {
+    function activeBundleId() external returns (bytes32);
+
+    function proposers(address) external returns (bool);
+}
```
RPate97, line 12: I added these interfaces so that our foundry tests can just import them instead of the full contracts. This circumvents some issues I was encountering related to running the open zeppelin safety checks in the plugins package.

packages/core/package.json
```diff
@@ -36,6 +36,7 @@
     "@ethersproject/abstract-provider": "^5.7.0",
     "@openzeppelin/hardhat-upgrades": "^1.22.1",
     "@openzeppelin/upgrades-core": "^1.20.6",
+    "chalk": "^4.1.2",
```
RPate97, line 39: Allows for nicer console outputs with color and stuff. I use this for the new `chugsplashLog` function.

packages/core/src/utils.ts
```diff
@@ -975,24 +995,47 @@ export const toOpenZeppelinProxyType = (
...
-    unsafeSkipStorageCheck: false,
+    unsafeAllowRenames: contractConfig.unsafeAllowRenames ?? false,
+    unsafeSkipStorageCheck: contractConfig.unsafeSkipStorageCheck ?? false,
   }
+
+  return withValidationDefaults(options)
 }
 
-export const getOpenZeppelinStorageLayout = (
+export const getOpenZeppelinUpgradableContract = (
```
RPate97, line 1027: Since open zeppelin automatically runs the safety checks when you create a `UpgradableContract`, I decided to refactor this function so that it returns the entire upgradable contract so it could be used for both the storage layout check and the rest of the safety checks.

packages/core/src/config/parse.ts
```diff
@@ -1191,30 +1230,25 @@ permission to call the 'upgradeTo' function on each of them.
...
           newStorageLayout
         )
 
-        // We could check for the `skipStorageCheck` in the outer for-loop, but this makes it easy to
-        // support more granular storage layout config options in the future.
-        if (parsedConfig.options.skipStorageCheck !== true) {
-          // Run OpenZeppelin's storage slot checker.
+        if (
+          // If the user has disabled storage checks for all contracts
+          parsedConfig.options.skipStorageCheck !== true &&
```
sam-goldman, line 1241: I think we should remove `skipStorageCheck` from `options`. It doesn't really make sense for it to be the only OpenZeppelin option that's on the granularity of the entire config, especially considering `unsafeAllowRenames` also operates on the storage layouts

sam-goldman:
I think we can remove the UUPS `upgradeTo` check [here](https://github.com/chugsplash/chugsplash/blob/3cd22d5cb9da21b0f0247346dde171ffefc91e34/packages/core/src/config/parse.ts#L1256-L1282) since OpenZeppelin does it for us now

packages/core/src/utils.ts
```diff
@@ -975,24 +995,47 @@ export const toOpenZeppelinProxyType = (
...
+  contractConfig: UserContractConfig
 ): Required<ValidationOptions> => {
-  return {
+  type UnsafeAllow = Required<ValidationOptions>['unsafeAllow']
+
+  const unsafeAllow: UnsafeAllow = [
+    'state-variable-assignment',
+    'constructor',
+    'state-variable-immutable',
+    'missing-public-upgradeto',
```
sam-goldman, line 1007: > I think we can remove the UUPS `upgradeTo` check [here](https://github.com/chugsplash/chugsplash/blob/3cd22d5cb9da21b0f0247346dde171ffefc91e34/packages/core/src/config/parse.ts#L1256-L1282) since OpenZeppelin does it for us now

If you agree, we should remove `'missing-public-upgradeto`' here

RPate97:
> I think we can remove the UUPS `upgradeTo` check [here](https://github.com/chugsplash/chugsplash/blob/3cd22d5cb9da21b0f0247346dde171ffefc91e34/packages/core/src/config/parse.ts#L1256-L1282) since OpenZeppelin does it for us now

I agree, but I am going to do this in a separate PR because testing this is somewhat difficult.

packages/core/src/utils.ts
RPate97: See my other comment, will do this in a separate PR

packages/core/src/utils.ts
```diff
@@ -975,24 +995,47 @@ export const toOpenZeppelinProxyType = (
...
+  }
+  if (contractConfig.unsafeAllow?.selfdestruct) {
+    unsafeAllow.push('selfdestruct')
+  }
+
+  const options = {
     kind: toOpenZeppelinProxyType(proxyType),
-    unsafeAllow: [],
+    unsafeAllow,
     unsafeAllowCustomTypes: false,
```
sam-goldman, line 1019: We can remove this line since `unsafeAllowCustomTypes` defaults to false in `withValidationDefaults`

packages/core/src/utils.ts
```diff
@@ -975,24 +995,47 @@ export const toOpenZeppelinProxyType = (
...
-    unsafeAllowLinkedLibraries: false,
-    unsafeAllowRenames: false,
-    unsafeSkipStorageCheck: false,
+    unsafeAllowRenames: contractConfig.unsafeAllowRenames ?? false,
+    unsafeSkipStorageCheck: contractConfig.unsafeSkipStorageCheck ?? false,
```
sam-goldman, line 1021: >     unsafeAllowRenames: contractConfig.unsafeAllowRenames ?? false,
>    unsafeSkipStorageCheck: contractConfig.unsafeSkipStorageCheck ?? false,

We can remove `.. ?? false` for both of these

packages/core/src/config/parse.ts
```diff
@@ -1173,14 +1175,51 @@ permission to call the 'upgradeTo' function on each of them.
...
+    const userContractConfig = userConfig.contracts[referenceName]
+
+    // First we do some validation on the contract that doesn't depend on whether or not we're performing an upgrade
+    // the validation happens automatically when we call `getOpenZeppelinUpgradableContract`
+
+    // In addition to doing validation the `getOpenZeppelinUpgradableContract` function also outputs some warnings related to
+    // the provided override options. We want to output our own warnings, so we temporarily disable console.error.
+    const tmp = console.error
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    console.error = () => {}
```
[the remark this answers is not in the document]
RPate97: This doesn't work b/c `getOpenZeppelinUpgradableContract` is not actually throwing an error, it's just printing to stderr.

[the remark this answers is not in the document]
RPate97: I agree this solution is janky though, so open to alternative solutions. To my knowledge, this is the only way to deal with the issue short of bringing all of the OpenZeppelin `UpgradableContract` logic into our codebase so we can modify it to remove the logging.

sam-goldman (approved):

PR #573, opened 2023-03-23
feat(core): Display all parsing validation errors

RPate97:
## Purpose
Refactors the parsing logic so that as many errors as possible are output at the same time.

There are a few cases where if a particular issue is detected, it actually stops the rest of the parsing logic from working properly without significant refactoring. In those cases, we continue to throw errors and stop execution but we will still output all errors detected up to that point.

### DIFF
--- a/packages/core/src/config/parse.ts
+++ b/packages/core/src/config/parse.ts
@@ -10,7 +10,7 @@ import { Fragment } from 'ethers/lib/utils'
 import {
   assertStorageUpgradeSafe,
   StorageLayout,
-  ValidationErrors,
+  UpgradeableContractErrorReport,
 } from '@openzeppelin/upgrades-core'
 import { OZ_UUPS_UPDATER_ADDRESS, ProxyABI } from '@chugsplash/contracts'
 import { getDetailedLayout } from '@openzeppelin/upgrades-core/dist/storage/layout'
@@ -66,10 +66,23 @@ import {
 import { ChugSplashRuntimeEnvironment } from '../types'
 
 class InputError extends Error {
-  constructor(message) {
+  constructor(message: string) {
     super(message)
     this.name = 'InputError'
   }
+}
+
+let validationErrors = false
+
+const logValidationError = (
+  logLevel: 'warning' | 'error' = 'warning',
+  title: string,
+  lines: string[],
+  silent: boolean,
+  stream: NodeJS.WritableStream
+) => {
+  validationErrors = true
+  chugsplashLog(logLevel, title, lines, silent, stream)
 }
 
 /**
@@ -83,7 +96,8 @@ export const readValidatedChugSplashConfig = async (
   configPath: string,
   artifactPaths: ArtifactPaths,
   integration: Integration,
-  cre: ChugSplashRuntimeEnvironment
+  cre: ChugSplashRuntimeEnvironment,
+  exitOnFailure: boolean = true
 ): Promise<ParsedChugSplashConfig> => {
   const userConfig = await readUnvalidatedChugSplashConfig(configPath)
   return parseAndValidateChugSplashConfig(
@@ -91,7 +105,8 @@ export const readValidatedChugSplashConfig = async (
     userConfig,
     artifactPaths,
     integration,
-    cre
+    cre,
+    exitOnFailure
   )
 }
 
@@ -128,11 +143,10 @@ export const isEmptyChugSplashConfig = (configFileName: string): boolean => {
  *
  * @param config Config file to validate.
  */
-export const assertValidUserConfigFields = (config: UserChugSplashConfig) => {
-  if (config.contracts === undefined) {
-    throw new Error('contracts field must be defined in ChugSplash config')
-  }
-
+export const assertValidUserConfigFields = (
+  config: UserChugSplashConfig,
+  cre: ChugSplashRuntimeEnvironment
+) => {
   const referenceNames: string[] = Object.keys(config.contracts)
 
   for (const [referenceName, contractConfig] of Object.entries(
@@ -140,8 +154,12 @@ export const assertValidUserConfigFields = (config: UserChugSplashConfig) => {
   )) {
     // Block people from accidentally using templates in contract names.
     if (referenceName.includes('{') || referenceName.includes('}')) {
-      throw new Error(
-        `cannot use template strings in reference names: ${referenceName}`
+      logValidationError(
+        'error',
+        `Cannot use template strings in reference names: ${referenceName}`,
+        [],
+        cre.silent,
+        cre.stream
       )
     }
 
@@ -150,8 +168,12 @@ export const assertValidUserConfigFields = (config: UserChugSplashConfig) => {
       contractConfig.contract.includes('{') ||
       contractConfig.contract.includes('}')
     ) {
-      throw new Error(
-        `cannot use template strings in contract names: ${contractConfig.contract}`
+      logValidationError(
+        'error',
+        `Cannot use template strings in contract name: ${contractConfig.contract}`,
+        [],
+        cre.silent,
+        cre.stream
       )
     }
 
@@ -160,8 +182,12 @@ export const assertValidUserConfigFields = (config: UserChugSplashConfig) => {
       contractConfig.externalProxy !== undefined &&
       !ethers.utils.isAddress(contractConfig.externalProxy)
     ) {
-      throw new Error(
-        `external proxy address is not a valid address: ${contractConfig.externalProxy}`
+      logValidationError(
+        'error',
+        `External proxy address is not a valid address: ${contractConfig.externalProxy}`,
+        [],
+        cre.silent,
+        cre.stream
       )
     }
 
@@ -170,8 +196,12 @@ export const assertValidUserConfigFields = (config: UserChugSplashConfig) => {
       contractConfig.externalProxyType !== undefined &&
       isExternalProxyType(contractConfig.externalProxyType) === false
     ) {
-      throw new Error(
-        `External proxy type is not valid: ${contractConfig.externalProxyType}`
+      logValidationError(
+        'error',
+        `External proxy type is not valid ${contractConfig.externalProxyType}`,
+        [],
+        cre.silent,
+        cre.stream
       )
     }
 
@@ -180,17 +210,23 @@ export const assertValidUserConfigFields = (config: UserChugSplashConfig) => {
       contractConfig.externalProxy !== undefined &&
       contractConfig.externalProxyType === undefined
     ) {
-      throw new Error(
-        `User included an 'externalProxy' field for ${contractConfig.contract} in ${config.options.projectName},\n` +
-          `but did not include an 'externalProxyType' field. Please include both or neither.`
+      logValidationError(
+        'error',
+        `User included an 'externalProxy' field, but did not include an 'externalProxyType'\nfield for ${contractConfig.contract}. Please include both or neither.`,
+        [],
+        cre.silent,
+        cre.stream
       )
     } else if (
       contractConfig.externalProxy === undefined &&
       contractConfig.externalProxyType !== undefined
     ) {
-      throw new Error(
-        `User included an 'externalProxyType' field for ${contractConfig.contract} in ${config.options.projectName},\n` +
-          `but did not include an 'externalProxy' field. Please include both or neither.`
+      logValidationError(
+        'error',
+        `User included an 'externalProxyType' field, but did not include an 'externalProxy'\nfield for ${contractConfig.contract}. Please include both or neither.`,
+        [],
+        cre.silent,
+        cre.stream
       )
     }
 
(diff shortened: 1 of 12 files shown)


packages/core/src/config/parse.ts
```diff
@@ -1077,21 +1194,27 @@ export const assertValidParsedChugSplashFile = async (
...
   } = cre
 
-  // Determine if the deployment is an upgrade
-  const projectName = parsedConfig.options.projectName
+  let isValid = true
+
+  const openzeppelinStorageLayouts = await fetchOpenZeppelinStorageLayouts(
+    cre.hre,
+    userConfig
+  )
```
sam-goldman, line 1212: Is there a reason we include the `hre` inside of the `cre` now? IMO it's more straightforward to call `importOpenZeppelinStorageLayouts` inside of `createChugSplashRuntime`, which is how we did this previously. I think it's a cleaner approach, plus we don't need to pass around the HRE inside of the core package

packages/core/src/utils.ts
```diff
@@ -1016,8 +1044,9 @@ export const getOpenZeppelinValidationOpts = (
   const options = {
     kind: toOpenZeppelinProxyType(proxyType),
     unsafeAllow,
-    unsafeAllowRenames: contractConfig.unsafeAllowRenames,
-    unsafeSkipStorageCheck: contractConfig.unsafeSkipStorageCheck,
+    unsafeAllowCustomTypes: false,
+    unsafeAllowRenames: contractConfig.unsafeAllowRenames ?? false,
+    unsafeSkipStorageCheck: contractConfig.unsafeSkipStorageCheck ?? false,
```
sam-goldman, line 1049: Seems like these weren't merged correctly? This was updated in #572

packages/plugins/src/foundry/index.ts
```diff
@@ -76,6 +76,11 @@ const command = args[0]
         'foundry',
         cre
       )
+
+      if (config === false) {
```
sam-goldman, line 80: It'd be nice if we handled this case inside of `readValidatedChugSplashConfig`. For example, we could `process.exit()`. That way, we don't need to always remember to handle this case, and we can get rid of all of these if-statements.

sam-goldman:
This PR uses this pattern in a bunch of places: `let isValid = true; isValid = false`. This seems error-prone and difficult to maintain because we need to ensure we set `isValid = false` in dozens of if-statements. I think a better pattern is what OpenZeppelin does in their `getErrors` function, which is called inside of the constructor of their `UpgradeableContract` class. They use a `ValidationError[]` array to accumulate the error messages / warnings.

RPate97:
> This PR uses this pattern in a bunch of places: `let isValid = true; isValid = false`. This seems error-prone and difficult to maintain because we need to ensure we set `isValid = false` in dozens of if-statements. I think a better pattern is what OpenZeppelin does in their `getErrors` function, which is called inside of the constructor of their `UpgradeableContract` class. They use a `ValidationError[]` array to accumulate the error messages / warnings.

I agree with the sentiment and have refactored the logic to make this more maintainable by removing most of the logic passing around the `isValid` variable. However, I did not go all the way to using a  `ValidationError[]` array. Instead, I used a global variable `validationErrors`. That variable is updated to true if any error is output. If it is true at the end of the parsing logic then we exit with code 1. 

I chose to do it this way because we cannot guarantee that the complete parsing logic will always be executed. There are several cases where invalid user inputs make the rest of the parsing logic impossible to execute (not literally, we'd just have to do some more significant refactoring to handle them). By logging the errors as we go, we're still able to output everything up until those cases are hit. The alternative is that we output nothing except the last error.

I'm open to coming back and handling those specific cases more cleanly to make the full process guaranteed to execute completely, but I'd rather not make this PR more complex than it already is. For now, I believe this is fine.

sam-goldman:
> > This PR uses this pattern in a bunch of places: `let isValid = true; isValid = false`. This seems error-prone and difficult to maintain because we need to ensure we set `isValid = false` in dozens of if-statements. I think a better pattern is what OpenZeppelin does in their `getErrors` function, which is called inside of the constructor of their `UpgradeableContract` class. They use a `ValidationError[]` array to accumulate the error messages / warnings.
> 
> I agree with the sentiment and have refactored the logic to make this more maintainable by removing most of the logic passing around the `isValid` variable. However, I did not go all the way to using a `ValidationError[]` array. Instead, I used a global variable `validationErrors`. That variable is updated to true if any error is output. If it is true at the end of the parsing logic then we exit with code 1.
> 
> I chose to do it this way because we cannot guarantee that the complete parsing logic will always be executed. There are several cases where invalid user inputs make the rest of the parsing logic impossible to execute (not literally, we'd just have to do some more significant refactoring to handle them). By logging the errors as we go, we're still able to output everything up until those cases are hit. The alternative is that we output nothing except the last error.
> 
> I'm open to coming back and handling those specific cases more cleanly to make the full process guaranteed to execute completely, but I'd rather not make this PR more complex than it already is. For now, I believe this is fine.

Got it, that makes sense. Maybe a sufficient solution is to have a few different ValidationError arrays that are grouped logically? That way, we wouldn't need to do a significant refactor to guarantee that the full process parses correctly. I'm good with saving that for a future PR

packages/core/src/config/parse.ts
```diff
@@ -857,15 +907,20 @@ export const parseAndValidateVariable = (
...
   contractConfig: UserContractConfig,
   storageLayout: SolidityStorageLayout,
-  compilerOutput: CompilerOutput
+  compilerOutput: CompilerOutput,
+  cre: ChugSplashRuntimeEnvironment
 ): {
-  [name: string]: ParsedConfigVariable
+  variables: {
+    [name: string]: ParsedConfigVariable
+  }
```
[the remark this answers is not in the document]
sam-goldman: Looks like this change may have been undone? Also a few lines below too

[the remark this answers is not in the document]
RPate97: huh, weird

packages/core/src/config/parse.ts
```diff
@@ -913,40 +939,74 @@ const parseContractVariables = (
...
+        `The following variables were defined in the contract ${contractConfig.contract} (or one of its parent contracts) but were not defined in the ChugSplash config file:`,
+        lines,
+        cre.silent,
+        cre.stream
+      )
+    }
   }
 
-  return parsedConfigVariables
+  return { variables: parsedConfigVariables }
```
sam-goldman, line 1009: I think this can be simplified to `parsedConfigVariables`

packages/core/src/config/parse.ts
```diff
@@ -857,15 +884,14 @@ export const parseAndValidateVariable = (
 const parseContractVariables = (
   contractConfig: UserContractConfig,
   storageLayout: SolidityStorageLayout,
-  compilerOutput: CompilerOutput
+  compilerOutput: CompilerOutput,
+  cre: ChugSplashRuntimeEnvironment
 ): {
-  [name: string]: ParsedConfigVariable
+  variables: ParsedConfigVariables
```
sam-goldman, line 890: > I think this can be simplified to `return parsedConfigVariables`

Similarly, I think this can be simplified from `{ variables: ParsedConfigVariables }` to just `ParsedConfigVariables`

packages/core/src/config/parse.ts
```diff
@@ -66,12 +66,25 @@ import {
...
+  constructor(message: string) {
     super(message)
     this.name = 'InputError'
   }
 }
 
+let validationErrors = false
+
+const logValidationError = (
+  logLevel: 'warning' | 'error' = 'warning',
```
sam-goldman, line 78: Is this the same as `logLevel: 'warning' | 'error'`? It appears to be when I hover over it in VSCode. If so, could simplify it

packages/core/src/config/parse.ts
```diff
@@ -857,15 +884,14 @@ export const parseAndValidateVariable = (
...
-  [name: string]: ParsedConfigVariable
+  variables: ParsedConfigVariables
 } => {
-  const parsedConfigVariables: {
-    [name: string]: ParsedConfigVariable
-  } = {}
+  const parsedConfigVariables: ParsedConfigVariables = {}
   if (!contractConfig.variables) {
-    return {}
+    return { variables: {} }
```
sam-goldman, line 894: > > I think this can be simplified to `return parsedConfigVariables`
> 
> Similarly, I think this can be simplified from `{ variables: ParsedConfigVariables }` to just `ParsedConfigVariables`

Likewise, this'd just be `{}`

packages/core/src/config/parse.ts
```diff
@@ -962,8 +1022,9 @@ const parseContractVariables = (
 export const parseAndValidateConstructorArgs = (
   userConstructorArgs: UserConfigVariables,
   referenceName: string,
-  abi: Array<Fragment>
-): ParsedConfigVariables => {
+  abi: Array<Fragment>,
+  cre: ChugSplashRuntimeEnvironment
+): { args: ParsedConfigVariables } => {
```
sam-goldman, line 1027: > > I think this can be simplified to `return parsedConfigVariables`
> 
> Similarly, I think this can be simplified from `{ variables: ParsedConfigVariables }` to just `ParsedConfigVariables`

Same thing here

packages/core/src/config/parse.ts
```diff
@@ -1077,21 +1159,15 @@ export const assertValidParsedChugSplashFile = async (
...
-) => {
-  const {
-    canonicalConfigPath,
-    remoteExecution,
-    autoConfirm,
-    openzeppelinStorageLayouts,
-  } = cre
+): Promise<{
+  upgrade: boolean
+}> => {
```
sam-goldman, line 1164: Could we just return a boolean here instead of an object containing a boolean?

packages/core/src/config/parse.ts
```diff
@@ -1500,12 +1589,22 @@ const parseAndValidateChugSplashConfig = async (
...
+      ) {
+        logValidationError(
+          'error',
+          `Entered a proxy address that does not exist: ${userContractConfig.externalProxy}`,
+          [],
+          cre.silent,
+          cre.stream
+        )
+      }
+    } catch {
```
sam-goldman, line 1605: The try/catch seems unnecessary here. We could just check if the proxy's valid via `ethers.utils.isAddress`, then call `logValidationError` if it's invalid

packages/core/src/tasks/index.ts
```diff
@@ -76,7 +77,7 @@ import {
 export const chugsplashRegisterAbstractTask = async (
   provider: ethers.providers.JsonRpcProvider,
   signer: ethers.Signer,
-  parsedConfig: ParsedChugSplashConfig,
+  parsedConfig: UserChugSplashConfig | ParsedChugSplashConfig,
```
sam-goldman, line 80: nit: `parsedConfig` -> `config`

packages/plugins/src/hardhat/artifacts.ts
```diff
@@ -99,43 +98,38 @@ export const getArtifactPaths = async (
...
       )
-      layouts[referenceName] = storageLayout
+      return storageLayout
     }
   }
 
-  return layouts
+  throw new Error(
+    'Should not attempt to import OpenZeppelin storage layout for non-OpenZeppelin proxy type. Please report this to the developers.'
+  )
```
sam-goldman, line 134: I don't think we should throw an error here because it'll prevent the error at the end of `getPreviousStorageLayoutOZFormat` from ever getting triggered. Instead, I think the cleanest way to handle this is to return `Promise<StorageLayout | undefined>` from this function, and throw the error at the end of `getPreviousStorageLayoutOZFormat`

packages/plugins/src/hardhat/deployments.ts
```diff
@@ -47,7 +47,7 @@ export const deployAllChugSplashConfigs = async (
   silent: boolean,
   ipfsUrl: string,
   fileNames?: string[]
-) => {
+): Promise<boolean> => {
```
sam-goldman, line 50: What's the purpose of returning a boolean here? It doesn't look like we ever return false

packages/plugins/src/hardhat/deployments.ts
```diff
@@ -71,7 +71,7 @@ export const deployAllChugSplashConfigs = async (
 
     // Skip this config if it's empty.
     if (isEmptyChugSplashConfig(configPath)) {
-      return
+      return true
```
sam-goldman, line 74: It seems like this should be `continue` instead of `return`

packages/plugins/src/utils.ts
```diff
@@ -1,10 +1,19 @@
...
-  readUnvalidatedChugSplashConfig,
+  ParsedContractConfig,
 } from '@chugsplash/core'
 import { HardhatRuntimeEnvironment } from 'hardhat/types'
 
-import { importOpenZeppelinStorageLayouts } from './hardhat/artifacts'
+import { importOpenZeppelinStorageLayout } from './hardhat/artifacts'
+
+const fetchOpenZeppelinStorageLayout = async (
+  hre: HardhatRuntimeEnvironment | undefined = undefined,
```
sam-goldman, line 10: If you agree with my comments above, we can remove this function and just use `importOpenZeppelinStorageLayout`

packages/plugins/src/hardhat/artifacts.ts
```diff
@@ -99,43 +98,38 @@ export const getArtifactPaths = async (
...
       const manifest = await Manifest.forNetwork(hre.network.provider)
       const deployData = await getDeployData(
         hre,
-        await hre.ethers.getContractFactory(userContractConfig.contract),
+        await hre.ethers.getContractFactory(parsedContractConfig.contract),
         getOpenZeppelinValidationOpts(
-          userContractConfig.externalProxyType ?? 'internal-default',
-          userContractConfig
+          parsedContractConfig.proxyType,
+          parsedContractConfig
```
sam-goldman, line 120: It looks like `getOpenZeppelinValidationOpts` should only accept `UserContractConfig` types, but strangely TypeScript doesn't throw an error when we pass in a `ParsedContractConfig` here

packages/core/src/tasks/index.ts
```diff
@@ -76,28 +77,28 @@ import {
 export const chugsplashRegisterAbstractTask = async (
   provider: ethers.providers.JsonRpcProvider,
   signer: ethers.Signer,
-  parsedConfig: ParsedChugSplashConfig,
+  config: UserChugSplashConfig | ParsedChugSplashConfig,
```
sam-goldman, line 80: It actually looks like we can replace this with `userConfig: UserChugSplashConfig` because this type is a subset of the `ParsedChugSplashConfig` type

sam-goldman (approved):

PR #577, opened 2023-03-27
fix(core): Add bytes32(0) set storage action for mappings

RPate97:
## Purpose
Explicitly sets mapping primary storage slot value to bytes32(0)

### DIFF
--- a/packages/core/src/languages/solidity/storage.ts
+++ b/packages/core/src/languages/solidity/storage.ts
@@ -461,7 +461,13 @@ export const encodeMapping: VariableHandler<
   } = props
 
   // Iterate over every key/value in the mapping to get the storage slot pair for each one.
-  let slots: Array<StorageSlotSegment> = []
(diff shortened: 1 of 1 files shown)


packages/core/src/languages/solidity/storage.ts
```diff
@@ -461,7 +461,13 @@ export const encodeMapping: VariableHandler<
   } = props
 
   // Iterate over every key/value in the mapping to get the storage slot pair for each one.
-  let slots: Array<StorageSlotSegment> = []
+  let slots: Array<StorageSlotSegment> = [
+    {
+      key: slotKey,
+      offset: 0,
+      val: padHexSlotValue('0x00', 0),
```
sam-goldman, line 468: nit: `padHexSlotValue('0x00', 0)` -> ethers.constants.HashZero

sam-goldman (approved):

PR #580, opened 2023-03-29
feat(core): Support deploying stateless non-proxied contracts

RPate97:
## Purpose
- Adds support for deploying non-proxied contracts as long as they are stateless. 
- Fixes a bug where contract references only work as long as the referenced contract is declared before the contract reference. You can now use contract references out of order.


packages/core/src/config/parse.ts
```diff
@@ -1022,7 +1039,8 @@ export const parseAndValidateConstructorArgs = (
   referenceName: string,
   abi: Array<Fragment>,
   cre: ChugSplashRuntimeEnvironment
-): ParsedConfigVariables => {
+): { args: ParsedConfigVariables; valid: boolean } => {
+  let validArguments = true
```
RPate97, line 1043: See comment below for more context on why I'm returning a valid boolean again for this function.

packages/core/src/config/parse.ts
```diff
@@ -1655,10 +1688,78 @@ const parseAndValidateChugSplashConfig = async (
...
+          getCreationCodeWithConstructorArgs(bytecode, args, referenceName, abi)
+
+        contracts[referenceName] = getContractAddress(
+          parsedConfig.options.projectName,
+          creationCodeWithConstructorArgs
+        )
+      }
+    } else {
+      contracts[referenceName] = proxy
+    }
```
RPate97, line 1743: Here we call the `parseAndValidateConstructorArgs` function to perform validation on the constructor arguments. I've reintroduced the valid return value pattern for this specific case because valid constructor arguments are necessary to calculate the address for non-proxied contracts. If we pass invalid arguments to the `getContractAddress` function then we get an esoteric error and the program crashes. 

This is kind of like some of the other cases where some validation failing causes the rest of the logic to not work correctly. But in those cases, we just throw a useful error and exit the program entirely when the relevant validation fails to avoid outputting weird errors. In this case, 99% of the time if the constructor argument validation fails it's fine because the arguments are only used for non-proxied contracts. So I didn't want to just exit the program if the constructor args aren't valid. Instead, I decided to just track if the constructor args are invalid, and use the zero address if so. This allows the rest of the parsing logic to work fine, and the program will exit at the end due to the constructor arguments being invalid.

packages/core/src/utils.ts
```diff
@@ -328,19 +329,42 @@ export const displayProposerTable = (proposerAddresses: string[]) => {
...
+              bytecode,
+              contractConfig.constructorArgs,
+              referenceName,
+              abi
+            )
+          address = getContractAddress(
+            parsedConfig.options.projectName,
+            creationCodeWithConstructorArgs
+          )
+        }
```
sam-goldman, line 359: I think we can refactor the logic that retrieves `address` for unproxied contracts. It looks like we repeat this pattern of `readContractArtifact` -> `getCreationCodeWithConstructorArgs` -> `getContractAddress` in a few functions: `displayDeploymentTable`, `parseAndValidateChugSplashConfig`, and `foundry/index.ts` (under `getAddress`). If we refactor this into a single function, we can use a ternary operator for the `address`, which IMO is cleaner than how we do it now

packages/plugins/src/hardhat/deployments.ts
```diff
@@ -152,16 +154,37 @@ export const getContract = async (
...
+      bytecode,
+      contractConfig.constructorArgs || {},
+      referenceName,
+      abi
+    )
+    address = getContractAddress(
+      userConfig.options.projectName,
+      creationCodeWithConstructorArgs
+    )
+  }
```
sam-goldman, line 177: IMO it's slightly cleaner to only assign the value of `address` once instead of re-assigning it if we're dealing with an unproxied contract

packages/core/src/config/parse.ts
```diff
@@ -1655,10 +1688,78 @@ const parseAndValidateChugSplashConfig = async (
     contracts: {},
   }
 
+  // Resolve all the contracts upfront so they can be used handle contract references regardless or ordering.
+  // We also cache the compiler output, constructor args, and other artifacts so we don't have to re-read them later when we complete the parsing
   const contracts = {}
+  const cachedCompilerOutput = {}
+  const cachedConstructorArgs = {}
+  const cachedArtifacts = {}
```
sam-goldman, line 1696: > I think we can refactor the logic that retrieves address for unproxied contracts. It looks like we
> repeat this pattern of readContractArtifact -> getCreationCodeWithConstructorArgs ->
> getContractAddress in a few functions: displayDeploymentTable, parseAndValidateChugSplashConfig, and
> foundry/index.ts (under getAddress). If we refactor this into a single function, we can use a
> ternary operator for the address, which IMO is cleaner than how we do it now

If you agree with this change, I think we should remove these `cached` variables in favor of a simple ternary operator for the value of `contracts[referenceName]`. It'd make this code simpler, and I don't think it'd make it meaningfully slower given that everything's synchronous

packages/core/src/config/parse.ts
```diff
@@ -1682,64 +1783,47 @@ const parseAndValidateChugSplashConfig = async (
...
-      cre
-    )
+    // Only run the parsing logic if the contract is proxied
+    let parsedVariables
+    if (userContractConfig.kind !== 'no-proxy') {
+      const variables = parseContractVariables(
+        JSON.parse(
+          Handlebars.compile(JSON.stringify(userContractConfig))({
+            ...contracts,
+          })
```
sam-goldman, line 1811: Instead of calling `Handlebars.compile` inside this for-loop, I think we should call it before the for-loop. It's only necessary to resolve the contract references once

packages/core/src/languages/solidity/types.ts
```diff
@@ -97,7 +97,7 @@ export interface CompilerOutputMetadata {
 
 export interface CompilerOutputContract {
   abi: Array<Fragment>
-  storageLayout: SolidityStorageLayout
+  storageLayout: SolidityStorageLayout | undefined
```
sam-goldman, line 100: When could this be undefined? Doesn't the logic in `readStorageLayout` ensure that it's not undefined?

packages/core/src/config/parse.ts
sam-goldman: IMO we should throw errors for invalid constructor args up-front instead of following this pattern.

> 99% of the time if the constructor argument validation fails it's fine because the arguments are only used for non-proxied contracts.

Constructor args are used for immutable variables in proxied contracts, so I expect they'll be used very frequently

Also, it seems like an anti-pattern to rely on logic outside of `parseAndValidateConstructorArgs` to throw an error for invalid constructor args. It's not clear to me where exactly this error would eventually be thrown in our codebase

packages/core/src/config/parse.ts
```diff
@@ -1264,13 +1287,23 @@ permission to call the 'upgradeTo' function on each of them.
...
+        // Since 'no-proxy' was introduced by us and not supported by OpenZeppelin, we need
+        // to pass in 'internal-default' instead so we can still run the validation on non-proxied contracts
+        contractConfig.kind !== 'no-proxy'
+          ? contractConfig.kind
+          : 'internal-default',
+        userContractConfig
+      )
+    } catch (e) {
+      throw e
+    }
```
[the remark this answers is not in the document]
RPate97: We can't skip it, we need it for the storage slot comparison.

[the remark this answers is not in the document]
RPate97: The try/catch here doesn't have anything to do with non-proxied contracts. I added it in because I realized that by overriding console.error, we're introducing the possibility that some unexpected error will be thrown and we miss it which could make debugging difficult.

As far as running the `getOpenZeppelinUpgradableContract`, IMO we should still run it for non-proxied contracts because there are checks that we do that are relevant to them. I.e checks for self-destruct and delegatecall. I think we should be consistent about the checks we're running for all contracts.

packages/core/src/config/parse.ts
RPate97: >> IMO we should throw errors for invalid constructor args up-front instead of following this pattern.
We do, the error is caused by invalid constructor arguments and the issue is logged during the `parseAndValidateConstructorArgs` function. As a result, the program will cleanly exit after the rest of the parsing logic is finished running.

>> Constructor args are used for immutable variables in proxied contracts, so I expect they'll be used very frequently
Not within the parsing logic which is the section of the code that I am referring to. In the parsing logic, we use the constructor arguments only to calculate the address of stateless immutable contracts. If the constructor arguments are invalid, then no other part of the codebase is relevant because the program will end after the parsing logic is completed.

>> it seems like an anti-pattern to rely on logic outside of `parseAndValidateConstructorArgs` to throw an error for invalid constructor args
I agree, that's exactly what I am trying to avoid with this logic. If we call `getCreationCodeWithConstructorArgs` with invalid arguments that an esoteric error will be thrown due to the arguments not matching the ABI. This logic avoids that by not calculating the address if and only if the arguments are valid. 

An alternative solution would be to throw an error and completely stop the program anytime an invalid constructor argument is detected in `parseAndValidateConstructorArgs`, but that doesn't seem like a great idea to me for two reasons:
1. This issue is not relevant to proxied contracts which means we'd be stopping execution early for no reason the vast majority of the time. 
2. We've explicitly set the goal of trying to output as many errors to the user at once as possible like a compiler. So I don't think it makes sense to deviate from that as long as it doesn't require large amounts of refactoring.

sam-goldman: Got it, I understand now. IMO we should still throw a separate set of errors for invalid constructor arguments. Here's my reasoning:
1. The code in its current state is non-intuitive to me. It took me a little while to understand why exactly we're doing it this way. That may just be a "me problem", but I wouldn't be surprised if future readers struggle to understand what's going on too
2. The issue of invalid constructor arguments will become a more frequent occurrence if/when we support stateful non-proxied contracts.
3. Compilers don't output all classes of errors at once. If you mess around with a Solidity contract, you'll see what I mean. Generally speaking, we should strive to output as many errors at once, but I think the tradeoff here isn't worth it.

Happy to hop on a call if you don't agree

packages/contracts/contracts/ChugSplashManager.sol
```diff
@@ -648,24 +656,36 @@ contract ChugSplashManager is OwnableUpgradeable, ReentrancyGuardUpgradeable {
...
-            require(adapter != address(0), "ChugSplashManager: proxy type has no adapter");
+            require(
+                action.proxyTypeHash == noProxyTypeHash || adapter != address(0),
+                "ChugSplashManager: proxy type has no adapter"
+            );
+            require(
+                action.proxyTypeHash != noProxyTypeHash ||
+                    action.actionType != ChugSplashActionType.SET_STORAGE,
+                "ChugSplashManager: cannot set storage in non-proxied contracts"
+            );
```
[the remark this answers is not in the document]
RPate97: Hmm, yeah I can see you're point.

I think the counterpoint is that we may want the rule that immutable contracts cannot have state to be an actual invariance. That's definitely how I was thinking about it and that's why this require statement is here. Without it, the statelessness part is just a rule of thumb. The deployment should fail when setStorage is called, but the target contract could implement the setStorage function in which case the invariance is violated.

packages/core/src/config/parse.ts
```diff
@@ -1604,64 +1705,47 @@ const parseAndValidateChugSplashConfig = async (
...
-
-    const args = parseAndValidateConstructorArgs(
-      constructorArgs ?? {},
-      referenceName,
-      compilerOutput.contracts[sourceName][contractName].abi,
-      cre
-    )
+    // Only run the parsing logic if the contract is proxied
+    let parsedVariables
+    if (userContractConfig.kind !== 'no-proxy') {
```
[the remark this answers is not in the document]
RPate97: Well, the variables field isn't guaranteed to be empty in this case. We do detect that it's not empty and output an error because of that earlier on, but it doesn't stop execution. So if we just run this logic then the user will get additional errors related to the variables that don't make much sense.

packages/plugins/src/hardhat/deployments.ts
```diff
@@ -154,16 +156,37 @@ export const getContract = async (
...
     getDefaultProxyAddress(userConfig.options.projectName, referenceName)
-  if ((await isContractDeployed(proxyAddress, hre.ethers.provider)) === false) {
+
+  if (contractConfig.kind === 'no-proxy') {
+    const { abi, bytecode } = hre.artifacts.readArtifactSync(
+      contractConfig.contract
+    )
+    const creationCodeWithConstructorArgs = getCreationCodeWithConstructorArgs(
+      bytecode,
+      contractConfig.constructorArgs || {},
```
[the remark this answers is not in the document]
RPate97: good catch

packages/core/src/config/parse.ts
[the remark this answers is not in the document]
sam-goldman: I think it's worth spending some time thinking about how we can clean this up, but let's do it in a future PR. I made a ticket for it. For this PR, I think the only change to this code snippet is to skip calling `getOpenZeppelinUpgradableContract` for non-proxied contracts.

packages/core/src/config/parse.ts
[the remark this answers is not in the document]
sam-goldman: Since we're splitting the constructor arg parsing into its own function and returning early if there are errors, can we go through with this change?

packages/contracts/contracts/ChugSplashManager.sol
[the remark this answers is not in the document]
sam-goldman: This LGTM

packages/core/src/config/parse.ts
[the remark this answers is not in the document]
RPate97: No, splitting the constructor arg parsing into a separate function doesn't resolve this issue. I'm making it a bit cleaner though.

packages/core/src/actions/types.ts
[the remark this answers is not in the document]
RPate97: `contract` is a reserved keyword, so we have to use something else.

packages/contracts/contracts/ChugSplashManager.sol
```diff
@@ -987,39 +1007,42 @@ contract ChugSplashManager is OwnableUpgradeable, ReentrancyGuardUpgradeable {
...
+     * @param _code          Creation bytecode of the contract.
      */
-    function _deployImplementation(string memory _referenceName, bytes memory _code) internal {
+    function _deployContract(string memory _referenceName, bytes memory _code) internal {
         // Calculate the salt for the Create2 call. Note that there can be address collisions
         // between implementations if their reference names are the same, but this is avoided with
         // off-chain tooling by skipping implementations that have the same reference name and
         // creation bytecode.
-        bytes32 salt = keccak256(bytes(_referenceName));
+        bytes32 salt = bytes32(0);
```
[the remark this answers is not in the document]
sam-goldman: Sorry, when I said hard-code, I meant removing the `bytes32 salt` definition, and instead doing this:
```
address expectedAddress = Create2.compute(address(this), bytes32(0), _code);
...
assembly {
    actualAddress := create2(0x0, add(_code, 0x20), mload(_code), 0x0)
}
```

packages/core/src/utils.ts
```diff
@@ -815,38 +829,62 @@ export const isEqualType = (
...
   projectName: string,
   referenceName: string,
-  creationCodeWithConstructorArgs: string
+  constructorArgs: UserConfigVariables,
+  maybeArtifact:
+    | ContractArtifact
+    | {
+        integration: Integration
+        artifactPaths: ArtifactPaths
+      }
```
sam-goldman, line 851: I think we can simplify this by adding a `creationCodeWithConstructorArgs` field on the CRE. We can save that for a future PR though

RPate97: Ya know, I thought about doing this, but IMO, I don't think we use this value enough for it to be worth adding to the CRE. I'm open to it though if there are other things it would simplify.

sam-goldman: We call `getContractAddress` ~7 times in our codebase, so I think it'd be worth it. It'd make this function less janky IMO

sam-goldman (approved):

PR #581, opened 2023-03-30
fix(ct): Fix security vulnerability on setStorage function

RPate97:
## Purpose
Fixes a vulnerability where an attacker could set arbitrary storage slots during execution. The issue is caused by the fact that the updater can be publicly available if it takes more than one block to execute a deployment. So we need to have logic on the setStorage function which prevents some arbitrary third party from calling it. 

This vulnerability was handled when we introduced the Updater pattern by introducing the ChugSplashAdmin and restricting access to the setStorage function to only proxy admins or ChugSplash admins depending on the proxy type. The vulnerability appears to have been reintroduced when a change was made to how we handle the setStorage action.

I resolved the issue by moving the ChugSplashAdmin logic onto the base ProxyUpdater and using it to secure the setStorage action for both updater types. This also involved adding some additional logic to the adapters ensuring that the initiate and complete functions are called for all proxy types because those functions are used to set and unset the ChugSplashAdmin at the beginning and end of execution.

### DIFF
--- a/packages/contracts/contracts/updaters/DefaultUpdater.sol
+++ b/packages/contracts/contracts/updaters/DefaultUpdater.sol
@@ -45,6 +45,10 @@ contract DefaultUpdater is ProxyUpdater {
         }
     }
 
+    function setStorage(bytes32 _key, uint8 _offset, bytes memory _segment) external ifAdmin {
+        super.setStorageValue(_key, _offset, _segment);
+    }
+
     receive() external payable {
         revert("DefaultUpdater: caller is not an admin");
     }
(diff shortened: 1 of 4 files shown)


sam-goldman:
I like the idea of standardizing the ownership mechanism with this pattern. However it does lead to redundancy for any proxy type besides UUPS, since we now have two admin slots: eip-1967 and chugsplash. An alternative approach is to use whichever ownership mechanism that a proxy type already uses. So, for eip1967 proxies, we'd just use `ADMIN_KEY`, which is already defined in the DefaultUpdater. Do you have a preference for either approach? I'm slightly in favor of the alternative approach, since it'd remove redundancy in most cases, but I'm not adamant about it

sam-goldman:
The more I think about it, the more in favor I am of the alternate approach. It'd make the code simpler to understand for the vast majority of cases (i.e. everything except UUPS). IMO it makes the most sense to introduce the concept of an owner key only if one isn't already present. Happy to hop on a call if you disagree

sam-goldman:
Looks a lot cleaner!

sam-goldman (approved):

PR #590, opened 2023-04-09
feat(core): make chugsplash-deploy task execute locally

sam-goldman:
Notably, this PR removes local execution logic from the @chugsplash/executor package. In other words, the standalone executor is only used for executing remote bundles. This is because 95% of the logic in the standalone executor is unnecessary for local execution. Only two function calls inside the standalone executor are necessary: `executeTask` and `verifyChugSplashConfig`. Another issue with keeping remote and local execution together is that the local executor will detect bundles that should be executed by the remote executor.


packages/demo/package.json
```diff
@@ -45,6 +45,8 @@
     "solhint-plugin-prettier": "^0.0.5"
   },
   "dependencies": {
+    "@0xsequence/create3": "https://github.com/0xsequence/create3",
```
RPate97, line 48: Doesn't look like this is actually used anywhere

packages/plugins/src/foundry/index.ts
```diff
@@ -289,10 +282,7 @@ const command = args[0]
       const silent = args[5] === 'true'
       const outPath = cleanPath(args[6])
       const buildInfoPath = cleanPath(args[7])
-      const withdrawFunds = args[8] === 'true'
       let newOwner = args[9]
```
RPate97, line 285:
```suggestion
      let newOwner = args[8]
```

RPate97 (commented):
Yeah, I think this is fine. I've created a separate ticket to add automated testing of the remote executor. I mostly just reviewed the execution logic which looks solid, but I left some notes on a few minor things I noticed as well.

PR #602, opened 2023-04-10
feat(ct): Support opt-in manager upgrades

RPate97:
## Purpose
Adds support for opt-in manager upgrades

Note: This PR does not include automated integration tests as this feature is somewhat difficult to test due to our mutisig being the default owner of the registry. We could add logic to make the registry owner configurable for tests, but that seems overly complex to me. Instead, I think we should just cover this in our contract unit tests.


packages/contracts/contracts/ChugSplashRegistry.sol
```diff
@@ -113,81 +114,80 @@ contract ChugSplashRegistry is Ownable, Initializable {
...
-            executorPaymentPercentage,
-            protocolPaymentPercentage
+        ChugSplashManagerProxy managerProxy = new ChugSplashManagerProxy{ salt: salt }(
+            address(this),
+            address(this),
+            initialManagerVersion
+        );
+        managerProxy.upgradeToAndCall(
+            _version,
+            abi.encodeWithSignature("initialize(bytes)", _data)
```
sam-goldman, line 181: I think we can put the `initialize(bytes)` function in `IChugSplashManager`, then import the interface into this contract and do `abi.encodeCall(IChugSplashManager.initialize, ...)`

packages/contracts/contracts/ChugSplashManagerProxy.sol
```diff
@@ -0,0 +1,75 @@
...
+
+    /**
+     * @param _registry              The ChugSplashRegistry's address.
+     * @param _admin                 Owner of this contract.
+     * @param _initialImplementation The initial implementation of this contract. This will always
+     *                               be the V1 ChugSplashManager to ensure that manager proxy
+     *                               addresses are predictable.
+     */
+    constructor(
+        address _registry,
```
sam-goldman, line 28: nit: This can be `ChugSplashRegistry _registry` to reduce the number of type conversions

packages/contracts/contracts/ChugSplashRegistry.sol
```diff
@@ -113,81 +114,80 @@ contract ChugSplashRegistry is Ownable, Initializable {
...
-        uint256 _ownerBondAmount,
-        uint256 _executionLockTime,
-        uint256 _executorPaymentPercentage,
-        uint256 _protocolPaymentPercentage
-    ) {
-        ownerBondAmount = _ownerBondAmount;
-        executionLockTime = _executionLockTime;
-        executorPaymentPercentage = _executorPaymentPercentage;
-        protocolPaymentPercentage = _protocolPaymentPercentage;
-
```
sam-goldman, line 151: Good idea removing these

packages/contracts/contracts/ChugSplashRegistry.sol
```diff
@@ -113,81 +114,80 @@ contract ChugSplashRegistry is Ownable, Initializable {
...
-        ChugSplashManager manager = new ChugSplashManager{ salt: salt }(
-            this,
-            executionLockTime,
-            ownerBondAmount,
-            executorPaymentPercentage,
-            protocolPaymentPercentage
+        ChugSplashManagerProxy managerProxy = new ChugSplashManagerProxy{ salt: salt }(
+            address(this),
+            address(this),
+            initialManagerVersion
```
sam-goldman, line 177: I think `initialManagerVersion` should be kept out of the constructor because it could potentially mess up the `ChugSplashManagerProxy`'s Create2 address. For example, say we need to slightly change the bytecode of the initial ChugSplashManager to support an L2. This would change the Create2 address of the ChugSplashManager address, which would in turn cause all of the `ChugSplashManagerProxy` addresses on that chain to be different.

If you agree, I think we can remove `initialManagerVersion` as a state variable

packages/core/src/constants.ts
```diff
@@ -77,23 +72,44 @@ export const CHUGSPLASH_REGISTRY_ADDRESS = utils.getCreate2Address(
...
+]
+
+const [managerConstructorFragment] = ChugSplashManagerABI.filter(
+  (fragment) => fragment.type === 'constructor'
+)
+const managerConstructorArgTypes = managerConstructorFragment.inputs.map(
+  (input) => input.type
+)
+
+export const CHUGSPLASH_MANAGER_ADDRESS = utils.getCreate2Address(
```
sam-goldman, line 90: nit: `CHUGSPLASH_MANAGER_V1_ADDRESS`?

packages/core/src/utils.ts
```diff
@@ -236,11 +234,19 @@ export const claimChugSplashProject = async (
     (await ChugSplashRegistry.projects(claimerAddress, organizationID)) ===
     constants.AddressZero
   ) {
+    // Encode the initialization arguments for the ChugSplashManager contract.
+    // Note: Future versions of ChugSplash may require different arguments encoded in this way.
+    const regiserData = ethers.utils.defaultAbiCoder.encode(
```
sam-goldman, line 239: nit: `regiserData` -> `claimData` or `initializerData`

sam-goldman (approved):

PR #603, opened 2023-04-11
Spec: Add Constructor Arg Specification

RPate97:
## Purpose
Completes the specification of the ChugSplash config file including:
- Specification for the constructor arguments
- Specification for the other input options

Also fixes a minor issue with the plugin package tests.

### DIFF
--- a/specs/chugsplash-config.md
+++ b/specs/chugsplash-config.md
@@ -21,32 +21,83 @@ ChugSplash config files may be written in Javascript, Typescript, or JSON. For J
 - [Structs](#structs)
 - [Arbitary Nested Types](#arbitary-nested-types)
 
-## Options
-- Project name
-- todo spec out other project options
+# Options
+There are three config option fields `organizationID`, `claimer`, and `projectName`. Each of these values impacts the addresses of contracts deployed using ChugSplash. All three of these fields are required for ChugSplash to uniquely identify the correct organization, project, and contract when performing a deployment or upgrade.
 
-## Contracts Config Field
+### Organization Id
+A 32 byte string which identifies the organization that the relevant ChugSplash config is a part of. Each organization gets a unique ChugSplashManager contract which manages deployments for their organization. Users of the ChugSplash managed service have an organization id recommended to them, but other users can just generate and use any arbitrary 32 byte string.
+
+You can think of the organization id as identifying the toplevel address space of the users organization.
+
+## Project Name
+An abitrary name string which identifies a specific project within an Organization. Project names are unique within each organization. If you create multiple configs in the same organization that use the same project name, then they will be treated like they are part of the same project when you perform deployments. We recommend that project names not be shared between multiple config files.
+
+You can think of the project name as identifying a subsection of the address space of the organization that the project is a part of. Project names allow ChugSplash to reuse the same ChugSplashManager within an organization while avoid the possibility of different projects deployments and upgrades impacting each other.
+
+## Claimer
+The address of the wallet used to claim the organization on chain. Since ChugSplash deterministically deploys contracts to the same addresses across chains, we have to account for the possibility that a malicious third party could attempt to snipe the contract addresses of well know protocols on chains they have not deployed to yet. To avoid this attack vector, we use the address of the claimer wallet as the final element in determining the address of the ChugSplash manager. This makes it impossible to snipe a ChugSplash users contract address without owning the account used to claim the organization.
+
+# Contracts Config Field
 JSON or Javascript object defining the contracts that should be deployed. Each key is a contracts reference name and its value is a contract definition.
 
-> Reference names are used internally by ChugSplash to identify specific instances of contracts. Each combination of a project name and contract reference name should be unique within the ChugSplash system.
+> Reference names identify specific instances of contracts in the ChugSplash system. Each combination of a project name and contract reference name identifies a unique contract within an organization.
 
-## Contract Config Definition
-JSON or Javascript object defining a contract to be deployed, it's constructor args, and variable definitions. Contract definitions must conform to the following format:
+# Contract Config Definition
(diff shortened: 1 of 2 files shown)


specs/chugsplash-config.md
```diff
@@ -21,32 +21,90 @@ ChugSplash config files may be written in Javascript, Typescript, or JSON. For J
 - [Structs](#structs)
 - [Arbitary Nested Types](#arbitary-nested-types)
 
-## Options
-- Project name
-- todo spec out other project options
+# Options
+There are three config option fields `organizationID`, `projectName`, and `claimer`. Each of these values impacts the address of the users ChugSplashManager which in turn impacts the address of their deployed contracts. So all three of these fields are required for ChugSplash to uniquely identify the correct organization, project, and contract when performing a deployment or upgrade.
```
sam-goldman, line 25: The `projectName` actually doesn't impact the address of the ChugSplashManager

specs/chugsplash-config.md
```diff
@@ -21,32 +21,90 @@ ChugSplash config files may be written in Javascript, Typescript, or JSON. For J
...
-ReferenceName: {
-  contract: 'LiteralContractName',
-  constructorArgs: {},
-  variables: {}
-}
-```
+You can think of the project name as identifying a subsection of the address space of the organization that the project is a part of. Project names allow ChugSplash to reuse the same ChugSplashManager within an organization while avoid the possibility of different projects deployments and upgrades impacting each other.
+
+## Claimer
+The address of the wallet used to claim the organization on chain. Since ChugSplash deterministically deploys contracts to the same addresses across chains, we have to account for the possibility that a malicious third party could attempt to snipe the contract addresses of well know protocols on chains they have not deployed to yet. To avoid this attack vector, we use the address of the claimer wallet as the final element in determining the address of the ChugSplash manager. This makes it impossible to snipe a ChugSplash users contract address without owning the contract used to claim the organization.
```
sam-goldman, line 38: "without owning the contract" -> "without owning the account"

Since a contract or EOA can claim it

specs/chugsplash-config.md
```diff
@@ -21,32 +21,90 @@ ChugSplash config files may be written in Javascript, Typescript, or JSON. For J
...
 
-## Contract Constructor Args
-A JSON or Javascript object defining a set of constructor args. All constructor arguements are required to be immutable variables...
+# Contracts Config Field
+JSON or Javascript object defining the contracts that should be deployed. Each key is a contracts reference name and its value is a contract definition.
 
-TODO finish defining constructor args spec
+> Reference names are used internally by ChugSplash to identify specific instances of contracts. Each combination of a project name and contract reference name should be unique within the ChugSplash system.
```
sam-goldman, line 43: > Reference names are used internally by ChugSplash to identify specific instances of contracts.

Maybe shorten this to: "Reference names identify specific instances of contracts in the ChugSplash system."

Since they're also used externally by users

specs/chugsplash-config.md
```diff
@@ -21,32 +21,90 @@ ChugSplash config files may be written in Javascript, Typescript, or JSON. For J
...
+A path to a build info file generated for a previous version of a contract that has already been deployed. This field needs to be defined when upgrading a contract that uses a proxy which was deployed outside of ChugSplash. Providing the previous build info allows us to run our storage slot checker
+
+## Previous Fully Qualified Name
+The fully qualified name of a contract that has been deployed outside of ChugSplash. This field needs to be defined along with the previous build info and is used to run the storage slot checker against the previous contract version.
+
+## Unsafe
+ChugSplash performs a variety of safety checks before deploying/upgrading contracts. Users may override these safety checks using this field if their situation requires it. The unsafe field is a JSON or JS object containing a set of fields which disable different safety checks including:
+- allowRenames: Allow renaming state variables
+- skipStorageCheck: Fully disable the storage slot checker
+- delegatecall: Allow use of delegatecall in implementation contracts
```
sam-goldman, line 88: nit: For consistency with OZ's tool, consider prepending "allow" to delegatecall, selfdestruct, and missingPublicUpgradeTo

specs/chugsplash-config.md
```diff
@@ -21,32 +21,90 @@ ChugSplash config files may be written in Javascript, Typescript, or JSON. For J
...
+For constructor arguments we accept a JSON or Javascript object defining a set of constructor arguments. Since constructor arguments are expected to only be used to set immutable variables, we only support the following value types:
+- Signed and Unsigned integers
+- bytes32
+- boolean
+- addresses
+- User defined types
+
+For all of these types, we expect the same input format as the corresponding variable input format which is defined below.
+
+Note that for addresses, we support the use of contract references in constructor arguments as long as the target contract is proxied. We do not support contract references in the constructor arguements of non-proxied contracts.
```
sam-goldman, line 105: This is sort of separate to this PR, but I actually think the rule should be slightly different: contract references can only contain proxied contracts. In other words, `{{ MyNonProxiedContract }}` isn't allowed, but this is allowed:
```
MyNonProxiedContract: {
  constructorArgs: {
    {{ MyProxy }}
  }
}
```

I think it's fine to define the above config because we can resolve the contract references for proxies in advance without worrying about ordering.

RPate97: Yeah, I think that's a good change and we should update this spec to include it.

sam-goldman:
Is this spec for the user config or parsed config? It looks like there are fields from both types here

RPate97:
> Is this spec for the user config or parsed config? It looks like there are fields from both types here

This is for the user config. There is obviously some overlap between that and the parsed config, but to my knowledge, nothing here is exclusive to the parsed config.

specs/chugsplash-config.md
RPate97: Actually, after thinking about this more I don't think we should make this change. I'll post more on the relevant ticket.

PR #607, opened 2023-04-17
maint(core): Import libraries from dependencies

RPate97:
## Purpose
Removes `Proxy.sol`, `MerkleTree.sol`, and `Create2.sol` in favor of importing them from dependencies.


packages/contracts/contracts/ChugSplashManager.sol
```diff
@@ -1023,7 +1027,11 @@ contract ChugSplashManager is OwnableUpgradeable, ReentrancyGuardUpgradeable {
      */
     function _deployContract(string memory _referenceName, bytes memory _code) internal {
         // Get the expected address of the contract.
-        address expectedAddress = Create2.compute(address(this), bytes32(0), _code);
+        address expectedAddress = Create2.computeAddress(
+            bytes32(0),
+            keccak256(_code),
+            address(this)
```
sam-goldman, line 1033: nit: We can use [this version](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/Create2.sol#L44-L46) of `computeAddress`, which doesn't require a third argument

packages/contracts/contracts/ChugSplashManager.sol
```diff
@@ -391,10 +393,12 @@ contract ChugSplashManager is OwnableUpgradeable, ReentrancyGuardUpgradeable {
     ) public view returns (address payable) {
         return (
             payable(
-                Create2.compute(
-                    address(this),
+                Create2.computeAddress(
```
sam-goldman, line 396: nit: We can use [this version](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/Create2.sol#L44-L46) of `computeAddress`, which doesn't require a third argument

sam-goldman (approved):

PR #610, opened 2023-04-20
fix(core,pg): only remotely execute when proposing on a live network

sam-goldman:
This PR makes `remoteExecution` true only if the user is calling the propose task on a live network. In all other cases, `remoteExecution` is false.


packages/core/src/utils.ts
```diff
@@ -1288,3 +1289,10 @@ export const getChainId = async (
   const network = await provider.getNetwork()
   return network.chainId
 }
+
+export const isLiveNetwork = async (
+  provider: providers.Provider
+): Promise<boolean> => {
+  const network = await provider.getNetwork()
+  return network.name !== 'unknown'
+}
```
RPate97, line 1298: This is very problematic b/c it relies on ethers.js to be aware of the target network. I just was testing with the polygon zkevm testnet and this returned false b/c ethers.js doesn't explicitly support that network yet. 

I think we basically have two options to resolve this issue:
1. Change how this function works to more accurately infer whether the target network is a live network or not. Maybe we could check if impersonate user works or something like that. 
2. Reconsider whether we really should be inferring this value. Given that the deploy task always performs the deployment locally and the other tasks are designed more for performing remote executions, maybe we can simply set `remoteExecution` to false in the deploy task and true in all other tasks.

PR #613, opened 2023-04-20
fix(core): Detect local networks with hardhat_impersonateAccount

RPate97:
## Purpose
Uses hardhat_impersonateAccount to detect if the target network is a local or live network.

### DIFF
--- a/packages/core/src/utils.ts
+++ b/packages/core/src/utils.ts
@@ -1291,10 +1291,18 @@ export const getChainId = async (
 }
 
 export const isLiveNetwork = async (
-  provider: providers.Provider
+  provider: providers.JsonRpcProvider
 ): Promise<boolean> => {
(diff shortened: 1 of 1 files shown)


packages/core/src/utils.ts
```diff
@@ -1291,10 +1291,16 @@ export const getChainId = async (
...
-  provider: providers.Provider
+  provider: providers.JsonRpcProvider
 ): Promise<boolean> => {
-  const network = await provider.getNetwork()
-  return network.name !== 'unknown'
+  try {
+    await provider.send('hardhat_impersonateAccount', [
+      ethers.constants.AddressZero,
+    ])
+  } catch (err) {
```
sam-goldman, line 1300: nit: Maybe add this documentation:
```
// This RPC method works on anvil because it's an alias for anvil_impersonateAccount
```

PR #624, opened 2023-04-23
fix(ct): skip deploying a contract if it already exists

sam-goldman:


packages/contracts/contracts/updaters/ProxyUpdater.sol
```diff
@@ -48,6 +45,8 @@ abstract contract ProxyUpdater is IProxyUpdater {
      *                 bytesN value, where N is in the range [1, 32] (inclusive).
      */
     function setStorageValue(bytes32 _key, uint8 _offset, bytes memory _segment) internal {
+        require(_segment.length <= 32, "ProxyUpdater: segment is too large");
```
sam-goldman, line 48: If the `_segment` is longer than 32 bytes, then our encoding logic is broken, and this function will fail regardless. We might as well get a useful error message if that happens.

PR #627, opened 2023-04-23
maint(core): Remove extra CLI tasks

RPate97:
## Purpose
Removes some CLI tasks which are only needed by managed service users. We will later replicate these features in the UI instead. They include:
- add proposer
- approve
- fund
- list proposers
- monitor
- withdraw


sam-goldman:
Command/find "monitor" in the repo. Is there a reason to keep any of this?

sam-goldman:
I noticed that the functions `successfulProposalMessage` and `alreadyProposedMessage` mention approving and funding the bundle via the CLI

sam-goldman:
It looks like `monitorExecution` mentions the `chugsplash-fund` task. If you don't want to remove this function, we should at least change that error message.

sam-goldman:
We should remove the relevant variables [here](https://github.com/chugsplash/chugsplash/blob/f67844959951cc26afa9a5040b3b2540fda80ec7/packages/plugins/src/hardhat/tasks.ts#L56-L80
)

sam-goldman (commented):
Looks good, just left a few comments about stuff that we might want to remove

RPate97:
@sam-goldman 
So I've been working on the gas estimation PR and I've realized that we've introduced a lot of changes that make it quite difficult to test the remote executor. So I've decided to put this PR on hold until after we've resolved that issue since quite a bit of this logic may turn out to be necessary for that.

PR #628, opened 2023-04-24
feat(pg): Only initialize ChugSplash on local networks

RPate97:
## Purpose
Prevents users from accidentally deploying ChugSplash on unsupported networks. ChugSplash will only be deployed automatically on local networks. If a user attempts to run ChugSplash on a live network where the contracts have not been deployed, we output an error stating that the network is not currently supported. 

Also fixes a bug where attempting to deploy ChugSplash against a network where it has already been deployed fails due to the `addVersion` step.

### DIFF
--- a/packages/core/src/languages/solidity/predeploys.ts
+++ b/packages/core/src/languages/solidity/predeploys.ts
@@ -53,6 +53,22 @@ import {
   CHUGSPLASH_MANAGER_V1_ADDRESS,
 } from '../../constants'
 
+export const ensureChugSplashInitialized = async (
+  provider: ethers.providers.JsonRpcProvider,
+  signer: ethers.Signer
+) => {
+  if (await isLiveNetwork(provider)) {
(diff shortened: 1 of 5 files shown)


packages/core/src/languages/solidity/predeploys.ts
```diff
@@ -269,12 +285,20 @@ export const initializeChugSplash = async (
...
+          await getGasPriceOverrides(provider)
+        )
+      ).wait()
+    } catch (e) {
+      // Silence the error if the version is already set.
+      // This error occurs if we attempt to deploy ChugSplash on a network where it's already been deployed.
+      if (!e.message.includes('version already set')) {
+        throw e
+      }
+    }
```
sam-goldman, line 301: nit: We'll lose some ETH every time this call fails on a live network, so, instead, we could remove the try/catch and only call `addVersion` if `registry.managerImplementations(ChugSplashManager.address)` returns `false`

sam-goldman (approved):

PR #631, opened 2023-04-24
fix(core): Improve execution gas cost estimation

RPate97:
## Purpose
Improves our off-chain gas cost estimation logic
- Only includes the cost of deploying a proxy if a proxy has not yet been deployed
- Uses the actual cost of deploying the default proxy instead of a fixed value
- Does not include the cost of deploying any implementation contracts that have already been deployed
- Reduces the estimated cost of each set storage action to 45k gas
- Adds a fixed buffer of 400k to account for the initiate and complete function calls

## Note
While this estimation logic is substantially better than what we had before, it is not perfect. We could improve it by adding more complex logic to estimate the cost of the initiate and complete function calls since those have a variable cost in practice. That being said, I think this is sufficient for now given that we are using a federated executor model and we expect anyone using remote execution to be a user of the managed service. We should come back to this if we find that it is not accurate enough for our immediate use case, or when we want to move to a trustless network of executors. 

## Testing
Unfortunately, this is somewhat difficult to test since we're no longer using this logic in the deploy task. We could do automated testing but that doesn't seem worth it to me in this case. So instead, I tested this by temporarily modifying the deploy task to call the `estimateExecutionCost` function, printing the result, and comparing that to the actual cost which was calculated based on the signers' balance before and after the deployment. I then deployed a bunch of different configs to ensure it worked well in a variety of situations.

### DIFF
--- a/packages/contracts/src/ifaces.ts
+++ b/packages/contracts/src/ifaces.ts
@@ -14,6 +14,7 @@ export const OZUUPSOwnableAdapterArtifact = require('../artifacts/contracts/adap
 export const OZUUPSAccessControlAdapterArtifact = require('../artifacts/contracts/adapters/OZUUPSAccessControlAdapter.sol/OZUUPSAccessControlAdapter.json')
 export const OZTransparentAdapterArtifact = require('../artifacts/contracts/adapters/OZTransparentAdapter.sol/OZTransparentAdapter.json')
 export const DefaultGasPriceCalculatorArtifact = require('../artifacts/contracts/DefaultGasPriceCalculator.sol/DefaultGasPriceCalculator.json')
+export const DefaultProxyArtifact = require('../node_modules/@eth-optimism/contracts-bedrock/artifacts/contracts/universal/Proxy.sol/Proxy.json')
 
 const directoryPath = path.join(__dirname, '../artifacts/build-info')
 const fileNames = fs.readdirSync(directoryPath)
--- a/packages/core/src/actions/artifacts.ts
+++ b/packages/core/src/actions/artifacts.ts
@@ -130,7 +130,6 @@ export const writeDeploymentArtifacts = async (
       )
       const { constructorArgValues } = getConstructorArgs(
         parsedConfig.contracts[referenceName].constructorArgs,
-        referenceName,
         abi
       )
       const { metadata, storageLayout } =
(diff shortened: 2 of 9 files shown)


sam-goldman:
Did you check that `estimateExecutionGas` gives a somewhat accurate upper bound for the gas cost of the plugins contract(s), demo contract, and/or RabbitHole's project? (with `remoteExecution = true`)

sam-goldman (approved):

sam-goldman:
> While this estimation logic is substantially better than what we had before, it is not perfect... That being said, I think this is sufficient for now given that we are using a federated executor model and we expect anyone using remote execution to be a user of the managed service.

Even though we're using a federated model, we need to ensure that this logic gives us an estimate that's higher than the actual gas cost. Otherwise, it's possible for us to tell the user to send an insufficient amount of funds to cover a deployment, which will cause the executor to skip deploying the bundle. This is especially true on mainnet, where gas spikes are common. Even though they'll be using the managed service for this feature, they're going to rely on us to give them a reasonable estimate.

RPate97:
> Otherwise, it's possible for us to tell the user to send an insufficient amount of funds to cover a deployment, which will cause the executor to skip deploying the bundle. 

I was actually thinking we'd loosen up the executors filters a bit such that it may accept some bundles that it won't necessarily be able to execute at a profit just to avoid this scenario ever happening. I think that's reasonable to do in the context of running a managed service where we're expecting to monetize separately.

sam-goldman:
> I was actually thinking we'd loosen up the executors filters a bit such that it may accept some bundles that it won't necessarily be able to execute at a profit just to avoid this scenario ever happening. I think that's reasonable to do in the context of running a managed service where we're expecting to monetize separately.


That sounds like a griefing risk, no? Is there anything stopping an attacker outside of the managed service who makes the executor execute underfunded bundles?

sam-goldman:
The only thing you should consider adding from #629 is `calldataGasUsed`. The inputs to the three execution functions in the ChugSplashManager comprise the vast majority of the calldata. I think this info is already available in the `estimateExecutionGas` function _.

I should mention that the calldata is usually a decent fraction of the total gas cost, and it accounted for >1M gas in the `executeActions` call that executed ~175 `SetStorage` actions in `Storage.sol`.

RPate97:
@sam-goldman 
I've been working on testing this PR with remote execution = true and found that quite a few changes have been introduced that make testing that case quite difficult. So I've decided to close this PR until that issue is resolved.

PR #634, opened 2023-04-26
fix(ct): use general create2 contract

sam-goldman:
This allows us to support non-standard Create2 formulas used by networks like ZK Sync.

Also includes some changes to fit the ChugSplashManager under the contract size limit.

### DIFF
--- a/packages/contracts/contracts/ChugSplashManager.sol
+++ b/packages/contracts/contracts/ChugSplashManager.sol
@@ -15,7 +15,6 @@ import { Proxy } from "@eth-optimism/contracts-bedrock/contracts/universal/Proxy
 import { ChugSplashRegistry } from "./ChugSplashRegistry.sol";
 import { IChugSplashManager } from "./interfaces/IChugSplashManager.sol";
 import { IProxyAdapter } from "./interfaces/IProxyAdapter.sol";
-import { Create2 } from "@openzeppelin/contracts/utils/Create2.sol";
 import {
     Lib_MerkleTree as MerkleTree
 } from "@eth-optimism/contracts/libraries/utils/Lib_MerkleTree.sol";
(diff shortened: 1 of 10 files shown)


packages/contracts/contracts/ChugSplashRegistry.sol
```diff
@@ -131,27 +130,21 @@ contract ChugSplashRegistry is Ownable, Initializable {
...
-        // Deploy the ChugSplashManager proxy and set the implementation to the requested version
         ChugSplashManagerProxy managerProxy = new ChugSplashManagerProxy{ salt: salt }(
             this,
             address(this)
         );
 
         require(
-            address(managerProxy) == managerProxyAddress,
-            "ChugSplashRegistry: manager proxy not deployed correctly"
+            address(managerProxy) != address(0),
```
sam-goldman, line 141: I believe this `!= address(0)` check is sufficient to ensure that the create2 call was successful. [OpenZeppelin does the same thing.](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/Create2.sol#L37)

packages/contracts/contracts/ChugSplashManager.sol
```diff
@@ -36,20 +36,22 @@ contract ChugSplashManager is
...
-    bytes32 public constant NO_PROXY_CONTRACT_KIND_HASH = keccak256("no-proxy");
+    bytes32 internal constant NO_PROXY_CONTRACT_KIND_HASH = keccak256("no-proxy");
 
     /**
      * @notice Address of the ChugSplashRegistry.
      */
     ChugSplashRegistry public immutable registry;
 
+    ICreate2 public immutable create2;
+
```
RPate97, line 54: Nit: It would be nice to have some inline documentation here just noting why this logic is in a different contract.

sam-goldman: I'll add that when I go through the contracts and add docs to everything

PR #638, opened 2023-04-26
maint(core): remove unnecessary encoding inplace bytes error

sam-goldman:
It appears that this error is unnecessary since we validate the inplace bytes type in `parseBytes`. Also, I believe we don't allow byte arrays, so the error is checking for the wrong thing anyways.

Note that this is the last error in the encoding logic that isn't a "this should never happen in any case" error.


RPate97 (approved):
Good catch

PR #640, opened 2023-04-27
fix(core): Improve gas estimation logic

RPate97:
## Purpose
Improves our off-chain gas cost estimation logic

- Only includes the cost of deploying a proxy if a proxy has not yet been deployed
- Does not include the cost of deploying any implementation contracts that have already been deployed
- Adds a fixed buffer of 150k to account for the initiate and complete function calls

## Testing
Tested using the remote executor with various cases including Rabbitholes contracts and a bunch of our test configs from the plugins package.

### DIFF
--- a/packages/core/src/fund.ts
+++ b/packages/core/src/fund.ts
@@ -1,9 +1,9 @@
 import { OWNER_BOND_AMOUNT } from '@chugsplash/contracts'
-import { ethers } from 'ethers'
+import { ethers, utils } from 'ethers'
 
 import {
   getChugSplashManager,
-  getDefaultProxyAddress,
+  getChugSplashManagerAddress,
   isContractDeployed,
 } from './utils'
 import {
@@ -14,6 +14,7 @@ import {
   isSetStorageAction,
 } from './actions'
 import { EXECUTION_BUFFER_MULTIPLIER } from './constants'
+import { ParsedChugSplashConfig } from './config'
 
 /**
  * Gets the amount ETH in the ChugSplashManager that can be used to execute a deployment. This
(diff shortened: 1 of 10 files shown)


packages/core/src/fund.ts
```diff
@@ -112,24 +115,20 @@ export const estimateExecutionGas = async (
 
   estimatedGas = estimatedGas.add(estimatedContractDeploymentGas)
 
-  return estimatedGas
+  return estimatedGas.add(BigNumber.from(150_000))
```
sam-goldman, line 118: Is there a reason you changed this from 400k? 150k seems sort of low since these functions have a for-loop and include the `_payExecutorAndProtocol` functions. Haven't tested it myself though

sam-goldman (commented):
Looks great! Just one minor comment

packages/core/src/fund.ts
```diff
@@ -74,31 +73,41 @@ export const estimateExecutionGas = async (
...
+        implementationAddress,
+        provider
+      ))
+        ? ethers.BigNumber.from(0)
+        : await provider.estimateGas({
+            data: action.code,
+          })
+      // We also tack on an extra 100k gas for each proxy target (including any that are not being upgraded) to account
+      // for the variable cost of the `initiateBundleExecution` and `completeBundleExecution` functions.
+      return deployCost.add(ethers.BigNumber.from(100_000))
```
RPate97, line 109: Accounts for the variable cost of the initiate and complete functions.

packages/core/src/fund.ts
RPate97: Ya know, I actually decided to just get rid of this entirely [see here](https://github.com/chugsplash/chugsplash/pull/640/files#r1179801665).

packages/core/src/fund.ts
sam-goldman: I think this is a good idea, but I think we should put it somewhere outside of this `map` statement. This is because the `actionBundle`, which is being iterated over here, can have a different number of proxies than the `targetBundle`. (See [here](https://github.com/chugsplash/chugsplash/pull/617) for an explanation).

I'd probably just do something like this:
```
const initiateAndCompleteCost = ethers.BigNumber.from(100_000).mul(bundles.targetBundle.targets.length)
```

sam-goldman (approved):

PR #643, opened 2023-04-28
feat(core): Standalone deployment script and configurable system owner

RPate97:
## Purpose
- Separates the deployment script from the executor (unless the executor is running against a local network)
- Adds a new standalone deployment script that calls initializeChugSplash with some set of executors to configure
- Allow us to configure a different owner from our multi-sig address using the `OWNER_PRIVATE_KEY` environment variable. This is designed to be used for deploying fresh systems owned by EOAs to live networks. Being able to do this makes live network testing a lot easier. It also allows us to run our own completely separate environments on test networks which we'll probably want to do once we have more users. 
- Adds logic to the deployment process that detects various different situations and uses the correct signer to perform the system setup transactions. More extensive details on this can be found in the `initializeChugSplash` function.


packages/core/scripts/deploy.ts
```diff
@@ -0,0 +1,32 @@
+import { Logger } from '@eth-optimism/common-ts'
+import hre from 'hardhat'
```
RPate97, line 2: I thought about moving all of the deployment logic into the contracts package b/c that makes more sense to me but didn't seem like a big enough deal to do in this PR. LMK if you think we should do this and I can make a ticket for it.

packages/core/src/addresses.ts
```diff
@@ -0,0 +1,123 @@
...
+      ['bytes', 'bytes'],
+      [
+        ChugSplashRegistryArtifact.bytecode,
+        utils.defaultAbiCoder.encode(
+          registryConstructorArgTypes,
+          getRegistryConstructorValues()
+        ),
+      ]
+    )
+  )
```
RPate97, line 61: Refactored the address constants to be function calls since that is required to allow for a configurable owner address. Since these addresses are also no longer constants, I've moved them into their own file as well.

packages/core/src/languages/solidity/predeploys.ts
```diff
@@ -292,175 +297,187 @@ export const initializeChugSplash = async (
...
+      ChugSplashManager.address
+    )) === false
+  ) {
+    await (
+      await ChugSplashRegistry.connect(signer).addVersion(
+        ChugSplashManager.address,
+        await getGasPriceOverrides(provider)
+      )
+    ).wait()
+  }
```
RPate97, line 351: The diff from here to the end of the predeploys.ts file is pretty jank. All I did was remove the isLiveNetwork check since we're now handling the signer separately.

packages/core/package.json
```diff
@@ -34,6 +34,7 @@
     "@eth-optimism/contracts-bedrock": "^0.8.0",
     "@eth-optimism/core-utils": "^0.9.1",
     "@ethersproject/abstract-provider": "^5.7.0",
+    "@nomiclabs/hardhat-ethers": "^2.2.1",
```
sam-goldman, line 37: I think this should be a devDependency like hardhat

packages/core/scripts/deploy.ts
sam-goldman: No preference; doesn't seem high priority to me

packages/core/src/languages/solidity/predeploys.ts
```diff
@@ -292,175 +297,187 @@ export const initializeChugSplash = async (
...
-      '[ChugSplash]: adding the initial ChugSplashManager version...'
-    )
-    const multisig = await getImpersonatedSigner(
-      OWNER_MULTISIG_ADDRESS,
-      provider
+  // We need to do some additional setup: adding the manager version, adding executor roles, etc
+  // This requires a signer with the owner role which we have to handle differently depending on the situation.
+  // 1. If the owner is the multisig and we're deploying on a test node then we can use an impersonated signer.
+  // 2. If the owner is the multisig and we're deploying on a live network then we have to use the gnosis safe ethers adapter (which we have not implemented yet).
+  // 3. We also allow the user to specify a different owner via process.env.OWNER_PRIVATE_KEY. This is useful for testing on live networks without using the multisig.
```
sam-goldman, line 304: If a user specifies an `OWNER_PRIVATE_KEY` in their .env file while using the ChugSplash packages, will this logic deploy the entire ChugSplash system? I'm picturing a scenario when a user unknowingly does this, since I'd imagine their private key could be stored in a .env variable named `OWNER_PRIVATE_KEY`.

If this is the case, maybe we could prefix this env var with `CHUGSPLASH__`

packages/core/src/languages/solidity/predeploys.ts
```diff
@@ -292,175 +297,187 @@ export const initializeChugSplash = async (
...
+  //    In this case, we need to create a signer using the OWNER_PRIVATE_KEY and use that.
+  let signer: ethers.Signer
+
+  // If deploying on a live network and the target owner is the multisig, then throw an error because
+  // we have not setup the safe ethers adapter yet.
+  if (
+    (await isLiveNetwork(provider)) &&
+    getOwnerAddress() === OWNER_MULTISIG_ADDRESS
+  ) {
+    // TODO: use the safe ethers adapter to send multisig transactions
```
sam-goldman, line 314: We got rid of TODOs in our code in favor of linear tix

packages/core/src/languages/solidity/predeploys.ts
```diff
@@ -292,175 +297,187 @@ export const initializeChugSplash = async (
...
+    throw new Error(
+      'Cannot run multisig transactions on a live network, please setup the safe ethers adapter first https://www.npmjs.com/package/@safe-global/safe-ethers-adapters'
     )
+  } else {
+    // if target owner is multisig, then use an impersonated multisig signer
+    if (getOwnerAddress() === OWNER_MULTISIG_ADDRESS) {
+      console.log('signer is owner multisig')
```
sam-goldman, line 321: Is this supposed to be a `logger?.info` instead of console.log?

packages/core/src/languages/solidity/predeploys.ts
```diff
@@ -292,175 +297,187 @@ export const initializeChugSplash = async (
...
 
-    // Fund the multisig.
+    // Fund the signer
     await (
       await deployer.sendTransaction({
-        to: OWNER_MULTISIG_ADDRESS,
+        to: await signer.getAddress(),
         value: ethers.utils.parseEther('0.1'),
       })
     ).wait()
```
sam-goldman, line 335: Do we need to do this for case 3? Or is it just necessary when we impersonate the multisig on a local network?

packages/core/src/languages/solidity/predeploys.ts
```diff
@@ -292,175 +297,187 @@ export const initializeChugSplash = async (
...
       )) === false
     ) {
-      await (
-        await ChugSplashRegistry.connect(multisig).addVersion(
-          ChugSplashManager.address,
-          await getGasPriceOverrides(provider)
-        )
-      ).wait()
+      await ManagedService.connect(signer).grantRole(
+        utils.defaultAbiCoder.encode(['bytes32'], [EXECUTOR_ROLE]),
```
sam-goldman, line 364: We can change `utils.defaultAbiCoder.encode(['bytes32'], [EXECUTOR_ROLE])` to just `EXECUTOR_ROLE`

packages/core/src/languages/solidity/predeploys.ts
```diff
@@ -292,175 +297,187 @@ export const initializeChugSplash = async (
...
+
+  logger?.info('[ChugSplash]: added the initial ChugSplashManager version')
 
+  logger?.info('[ChugSplash]: assigning executor roles...')
+  for (const executor of executors) {
     if (
-      (await ChugSplashRegistry.managerImplementations(
-        ChugSplashManager.address
+      (await ManagedService.hasRole(
+        utils.defaultAbiCoder.encode(['bytes32'], [EXECUTOR_ROLE]),
```
sam-goldman, line 359: We can change `utils.defaultAbiCoder.encode(['bytes32'], [EXECUTOR_ROLE])` to just `EXECUTOR_ROLE`

packages/core/src/languages/solidity/predeploys.ts
RPate97: Good call

sam-goldman: Thx. Slight preference for `CHUGSPLASH_INTERNAL__` just to fully remove this possibility

packages/core/src/languages/solidity/predeploys.ts
RPate97: It's just a convenience to handle this for case 3 as well b/c you can still use `CHUGSPLASH_INTERNAL__OWNER_PRIVATE_KEY` on a local node, and we'll occasionally want to do that when testing this logic locally. Automatically funding the owner's wallet just makes things a bit easier in that case. 

However, I do think I'll wrap this in a !isLiveNetwork check b/c it's weird to do this on a live network.

sam-goldman (approved):

PR #653, opened 2023-05-01
maint: rename bundle to deployment in certain contexts

sam-goldman:
Throughout our codebase, the term "bundle" currently describes two things:
1. A static data type that contains the Merkle root, its leafs, and proofs. I.e. the action bundle and the target bundle.
2. A deployment/upgrade. I.e. the bundle state, bundle status, bundle ID.

Using "bundle" as a catch-all term was more acceptable when a deployment/upgrade was basically just a single bundle. (See [here](https://github.com/chugsplash/chugsplash/blob/3a7b19cd4b6ec5556dc48d3a9f7b25de983ea713/packages/contracts/contracts/ChugSplashManager.sol#L156) for a very early version of the propose function). However, now that we use two bundles, it makes more sense to call it a "bundles ID", "bundles state", etc, but that's grammatically awkward.

In this PR, I changed all instances of case 2 to "deployment" instead of "bundle". In other words, it's now a deployment ID, deployment state, and deployment status. Case 1 remains the same, so action bundles and target bundles keep the same names.

This solution isn't perfect, since now "deployments" refer to upgrades too, which are distinct in smart contract land. We could call it an "update" instead, but I'm personally fine with this solution

### DIFF
--- a/docs/how-chugsplash-works.md
+++ b/docs/how-chugsplash-works.md
@@ -120,7 +120,7 @@ In addition to committing this info to IPFS, the user also submits a single `pro
 2. The root of the Merkle tree. Each Merkle proof supplied by the executor must yield this Merkle root.
 3. The number of leafs in the Merkle tree (i.e. the number of `SetStorage` and `DeployImplementation` actions). It's necessary to explicitly specify this to ensure that each action is only executed exactly once.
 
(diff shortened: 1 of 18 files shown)


sam-goldman:
@RPate97 You don't need to review every single name change. Just want to get your general thoughts on this.

RPate97 (commented):
Yeah, I think this is a good change. I think using just deployment to cover both deployments and upgrades is fine.

PR #660, opened 2023-05-04
feat(core): Support flexible constructors and wider variety of constructor arguments

RPate97:
## Purpose
- Adds a new option `unsafeAllowFlexibleConstructor` which disables our deterministic constructor checks
- Adds support for a wider variety of constructor arguments

## Note
- The `unsafeAllowFlexibleConstructor` option is only allowed on non-proxied contracts. We can consider expanding it to proxied contracts later, but we'll need to address some issues related to that. I.e we currently require all variables to be defined in the config, what if some are set via other methods in the constructor? 
- Does not include comprehensive tests of the validation logic for the newly supported constructor input types
- Does not include support for structs (though this PR does include some logic related to supporting structs, I chose to keep it disabled since I encountered some issues with it which I'd like to address separately)


packages/core/src/config/parse.ts
```diff
@@ -1185,6 +1222,127 @@ const parseContractVariables = (
...
+    )
+  }
+
+  return parsedValues
+}
+
+export const parseStructConstructorArg = (
+  paramType: ParamType,
+  constructorArgValue: UserConfigVariable
+) => {
```
RPate97, line 1256: I kept this because I know it works and I'll need it when I add support for structs in a followup PR.

packages/core/src/config/parse.ts
```diff
@@ -1185,6 +1222,127 @@ const parseContractVariables = (
...
+    parsedValues[key] = parseAndValidateConstructorArg(inputChild, value)
+  }
+
+  return parsedValues
+}
+
+const parseAndValidateConstructorArg = (
+  input: ParamType,
+  constructorArgValue: UserConfigVariable
+): ParsedConfigVariable => {
```
RPate97, line 1284: I split the constructor arg validation out into it's own function which follows a similar recursive pattern to the parseAndValidateVariable function. I don't think it's worth trying to reuse the `recursiveLayoutIterator` for this though because the context is meaningfully different. So I think it would just add a lot of unnecessary bloat to it.

sam-goldman (approved):

PR #661, opened 2023-05-05
fix(core): add support for structs in constructor args

sam-goldman:


packages/core/src/utils.ts
```diff
@@ -831,43 +831,35 @@ export const getContractAddress = (
...
-  )
+  const constructorArgValues = getConstructorArgs(constructorArgs, abi)
+
+  const iface = new ethers.utils.Interface(abi)
 
   const creationCodeWithConstructorArgs = bytecode.concat(
-    remove0x(
-      utils.defaultAbiCoder.encode(constructorArgTypes, constructorArgValues)
-    )
+    remove0x(iface.encodeDeploy(constructorArgValues))
```
sam-goldman, line 862: I replaced `defaultAbiCoder.encode` with `iface.encodeDeploy` because `defaultAbiCoder.encode` would silently fail if the `constructorArgTypes` weren't defined correctly

packages/core/src/config/parse.ts
```diff
@@ -1958,7 +1958,10 @@ export const assertValidContracts = (
         }
       }
 
-      if (!contractConfig.unsafeAllowEmptyPush) {
+      if (
+        !contractConfig.unsafeAllowEmptyPush &&
+        contractConfig.kind !== 'no-proxy'
```
RPate97, line 1963: What's the purpose of this change?

sam-goldman: It's not related to the main purpose of this PR. It just skips the `unsafeAllowEmptyPush` check for non-proxied contracts. This is because `push()` can only be dangerous in an upgradeable contract.

packages/core/src/config/parse.ts
```diff
@@ -1262,33 +1303,82 @@ export const parseStructConstructorArg = (
...
+      'warning',
+      `Detected array(s) in struct constructor argument. Struct members that are arrays will be ignored by the Solidity compiler and cannot be set in the constructor.`,
+      [
+        'The following struct members will be ignored: ',
+        ...arrayMembers.map((el) => `${el.name}: ${el.paramType.type}`),
+      ],
+      cre.silent,
+      cre.stream
+    )
+  }
```
sam-goldman, line 1328: I actually don't think this is necessary. The Solidity compiler doesn't ignore these values. I ran into this same issue a while ago, and it turns out that you just need to explicitly define a getter function for them in Solidity. See [here](https://github.com/chugsplash/chugsplash/blob/develop/packages/contracts/contracts/ChugSplashManager.sol#L740-L754) for an explanation (esp the linked github issue)

RPate97: Wow, that's pretty wonky

sam-goldman: Lol yep, welcome to Solidity

PR #670, opened 2023-05-10
feat(core): Support gasless proposals with meta transactions

RPate97:
## Purpose
Adds support for gasless proposals using meta transactions. Note that this includes an automated test that requires the use of IPFS credentials. 

## Testing
You must add the following environment variable and value to your plugin's `.env` file for the new test to work. Note this is just b/c we check for the API key to determine if a gasless proposal should be done or not. The value of the key doesn't matter right now. 
`CHUGSPLASH_API_KEY=testing`

### DIFF
--- a/packages/contracts/contracts/ChugSplashManager.sol
+++ b/packages/contracts/contracts/ChugSplashManager.sol
@@ -25,6 +25,12 @@ import { IAccessControl } from "@openzeppelin/contracts/access/IAccessControl.so
 import { ICreate2 } from "./interfaces/ICreate2.sol";
 import { Semver, Version } from "./Semver.sol";
 import { IGasPriceCalculator } from "./interfaces/IGasPriceCalculator.sol";
+import {
+    ERC2771ContextUpgradeable
+} from "@openzeppelin/contracts-upgradeable/metatx/ERC2771ContextUpgradeable.sol";
+import {
+    ContextUpgradeable
+} from "@openzeppelin/contracts-upgradeable/utils/ContextUpgradeable.sol";
 
 /**
  * @title ChugSplashManager
@@ -46,18 +52,13 @@ contract ChugSplashManager is
     OwnableUpgradeable,
     ReentrancyGuardUpgradeable,
     Semver,
-    IChugSplashManager
+    IChugSplashManager,
+    ERC2771ContextUpgradeable
 {
     /**
      * @notice Role required to be a remote executor for a deployment.
      */
     bytes32 internal constant REMOTE_EXECUTOR_ROLE = keccak256("REMOTE_EXECUTOR_ROLE");
-
-    /**
-     * @notice Role required to collect the protocol creator's payment.
-     */
-    bytes32 internal constant PROTOCOL_PAYMENT_RECIPIENT_ROLE =
-        keccak256("PROTOCOL_PAYMENT_RECIPIENT_ROLE");
 
     /**
      * @notice Role required to propose deployments through the ManagedService contract.
@@ -270,14 +271,6 @@ contract ChugSplashManager is
     event ExecutorPaymentClaimed(address indexed executor, uint256 withdrawn, uint256 remaining);
 
     /**
-     * @notice Emitted when a protocol payment recipient claims a payment.
-     *
-     * @param recipient The recipient that withdrew the funds.
-     * @param amount    Amount of ETH withdrawn.
-     */
-    event ProtocolPaymentClaimed(address indexed recipient, uint256 amount);
-
-    /**
      * @notice Emitted when the owner withdraws ETH from this contract.
      *
      * @param owner  Address of the owner.
@@ -439,15 +432,16 @@ contract ChugSplashManager is
     error InsufficientExecutorDebt();
 
     /**
+     * @notice Reverts if there's not enough funds in the contract pay the protocol fee and the
+     *  withdraw amount requested by the executor.
+     */
+    error InsufficientFunds();
+
+    /**
      * @notice Reverts if a withdrawal transaction fails. This is likely due to insufficient funds
        in this contract.
      */
     error WithdrawalFailed();
-
-    /**
-     * @notice Reverts if the caller is not a protocol payment recipient.
-     */
-    error CallerIsNotProtocolPaymentRecipient();
 
     /**
      * @notice Reverts if there is no bytecode at a given address.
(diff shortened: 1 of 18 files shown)


packages/contracts/contracts/ChugSplashManager.sol
```diff
@@ -839,44 +868,29 @@ contract ChugSplashManager is
...
 
-        registry.announce("ExecutorPaymentClaimed");
-    }
-
-    /**
-     * @notice Allows the protocol creators to claim their royalty, which is only earned during
-       remotely executed deployments.
-     */
-    function claimProtocolPayment() external {
-        if (!managedService.hasRole(PROTOCOL_PAYMENT_RECIPIENT_ROLE, msg.sender)) {
```
RPate97, line 864: I updated the executor payment claiming logic to also send funds to the managed service contract and moved the withdraw logic over to that contract.

packages/contracts/contracts/Import.sol
```diff
@@ -0,0 +1,4 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.15;
+
+import { Forwarder } from "@thirdweb-dev/contracts/forwarder/Forwarder.sol";
```
RPate97, line 4: This just ensures the artifact is generated so it can be used during the deployment process.

packages/plugins/test/Metatx.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+import { FORWARDER_ADDRESS, ForwarderArtifact } from '@chugsplash/contracts'
+import { expect } from 'chai'
+
+import { getArtifactPaths } from '../../plugins/dist'
+import { createChugSplashRuntime } from '../../plugins/src/utils'
+
+const configPath = './chugsplash/Metatx.config.ts'
+
+describe('Meta txs', () => {
+  process.env['LOCAL_TEST_METATX_PROPOSE'] = 'true'
```
RPate97, line 21: We only perform meta transactions on live networks when the API key is available. I've included this environment variable to override the network check for testing.

packages/contracts/contracts/ManagedService.sol
sam-goldman, line 1: I think every instance of `_msgSender()` in this contract can be changed to `msg.sender`

packages/contracts/contracts/ChugSplashManager.sol
sam-goldman, line 1: Are we supposed to use `_msgSender()` instead of `msg.sender` everywhere in this contract? Or only in the places where we're using meta txns?

packages/core/src/tasks/index.ts
```diff
@@ -1065,31 +1078,60 @@ export const proposeChugSplashDeployment = async (
...
+  if (
+    process.env.CHUGSPLASH_API_KEY &&
+    ((await isLiveNetwork(provider)) ||
+      process.env.LOCAL_TEST_METATX_PROPOSE === 'true')
+  ) {
+    if (!process.env.PRIVATE_KEY) {
+      throw new Error(
+        'Must provide a PRIVATE_KEY environment variable to sign gasless proposal transactions'
+      )
+    }
```
sam-goldman, line 1095: nit: I don't think this if-statement can ever be triggered

sam-goldman (commented):
Looks great

packages/contracts/contracts/ChugSplashManager.sol
RPate97: Technically because this contract has a trusted forwarder, it's actually possible to use meta transactions on any function in the contract. So it's best practice to use `_msgSender()` everywhere as it handles both cases. You can checkout the source code for `ERC2771ContextUpgradeable` if you're interested in how that works.

packages/core/src/tasks/index.ts
RPate97: To my knowledge, neither Hardhat nor Foundry require the use of the PRIVATE_KEY environment variable. If the user chose to use any other environment variable name for their private key (or used something else like a mnemonic phrase in Foundry), then our meta transaction logic would break as it relies on reading a private key from specifically this environment variable.

packages/contracts/contracts/ChugSplashManager.sol
sam-goldman: I checked out `ERC2771ContextUpgradeable`, and I think I have a grasp of how it works. It's fine to leave the contract as-is for now, but when we go to prod, I'd imagine we'd want to change `_msgSender()` to `msg.sender` everywhere we don't intend to use meta transactions. It seems like it'd add unnecessary surface area of attack if we keep it everywhere

packages/core/src/tasks/index.ts
sam-goldman: Sorry I misread the original code. Disregard this

sam-goldman (approved):

packages/contracts/contracts/ChugSplashManager.sol
sam-goldman: Just curious, what's the reason it's a best practice to use `_msgSender()` everywhere, even where we aren't using meta transactions? I might be missing something

RPate97: Because meta-transactions are implemented at the contract level, not the function level. Since meta transactions are implemented on this contract, there's nothing stopping us (or anyone else) from sending transactions using the forwarder.  Any functions where we don't use `_msgSender()` would simply not function as expected if meta transactions are used with them. While it may not be strictly necessary to use `_msgSender()` for all other functions, there are essentially zero downsides to it. 

If anything, not using `_msgSender()` in all cases adds complexity because now we have to think about which functions are intended to be used with meta transactions and which are not. Also, anyone integrating with the contract (including us, or future developers working with us) needs to be aware that while meta transactions are available for the entire contract, they'll only actually work for specific functions.

sam-goldman: IMO we should use meta transactions on the function level instead of the contract level. We can make this nice and explicit by using `ERC2771ContextUpgradeable._msgSender()` for meta txns. This means `_msgSender()` will have standard behavior everywhere else.

Let me elaborate on why I think this is important. Using meta txns for the entire contract leads to subtle differences in key functionality that should remain untouched. For example, it's non-obvious that meta transactions are allowed in the `onlyOwner` modifier in the `approve` function. (The `_msgSender()` call is nested two call frames inside the modifier). Not only is this surprising, it's entirely unnecessary, and obfuscates the actual meaning of the `onlyOwner` statement. If a reader clicks into the `onlyOwner` modifier, they need to know what's going on [here](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v4.8.2/contracts/metatx/ERC2771Context.sol#L25C12-L30). What's a trusted forwarder? What's going on with the assembly? I figured out these questions, but readers shouldn't need to bother.

If this sounds reasonable, we also just need to change [this line](https://github.com/chugsplash/chugsplash/blob/4b0d5f24bf681a74c1ea547f173956bdd01cb91b/packages/contracts/contracts/ChugSplashManager.sol#L1536) to be:

```
sender = ContextUpgradeable._msgSender();
```

PR #674, opened 2023-05-16
fix(core): fix behavior of create3 deployments

sam-goldman:
When deploying a proxy's implementation via Create3, we now use a salt equal to `hash(creationCodeWithConstructorArgs)`. This essentially turns Create3 into Create2 for proxy implementations.

When deploying a non-proxy contract, we log a warning instead of an error when the contract at the Create3 address has a `hash(creationCodeWithConstructorArgs)` equal to the non-proxy contract in the config.

### DIFF
--- a/packages/core/src/config/parse.ts
+++ b/packages/core/src/config/parse.ts
@@ -2648,7 +2648,7 @@ const assertAvailableCreate3Addresses = async (
       if (match) {
         logValidationError(
           'warning',
-          `Skipping deployment of ${referenceName}. Add a new 'salt' value to re-deploy it at a new address.`,
+          `Skipping deployment of ${referenceName} since it has already been deployed and has not changed. Add a new 'salt' value to re-deploy it at a new address.`,
           [],
           cre.silent,
           cre.stream
--- a/packages/core/src/config/parse.ts
+++ b/packages/core/src/config/parse.ts
@@ -49,6 +49,8 @@ import {
   isContractDeployed,
   assertValidBlockGasLimit,
   getCreationCodeWithConstructorArgs,
+  getDeployedCreationCodeWithArgsHash,
+  getNonProxyCreate3Salt,
 } from '../utils'
 import {
   UserChugSplashConfig,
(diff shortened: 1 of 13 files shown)


packages/core/src/tasks/index.ts
```diff
@@ -738,6 +747,12 @@ export const chugsplashDeployAbstractTask = async (
 
   // At this point, the deployment has been completed.
   if (integration === 'hardhat') {
+    if (!(await isLiveNetwork(provider))) {
+      // We save the snapshot ID here so that tests on the stand-alone Hardhat network can be run
+      // against the most recently deployed contracts.
+      await writeSnapshotId(provider, networkName, deploymentFolder)
+    }
```
sam-goldman, line 754: Needed to add this so that tests on the stand-alone Hardhat network work properly after upgrading a proxy via the chugsplash-deploy task.

packages/plugins/test/Validation.spec.ts
```diff
@@ -63,23 +63,31 @@ describe('Validate', () => {
...
+    try {
+      await readValidatedChugSplashConfig(
+        provider,
+        constructorArgConfigPath,
+        constructorArgsValidationArtifactPaths,
+        'hardhat',
+        cre,
+        false
+      )
+    } catch (e) {
```
sam-goldman, line 88: Needed to add these try/catch statements because I added a call to `getCreationCodeWithConstructorArgs` in `constructParsedConfig`, which throws an error when there's incorrectly encoded constructor args.

packages/core/src/config/parse.ts
```diff
@@ -2590,7 +2618,44 @@ const assertAvailableCreate3Addresses = async (
...
+          contractConfig.constructorArgs,
+          abi
+        )
+      )
+
+      const match = BigNumber.from(deployedHash).eq(BigNumber.from(currHash))
+      if (match) {
+        logValidationError(
+          'warning',
+          `Skipping deployment of ${referenceName}. Add a new 'salt' value to re-deploy it at a new address.`,
```
RPate97, line 2651: Nit: Would be nice to include a little bit more context on exactly why it was skipped. 

```suggestion
          `Skipping deployment of ${referenceName} since it has already been deployed and has not changed. Add a new 'salt' value to re-deploy it at a new address.`,
```

RPate97 (approved):
Looks good, just left one minor nitpick

PR #696, opened 2023-05-19
maint(core): remove `ArtifactPaths` in favor of `ConfigArtifacts`

sam-goldman:
This removes a lot of unnecessary file I/O, which should speed things up a bit. Using a single persistent object for the artifacts also makes it easy to modify the build info objects, which is necessary while [this Foundry bug](https://github.com/foundry-rs/foundry/issues/4981) is pending.


RPate97 (approved):
Looks good

PR #718, opened 2023-05-24
feat(ct): Deploy in Foundry with bootloader

RPate97:
## Purpose
- Updates the deployment process to use two new contracts the `AdapterDeployer` and the `ChugSplashBootloader`.
- The `AdapterDeployer` handles deploying all of our various adapters, updaters, and the v1 manager
- The `ChugSplashBootloader` deploys the rest of the system

## Notes
I wasn't able to get the foundry plugin working after making this change even though I am 100% sure that the contracts are being deployed properly to the correct addresses in the foundry `_ensureChugSplashInitialized` function. I believe this is due to how foundry handles sending transactions where they are just simulated until the user chooses to actually broadcast them. This is probably solvable now, but I don't see a great reason to bother with it since I think the issue will resolve itself once we port the rest of the tasks to send all transactions from Solidity. Let me know if you disagree with this decision. I have run the hardhat plugin tests for this and they are all passing.


sam-goldman:
> I don't see a great reason to bother with it since I think the issue will resolve itself once we port the rest of the tasks to send all transactions from Solidity. Let me know if you disagree with this decision

Yeah that's totally fine

packages/plugins/foundry-contracts/ChugSplash.sol
```diff
@@ -58,17 +73,76 @@ contract ChugSplash is Script, Test {
...
 
-        vm.ffi(cmds);
+    function _ensureChugSplashInitialized() private {
+        // Fetch bytecode from artifacts
+        DeploymentBytecode memory bootloaderBytecode = getBootloaderBytecode();
+
+        // If the registry is not already deployed
+        if (bootloaderBytecode.registryAddress.code.length == 0) {
+            // If the target chain is a local network
+            if (block.chainid == 31337) {
```
sam-goldman, line 94: Hard-coding `31337` won't work if the user spins up a local node with a different chain ID. I looked through the Foundry docs pretty thoroughly, and couldn't think of a way to determine if the current network is local or not. I think the best option is FFI.

packages/plugins/foundry-contracts/ChugSplash.sol
```diff
@@ -58,17 +73,76 @@ contract ChugSplash is Script, Test {
...
+                // Setup determinisitic deployment proxy
+                address DeterministicDeploymentProxy = 0x4e59b44847b379578588920cA78FbF26c0B4956C;
+                vm.etch(DeterministicDeploymentProxy, hex"7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe03601600081602082378035828234f58015156039578182fd5b8082525050506014600cf3");
+
+                // Deploy the adapters
+                bytes memory adapterCreationCode = bootloaderBytecode.adapterDeployer;
+                address adapterDeployerAddress = Create2.computeAddress(bytes32(0), keccak256(adapterCreationCode), DeterministicDeploymentProxy);
+                (bool successAdapterDeployer, ) = DeterministicDeploymentProxy.call(bytes.concat(bytes32(0), adapterCreationCode));
+                require(successAdapterDeployer, string.concat("Adapter deployer deployment failed"));
```
sam-goldman, line 103: nit: `string.concat` isn't necessary here

packages/plugins/foundry-contracts/ChugSplash.sol
```diff
@@ -58,17 +73,76 @@ contract ChugSplash is Script, Test {
...
+
+                // Deploy the bootloader
+                bytes memory bootloaderCreationCode = abi.encodePacked(
+                    bootloaderBytecode.bootloader,
+                    abi.encode(systemOwnerAddress)
+                );
+                address bootloaderAddress = Create2.computeAddress(bytes32(0), keccak256(bootloaderCreationCode), DeterministicDeploymentProxy);
+                (bool successBootloader, ) = DeterministicDeploymentProxy.call(bytes.concat(bytes32(0), bootloaderCreationCode));
+                require(successBootloader, string.concat("Bootloader deployment failed"));
```
sam-goldman, line 112: > nit: `string.concat` isn't necessary here

packages/core/src/languages/solidity/compiler.ts
```diff
@@ -104,3 +108,37 @@ export const getMinimumCompilerInput = (
...
+
+/**
+ * Returns the minimum compiler output for a given source name.
+ *
+ * @param fullCompilerOutput The full compiler output object.
+ * @param fullOutputContracts The full compiler output contract object.
+ * @param sourceName The source name.
+ * @returns Minimum compiler input necessary to compile the source name.
+ */
+export const getMinimumCompilerOutput = (
```
sam-goldman, line 120: What's the purpose of this function?

packages/core/src/languages/solidity/compiler.ts
```diff
@@ -104,3 +108,37 @@ export const getMinimumCompilerInput = (
...
+  const minimumSources: CompilerOutput['sources'] = {}
+  const minimumContracts: CompilerOutput['contracts'] = {}
+  for (const newSourceName of Object.keys(metadata.sources)) {
+    minimumSources[newSourceName] = fullCompilerOutput.sources[newSourceName]
+    minimumContracts[newSourceName] =
+      fullCompilerOutput.contracts[newSourceName]
+  }
+
+  fullCompilerOutput.contracts = minimumContracts
+  fullCompilerOutput.sources = minimumSources
```
sam-goldman, line 141: I don't think we should overwrite these members of `fullCompilerOutput`, since we may use them elsewhere. Instead, I think we should make a new object like in `getMInimumCompilerInput`

sam-goldman:
nit: Most of the variable names in the bootloader contracts are in PascalCase, which makes my Solidity vscode extension think they're contracts by shading them green. This doesn't impact the functionality, but slight preference for making these variables camelcase. Notice how `DefaultUpdaterAddr` and `DeterministicDeployer` are both green:

Screen Shot 2023-05-23 at 9 19 57 PM

packages/contracts/contracts/deployment/AdapterDeployer.sol
```diff
@@ -0,0 +1,68 @@
...
+import { Create2 } from "@openzeppelin/contracts/utils/Create2.sol";
+import { DeterministicDeployer } from "./DeterministicDeployer.sol";
+
+contract AdapterDeployer {
+    address public DefaultUpdaterAddr;
+    address public OZUUPSUpdaterAddr;
+    address public OZTransparentAdapterAddr;
+    address public OZUUPSOwnableAdapterAddr;
+    address public OZUUPSAccessControlAdapterAddr;
+    address public DefaultAdapterAddr;
```
sam-goldman, line 20: nit: The variables here and in the other bootloader should be camel case

packages/core/src/addresses.ts
```diff
@@ -117,17 +134,128 @@ export const getReferenceDefaultProxyAddress = () =>
...
+      ]
+    )
+  )
+
+export const getForwarderAddress = () =>
+  utils.getCreate2Address(
+    DETERMINISTIC_DEPLOYMENT_PROXY_ADDRESS,
+    constants.HashZero,
+    utils.solidityKeccak256(['bytes'], [ForwarderArtifact.bytecode])
+  )
```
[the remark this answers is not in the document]
RPate97: They don't need to be. While I was working on this, there was just an iteration where it was necessary to do this.

packages/core/src/languages/solidity/compiler.ts
RPate97: This may not be necessary anymore, so I'll test out removing it. 

It was needed to filter out some unnecessary artifacts that were included in the compiler output due to the user importing our foundry contract which imports our core contracts. This resulted in some dependency contracts being included which could generate errors due to the user not having the correct dependencies installed.

packages/plugins/foundry-contracts/ChugSplash.sol
```diff
@@ -58,17 +73,76 @@ contract ChugSplash is Script, Test {
...
         cmds[1] = "node";
         cmds[2] = filePath;
-        cmds[3] = "initializeChugSplash";
-        cmds[4] = rpcUrl;
-        cmds[5] = network;
-        cmds[6] = privateKey;
+        cmds[3] = "getBootloaderBytecode";
+
+        bytes memory result = vm.ffi(cmds);
+        return abi.decode(result, (DeploymentBytecode));
```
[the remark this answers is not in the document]
RPate97: The bootloaders are not deployed here. The purpose of this function is to fetch the exact bytecode of the bootloaders from the artifacts on the contracts package which are then deployed using create 2. `type(bootloader).creationCode` doesn't work because the bytecode generated is dependent on the users' compiler configuration which may be different from ours.

packages/plugins/foundry-contracts/ChugSplash.sol
RPate97: So just using ffi won't work for the internal foundry node, but we could probably do something where we first check if the chain id is `31337`. Then if there is an rpc url provided, we could use ffi to check if that node is anvil node using our `isLiveNetwork` function.

sam-goldman: That unfortunately won't work if the user changes the chain ID of the internal foundry node (i.e. `forge script --chain-id <number>`). Also, I'm not the biggest fan of our solution to get the current rpc url, since it requires the user to hard-code a `NETWORK` var in their `.env` file, which may be incorrect. I wasn't able to figure out a better solution yet, but it's probably possible to do better

packages/core/src/languages/solidity/compiler.ts
sam-goldman: Sounds good, lmk if you want to keep it in and i'll review this function

packages/plugins/foundry-contracts/ChugSplash.sol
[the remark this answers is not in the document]
RPate97: Yeah, that's fine, I'll update this to use a separate getter here instead of including the address on the same struct.

packages/core/src/languages/solidity/compiler.ts
RPate97: So I double-checked this and it is required whenever the foundry library contract is included in the compilation process because it causes all of our system contracts to be compiled. Then some of our system dependencies are filtered out by the `getMinimumCompilerInput` function which causes an error when we try to create the `UpgradableContract` with them. So this solution handles that by filtering everything from both the compiler input and output. 

An alternative solution would be to not use the `getMinimumCompilerInput` function at all, or update it to avoid filtering out any of our system contracts or their dependencies. But IMO, the current solution of using a mirror `getMinimumCompilerOutput` function seems better to me. So yeah, you should review it.

sam-goldman: LGTM

sam-goldman (approved):

PR #732, opened 2023-05-26
Port executeTask to Solidity

RPate97:


packages/plugins/foundry-contracts/ChugSplashLocalExecutor.sol
```diff
@@ -0,0 +1,203 @@
...
+            uint batchSize = findMaxBatchSize(filteredActions.slice(executed, filteredActions.length), manager, maxGasLimit);
+            BundledChugSplashActions[] memory batch = filteredActions.slice(executed, executed + batchSize);
+
+            (RawChugSplashAction[] actions, uint256[] memory _actionIndexes, bytes32[][] memory _proofs) = disassembleActions(batch);
+
+            // Execute the batch
+            // We specifically start and stop the broadcast here to make sure this is the only time transactions are actually sent
+            vm.startBroadcast();
+            manager.executeActions(actions, _actionIndexes, _proofs);
+            vm.stopBroadcast();
```
RPate97, line 123: Note the use of `vm.startBroadcast()` here, we need to do this or something like it b/c we don't want to broadcast transactions during the `findMaxBatchSize` function. I'm not sure if this is exactly how it'll need to work in practice since I haven't used Foundry much, but I know we'll need to do something like this.

packages/plugins/foundry-contracts/ChugSplashLocalExecutor.sol
```diff
@@ -0,0 +1,203 @@
...
+    function executable(
+        selected[] BundledChugSplashAction,
+        ChugSplashManager manager,
+        uint masGasLimit
+    ) public view returns (bool) {
+        uint256 initialGasLeft = gasleft();
+
+        (RawChugSplashAction[] actions, uint256[] memory _actionIndexes, bytes32[][] memory _proofs) = disassembleActions(selected);
+
+        manager.executeActions(actions, _actionIndexes, _proofs);
```
sam-goldman, line 37:
```suggestion
        (bool success, ) = manager.staticcall{ gas: masGasLimit }(abi.encodeCall(ChugSplashManager.executeActions, (actions, _actionIndexes, _proofs)));
        return success;
```

This mirrors the TypeScript logic. If it looks good to you, we can remove the "gasLeft" stuff in this function

packages/plugins/foundry-contracts/ChugSplashLocalExecutor.sol
sam-goldman: I don't think we can use `broadcast` in our solidity code because it may send transactions to the blockchain when the user doesn't intend to (e.g. on a forked network). The user must be the one to add `broadcast` to their code. This is fine because we can use `staticcall`, which doesn't broadcast anything, in `findMaxBatchSIze`.

packages/plugins/foundry-contracts/ChugSplashLocalExecutor.sol
```diff
@@ -0,0 +1,203 @@
...
+        return min;
+    }
+
+    /**
+     * Helper function for executing a list of actions in batches.
+     */
+    function executeBatchActions(
+        BundledChugSplashActions[] actions,
+        ChugSplashManager manager
+    ) public {
```
sam-goldman, line 92:
```suggestion
    ) public returns (DeploymentStatus memory) {
```

packages/plugins/foundry-contracts/ChugSplashLocalExecutor.sol
```diff
@@ -0,0 +1,203 @@
...
+        manager.initiateUpgrade(targets, proofs);
+        vm.stopBroadcast();
+
+        // Execute all the set storage actions
+        executeBatchActions(setStorageActions, manager);
+
+        // Complete the upgrade
+        vm.startBroadcast();
+        manager.finalizeUpgrade(targets, proofs);
+        vm.stopBroadcast();
```
sam-goldman, line 199: We can remove these broadcasts

packages/plugins/foundry-contracts/ChugSplashLocalExecutor.sol
```diff
@@ -0,0 +1,203 @@
...
+
+// SPDX-License-Identifier: MIT
+import { ChugSplashBundles, DeploymentState, BundledChugSplashAction, RawChugSplashAction } from "@chugsplash/contracts/contracts/ChugSplashDataTypes.sol";
+import { UD60x18, ud, convert } from "@prb/math/UD60x18.sol";
+
+abstract contract ChugSplashLocalExecutor {
+    /**
+     * @notice Splits up a bundled action into its components
+     */
+    function disassembleActions(actions[] BundledChugSplashAction) public returns (RawChugSplashAction[], uint256[], bytes32[][]) {
```
sam-goldman, line 11: Could you potentially change the `ChugSplashManager` to accept the `BundledChugSplashTarget`/`BundledChugSplashAction` structs in its execution functions? That way we don't need to disassemble the structs. This'll also make unit testing easier in the future

RPate97: We could do that, but then we'd need to do a breaking release or deal with releasing a new manager version. That doesn't seem worth it to me for this, if you want this maybe make an issue for it and we can do it whenever we want to push out a new set of contracts.

packages/plugins/foundry-contracts/ChugSplashLocalExecutor.sol
RPate97: Nice that's much better, didn't realize you could do this on chain.

packages/plugins/foundry-contracts/ChugSplashLocalExecutor.sol
sam-goldman: Sounds good, made a ticke

sam-goldman (approved):

PR #734, opened 2023-05-30
feat(pg): Implement config parsing ffi tasks

RPate97:


packages/plugins/foundry-contracts/ChugSplash.sol
```diff
@@ -340,24 +338,16 @@ contract ChugSplash is Script, Test, DefaultCreate3, ChugSplashManagerEvents, Ch
             );
         }
 
-        if (_route == ProposalRoute.RELAY || _route == ProposalRoute.REMOTE_EXECUTION) {
-            ffiCommitToIPFS(_deploymentId);
-        }
-
-        if (_route == ProposalRoute.RELAY) {
-            ffiRelayProposal(_deploymentId);
-        } else {
```
RPate97, line 349: Since the standalone propose task is only ever used against live networks and only sends transactions using our backend, I don't see a strong reason to maintain a Solidity implementation of this extra path. I believe the propose task can be largely shared between the hardhat and foundry implementations. 

I will verify this in a separate task once we're done with the deploy task. If it is necessary to port more logic for the propose task, then I will do it then.

packages/plugins/src/foundry/index.ts
```diff
@@ -36,6 +45,59 @@ import { createChugSplashRuntime } from '../utils'
...
+    liveNetwork: configCache.liveNetwork,
+    networkName: configCache.networkName,
+    contractConfigCache: {},
+  }
+
+  for (const cachedContract of configCache.contractConfigCache) {
+    structuredConfigCache.contractConfigCache[cachedContract.referenceName] = {
+      referenceName: cachedContract.referenceName,
+      isTargetDeployed: cachedContract.isTargetDeployed,
+      deploymentRevert: {
```
RPate97, line 72: This is probably the jankiest part of this PR. Here I am taking the ConfigCache object which has been encoded as a bytes string, and decoding it using the Struct ABI. I then do some modifications to it to get it into the expected format for our Typescript logic. This is required because the ConfigCache object type is slightly different from the typescript version.

So yeah, this could end up being annoying to maintain b/c if we ever need to change this object it'll require updating this logic to make sure it's properly translated. We could probably simplify this by removing the `Optional<Type>` pattern in favor of using expected defaults. I.e an undefined address is just `address(0)` and then our typescript logic can detect that and handle it. This is just an idea though, I don't have a particularly strong opinion if it's necessary or not.

packages/plugins/src/foundry/index.ts
```diff
@@ -369,20 +430,138 @@ const command = args[0]
...
+        cre,
+        configCache,
+        true
+      )
+      break
+    }
+    case 'getCurrentChugSplashManagerVersion': {
+      const artifactStructABI =
+        'tuple(uint256 major, uint256 minor, uint256 patch)'
+      const encodedVersion = ethers.utils.AbiCoder.prototype.encode(
```
sam-goldman, line 500: nit: I think the syntax that Ethers encourages is `ethers.utils.defaultAbiCoder.encode`

packages/plugins/src/foundry/index.ts
```diff
@@ -369,20 +430,138 @@ const command = args[0]
...
+      }
+      fs.writeFileSync(
+        './cache/chugsplash-config-cache.json',
+        JSON.stringify(configCache, null, 2),
+        'utf-8'
+      )
+
+      const ChugSplashFoundryABI =
+        // eslint-disable-next-line @typescript-eslint/no-var-requires
+        require('../../out/artifacts/ChugSplash.sol/ChugSplash.json').abi
```
sam-goldman, line 449: How about we use `forge inspect` here? I think it'd be a little less brittle in case we ever change the artifact folder.
```
const abiRaw = await execAsync('forge inspect ChugSplash abi')
const abi: Array<Fragment> = JSON.parse(abiRaw.stdout)
```
Same thing for the other two places we get the ChugSplashFoundryABI

packages/plugins/src/foundry/index.ts
```diff
@@ -369,20 +430,138 @@ const command = args[0]
...
+
+      const {
+        parsedConfig,
+        configArtifacts,
+      }: {
+        parsedConfig: ParsedChugSplashConfig
+        configArtifacts: ConfigArtifacts
+      } = JSON.parse(
+        (
+          await fs.readFileSync('./cache/chugsplash-config-cache.json')
```
sam-goldman, line 473: nit: unnecessary `await`

packages/plugins/src/foundry/index.ts
```diff
@@ -369,20 +430,138 @@ const command = args[0]
...
```
sam-goldman, line 520: nit: unnecessary `await`

packages/plugins/foundry-contracts/ChugSplash.sol
RPate97: Yeah, I believe that's what I did, let me know if you meant something else.

sam-goldman: Yeah my bad, didn't realize you already removed them

packages/plugins/src/foundry/index.ts
sam-goldman: Ah i see. I'll reconsider the struct design. LGTM though

packages/plugins/src/foundry/index.ts
RPate97: This seems extremely slow in some situations, I think because it's potentially triggering a rebuild. I.e if I run `forge build` and then run this command it takes 30 seconds before it outputs anything despite the fact that the artifact should be readily available. Doesn't seem worth using it to me.

sam-goldman: Ah i see. What about running `execAsync(forge config --json)`, like in `getPaths`, to get the "out" folder?

sam-goldman: In fact I think we can use the `artifactFolder`, which is already available

PR #740, opened 2023-06-06
Fix incorrect gas limit when broadcasting

RPate97:
## Purpose
- Fixes a few bugs when deploying the core contracts against Anvil
- Calls `ffiDeployOnAnvil` in the ChugSplash library constructor
- Specifies a custom gas limit for transactions sent from the ChugSplash library
- Removes the call to `ensureChugSplashInitialized` from the deploy function b/c it does not work when broadcasting yet. For the time being, we will have to call it from outside the broadcast. Since it requires the rpcurl, this means it must be called from the deployment script. See the `src/scripts/ChugSplash.s.sol` for an example. This will be resolved when the raw transaction cheatcode is released.

### DIFF
--- a/packages/core/src/utils.ts
+++ b/packages/core/src/utils.ts
@@ -26,7 +26,6 @@ import { add0x, remove0x } from '@eth-optimism/core-utils'
 import chalk from 'chalk'
 import {
   ProxyDeployment,
-  StorageLayout,
   UpgradeableContract,
   ValidationOptions,
   withValidationDefaults,
@@ -64,7 +63,6 @@ import {
 import { CURRENT_CHUGSPLASH_MANAGER_VERSION, Integration } from './constants'
 import { getChugSplashRegistryAddress } from './addresses'
 import 'core-js/features/array/at'
-import { ChugSplashRuntimeEnvironment } from './types'
 import {
   BuildInfo,
   CompilerOutput,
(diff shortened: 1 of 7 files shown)


packages/core/src/utils.ts
```diff
@@ -1158,8 +1156,12 @@ export const isLocalNetwork = async (
...
   }
 
-  if (await isHardhatFork(provider)) {
-    return false
+  try {
+    if (await isHardhatFork(provider)) {
+      return false
+    }
+  } catch (e) {
+    return true
```
sam-goldman, line 1164: Just curious, what situation is the try/catch necessary? The first try/catch in this function should prevent `isHardhatFork` function from being called on live networks

yarn.lock
sam-goldman, line 1: Should this file be modified? Doesn't look like you added or removed any packages in this PR

sam-goldman (approved):

packages/core/src/utils.ts
RPate97: `isHardhatFork` throws an error when run against anvil, and `isLocalNetwork` is called from the initializeChugSplash function which I'm using here to deploy the contracts against anvil. So this try-catch, makes it assume anvil is a local network no matter what. I did it this way b/c I didn't want to refactor the `initializeChugSplash` function to avoid using `isLocalNetwork` against anvil since we only need to use it until we have the raw transaction cheat code.

yarn.lock
RPate97: This file can be modified even if you do not add or remove any packages. This is because the way our dependency versions are defined in package.json technically allows us to install any minor or patch version which is higher than the specified version. So if you run `yarn install` you can sometimes get new versions which may also include new peer or subdependencies.

PR #746, opened 2023-06-09
Add support for other tasks in foundry, cancel, propose, init, import, export

RPate97:
- Adds support for other tasks in foundry
- Adds a standalone CLI tool that can be used in with the foundry plugin to run init and propose without needing to create Solidity scripts: npx chugsplash init, npx chugsplash propose. It'd be cool to add support for other stuff in this like the deploy task and exporting/importing proxies. Using the functions themselves in scripts is great, but I think it's nice to have the option of using a CLI tool as well.
- Cleans up the test suite so all the checks should pass (does not include additional tests, we should evaluate the testing requirements of the new foundry plugin separately)


sam-goldman:
> - Adds a standalone CLI tool that can be used in with the foundry plugin to run init and propose without needing to create Solidity scripts: `npx chugsplash init`, `npx chugsplash propose`.

So sick! Great idea

.gitmodules
```diff
@@ -0,0 +1,6 @@
+[submodule "packages/demo_foundry/lib/forge-std"]
+	path = packages/demo_foundry/lib/forge-std
+	url = https://github.com/foundry-rs/forge-std
+[submodule "packages/demo_foundry/lib/solidity-stringutils"]
+	path = packages/demo_foundry/lib/solidity-stringutils
+	url = https://github.com/Arachnid/solidity-stringutils
```
sam-goldman, line 6: Should these be included? I thought we were using `node_modules` from now on

sam-goldman (approved):

PR #754, opened 2023-06-12
Default unsafeAllow.flexibleConstructor to false

RPate97:
## Purpose
Sets the value of `unsafeAllow.flexibleConstructor` to false. This is a simple way to do it, for now, later we can come back and we'll have an easy spot to add in more complex logic for determining the value based on the type of contract or some other criteria.

### DIFF
--- a/packages/core/src/config/parse.ts
+++ b/packages/core/src/config/parse.ts
@@ -2208,6 +2208,23 @@ const constructParsedConfig = (
   return parsedConfig
 }
 
+export const setDefaultContractOptions = (
+  userConfig: UserChugSplashConfig
+): UserChugSplashConfig => {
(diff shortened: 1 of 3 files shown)


packages/core/src/config/parse.ts
```diff
@@ -2208,6 +2208,24 @@ const constructParsedConfig = (
   return parsedConfig
 }
 
+export const setDefaultContractOptions = (
+  userConfig: UserChugSplashConfig
+): UserChugSplashConfig => {
+  // Determine the addresses for all contracts.
```
sam-goldman, line 2214: nit: looks like a misplaced comment

sam-goldman (approved):

PR #756, opened 2023-06-13
Remove force during compilation

RPate97:
## 👀


packages/core/src/utils.ts
```diff
@@ -619,36 +619,39 @@ export const readBuildInfo = (buildInfoPath: string): BuildInfo => {
...
-  ) {
-    throw new Error(
-      `Did you forget to set the "storageLayout" compiler option in your Hardhat/Foundry config file?`
-    )
-  }
+  if (integration === 'hardhat') {
+    if (
+      !buildInfo.input.settings.outputSelection['*']['*'].includes(
+        'storageLayout'
+      )
```
RPate97, line 639: This check fails during partial rebuilds in Foundry b/c for some reason foundry outputs the settings on a per-contract basis so we have to do this check separately. That being said, there doesn't seem to be any underlying issue that makes the required info unavailable to us.

packages/plugins/src/foundry/get-bundle-info.ts
```diff
@@ -30,6 +33,19 @@ const userConfig = JSON.parse(userConfigStr)
...
+    const forgeConfig = JSON.parse(forgeConfigOutput.stdout)
+    const { extra_output }: { extra_output: string[] } = forgeConfig
+    if (
+      !extra_output.includes('storageLayout') ||
+      !extra_output.includes('evm.gasEstimates')
+    ) {
+      throw Error(
+        "foundry.toml file must include both 'storageLayout' and 'evm.gasEstimates' in 'extra_output':\n extra_output = ['storageLayout', 'evm.gasEstimates']"
+      )
+    }
```
RPate97, line 47: We check the storage layout and gas estimate options are properly enabled by reading them from the config directly.

packages/plugins/src/foundry/utils/index.ts
```diff
@@ -27,8 +27,8 @@ export const getBuildInfo = (
 ): BuildInfo => {
   // Find the correct build info file
   for (const input of buildInfos) {
-    if (input?.output?.sources[sourceName] !== undefined) {
-      validateBuildInfo(input)
+    if (input?.output?.contracts[sourceName] !== undefined) {
+      validateBuildInfo(input, 'foundry')
```
RPate97, line 31: During partial rebuilds, Foundry may sometimes generate build info files that have lots of contracts listed as sourced but which do not include those same contracts in the output (b/c they've already been compiled so there's no need). This means we need to find the correct build info file for each contract by searching for the first file which has the target contract listed in the outputs.

sam-goldman: > During partial rebuilds, Foundry may sometimes generate build info files that have lots of contracts listed as sources but do not include those same contracts in the output (b/c they've already been compiled so there's no need).

Sounds good to me. Fyi, this is precisely why Foundry's bug causes an error in `getOpenZeppelinStorageLayout`. The reason that it errors is because the `buildInfo.output.sources` for these contracts looks like this:
```
      "../contracts/contracts/ChugSplashManager.sol": {
        "id": 1,
        "ast": {}
      },
```

packages/plugins/src/foundry/utils/index.ts
```diff
@@ -51,7 +51,7 @@ export const getContractArtifact = async (
 
   if (!(await existsAsync(completeFilePath))) {
     throw new Error(
-      `Could not find artifact for: ${name}. Did you forget to import it in your script file?`
+      `Could not find artifact for: ${name}. Are you sure you've properly configured your src, script, and test directories in your foundry.toml file?`
```
sam-goldman, line 54: This sounds a bit vague to me. Maybe something like: `Could not find artifact for: ${name}. Please make sure that this contract exists in either the src, script, or test file that you've configured in your foundry.toml.`

packages/core/src/utils.ts
RPate97: This isn't really an issue in practice b/c foundry will automatically do a full rebuild if you run a script after changing your config.

packages/core/src/utils.ts
```diff
@@ -619,36 +619,39 @@ export const readBuildInfo = (buildInfoPath: string): BuildInfo => {
     fs.readFileSync(buildInfoPath, 'utf8')
   )
 
-  validateBuildInfo(buildInfo)
```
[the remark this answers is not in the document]
RPate97: This function is only called from `getPreviousStorageLayoutOZFormat` which is called after all the build info files are checked anyway, so it's really not necessary to do this check and keeping it would add unnecessary complexity. We could honestly probably refactor `getPreviousStorageLayoutOZFormat` to remove the need for this function at all since it's the only place where we still read the build info files separately from `makeGetConfigArtifacts`.

packages/plugins/src/foundry/utils/index.ts
RPate97: Interesting, since the previous build info files contain this information don't we technically have the necessary info to reconstruct the complete build info ourselves?

packages/core/src/utils.ts
sam-goldman: Ah nice, could you just document this in the relevant section of `get-bundle-info` then?

packages/plugins/src/foundry/get-bundle-info.ts
```diff
@@ -35,6 +37,19 @@ const broadcasting = args[2] === 'true'
     const { artifactFolder, buildInfoFolder, canonicalConfigFolder } =
       await getPaths()
 
+    const execAsync = promisify(exec)
+    const forgeConfigOutput = await execAsync('forge config --json')
```
sam-goldman, line 41: We already call `await execAsync('forge config --json')` once in `getPaths()` above. It's not the slowest operation, but it's not instant (~0.1s), so it'd be great if we could refactor this to only call `execAsync` once. Maybe we could pass in the foundryConfig as a parameter to `getPaths()`

packages/plugins/src/foundry/utils/index.ts
sam-goldman: We don't because we need the AST IDs, which differ across build infos

packages/core/src/utils.ts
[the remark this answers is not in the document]
sam-goldman: Sounds good, we can leave it as is. I don't think it makes sense to refactor the `readBuildInfo(previousBuildInfo)` call though. The`previousBuildInfo` is the build info file that was used in a previous deployment, not the current one, so I don't think it makes sense to add this to `getConfigArtifacts`. I still think it's fine not to include `validateBuildInfo` inside `readBuildInfo` though because it's probably safe to assume that the build info file contains the storage layouts.

sam-goldman (approved):

PR #759, opened 2023-06-13
Foundry Docs

RPate97:
## Purpose
- Updates the foundry getting started guide to reflect the current state of the foundry library. We'll still have to double check everything works once we release the plugin.
- Updates the sample project to generate a project that deploys an immutable contract.


docs/foundry/getting-started.md
```diff
@@ -87,232 +88,53 @@ forge-std/=lib/forge-std/src/
...
-        assertEq(helloChugSplash.otherStorage(), 0x1111111111111111111111111111111111111111);
-    }
-}
+forge test
 ```
 
-Notice that you need to call `chugsplash.refresh()` after calling `chugsplash.deploy(...)`.
-
-Run your tests using the command:
+## 7. Deploy on anvil
```
sam-goldman, line 115: IMO we should just deploy locally here via `forge script ...`. Then, we can skip the .env stuff and the need to start up a new anvil node.

docs/foundry/getting-started.md
```diff
@@ -87,232 +88,53 @@ forge-std/=lib/forge-std/src/
...
-  HelloChugSplash: 0x...
+forge script script/HelloChugSplash.s.sol
 ```
 
-If you update your tests in your test file, you'll be able to confirm that the contract was upgraded correctly.
+## Learn More
 
-That's all it takes to do an upgrade!
+Once you've set up your project, the next step is to learn about the [ChugSplash
+file](https://github.com/chugsplash/chugsplash/blob/develop/docs/chugsplash-file.md), which is where
```
sam-goldman, line 139: ChugSplash file -> ChugSplash config file

sam-goldman (approved):

docs/foundry/getting-started.md
RPate97: IMO the value of this step is to demonstrate the real deployment process including actually broadcasting the transactions so that it's clear how it works when broadcasting and so that this process can be replicated on a real network. I don't think there's much value in having this step at all if we aren't actually doing the broadcast. 

Also I realized I forgot to include a step on generating the deployment artifacts which is important IMO and will only work if the transactions are actually broadcast.

PR #766, opened 2023-06-15
readcallers

sam-goldman:


packages/plugins/contracts/foundry/ChugSplash.sol
```diff
@@ -827,12 +815,13 @@ contract ChugSplash is
...
 
-    function ffiDeployOnAnvil() private {
-        string[] memory cmds = new string[](6);
+    function ffiDeployOnAnvil(string memory _rpcUrl) private {
+        string[] memory cmds = new string[](5);
         cmds[0] = "npx";
         cmds[1] = "node";
         cmds[2] = mainFfiScriptPath;
         cmds[3] = "deployOnAnvil";
+        cmds[4] = _rpcUrl;
```
RPate97, line 824: I don't think we need the rpcUrl parameter here. There's no case where it would be anything other than `localhost:8545` or `127.0.0.1:8545`.

PR #770, opened 2023-06-16
Fix Proposal Bugs

RPate97:
## Purpose
- Fixes an issue where proposal would not exit early if there is nothing to deploy
- Fixes an issue where errors caught during proposals were not properly output
- Fixes an issue where proposal success messages do not accurately reflect the outcome (i.e successful b/c a deployment was actually proposed, or successful b/c there were no changes since the last deployment)

## Note
Does not fix a related issue with the deploy function where the success message also does not accurately reflect the actual outcome of the deployment. If you deploy a config with a single immutable contract and then attempt to deploy that same config again, it will succeed with the same message both times: 
```
== Logs ==
  Success!
  Stateless: <address>
```
But the underlying behavior is different. The first time the contract will actually be deployed, the second time the contract deployment is skipped over. In the hardhat plugin, we detect this and output a warning that the deployment is being skipped. If there is nothing to deploy at all, then we exit early with a message indicating that there was nothing to deploy.

[I've created a ticket for this.](https://linear.app/chugsplash/issue/CHU-258/fix-deployment-success-message-in-foundry-when-nothing-is-deployed)

### DIFF
--- a/packages/plugins/contracts/foundry/Propose.sol
+++ b/packages/plugins/contracts/foundry/Propose.sol
@@ -1,3 +1,6 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.15;
+
 import { ChugSplash } from "./ChugSplash.sol";
 
 contract Propose is ChugSplash {
--- a/packages/core/src/tasks/index.ts
+++ b/packages/core/src/tasks/index.ts
@@ -55,7 +55,11 @@ import {
   getOwnerWithdrawableAmount,
 } from '../fund'
 import { monitorExecution } from '../execution'
-import { ChugSplashRuntimeEnvironment, ProposalRoute } from '../types'
+import {
+  ChugSplashRuntimeEnvironment,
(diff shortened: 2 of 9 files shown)


packages/plugins/contracts/foundry/Propose.sol
```diff
@@ -1,3 +1,6 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.15;
```
sam-goldman, line 2: What happens here if the user is trying to deploy with a lower Solidity version, e.g. 0.8.0?

RPate97: They just specify the compiler version they want in their contract files. I believe this only specifies the version to use for this specific file. Our current test contracts are actually already like this. They use various versions below 0.8.15, but `ChugSplash.sol` uses 0.8.15 and it doesn't seem like an issue.

sam-goldman (approved):

PR #772, opened 2023-06-16
ci: build and push docker image

sam-goldman:


.github/workflows/docker-image.yml
```diff
@@ -0,0 +1,35 @@
+name: Publish Docker image
+
+on:
+  push:
+    branches:
+      - develop
```
RPate97, line 6: This is correct, but if you want to test this workflow before merging then you should use this:
```suggestion
on:
  pull_request:
    types: [ synchronize ]
```

Just make sure to switch it back when you're done.

.github/workflows/docker-image.yml
```diff
@@ -0,0 +1,54 @@
...
+    outputs:
+      changed: ${{ steps.step1.outputs.didChange }}
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - id: step1
+        run: >
+          chmod +x ./ops/ci-builder/check-changed.sh &&
```
RPate97, line 20: Might be able to remove the use of chmod here. Generally the ci runner will have full permissions. 

Optimism had to use this when constructing their docker container, but that’s a bit of a special case because they wanted the container to be able to use it while it’s running.

.circleci/config.yml
```diff
@@ -3,7 +3,10 @@ version: 2.1
 jobs:
   yarn-monorepo:
     docker:
-      - image: ethereumoptimism/ci-builder:latest
+      - image: samgoldman/chugsplash:pr-772
```
RPate97, line 6: Couple things here:

1. ChugSplash is a bit of a generic name for this, it would be nice to give the container a more specific name. “ci-builder” is just fine.

2. Since we’ve hardcoded a specific tag here, this won’t get auto updated when we publish a new package. So we’ll have to come bump this manually whenever we update the container. Not a huge deal, just kind of annoying. If you want to fix it, the solution is to use a latest tag for this (samgoldman/ChugSplash:latest). The best practice is to supply two tags when you push a new container version. One should be for the version number and the other for latest. That way we’ll end up with a set of container version tags in the registry and a singe latest tag which just points to the most recent one. 

3. The use of a the pr number as a version number is pretty wonky. However, I see why that would be easy and since this isn’t for external usage I think it’s fine. I just wanted to point out that’s pretty atypical.

RPate97 (approved):
Looks pretty solid, I left a few comments on using container tags. None of them strictly need to be addressed though, they’re just nice to have’s.

.circleci/config.yml
sam-goldman: Sounds good, I'll make these changes

> The best practice is to supply two tags when you push a new container version. One should be for the version number and the other for latest.

Which version number are you referring to?

RPate97: Just some version indicator. In this case, it seems to be automatically using the PR number. Honestly, I think that’s fine b/c there’s not really an easy way to generate a real version number here.

sam-goldman: Sounds good, I ended up using the commit hash as the second tag.

PR #791, opened 2023-06-17
fix(pg): Fix bugs in sample project generation

RPate97:
## Purpose
Fixes some issues with the sample foundry project.


packages/plugins/src/foundry/utils/index.ts
```diff
@@ -133,7 +133,7 @@ export const makeGetConfigArtifacts = (
  */
 export const inferSolcVersion = async (): Promise<string> => {
   // This is the default solc version used by Foundry's "Getting Started" guide.
-  const defaultSolcVersion = '0.8.20'
+  const defaultSolcVersion = '0.8.19'
```
RPate97, line 136: I couldn't find any references to this compiler version in the foundry book, and when I actually generated a project there was a linter error due to this default. So I switched it to 0.8.19 since that seems to work fine.

PR #824, opened 2023-06-23
fix(pg): Use cache to avoid reading all build info files

RPate97:
## Purpose
Uses a cache to fetch the correct build info files for each contract source name to avoid reading all of the build info files.

## Note
I investigated a separate issue where we require that contracts be located in a file that matches their exact name. This is an issue with fetching the correct contract artifact and not the build info. I believe that this caching scheme would actually work fine for contracts in files with different names. We'll have to test it and see when we address that issue. I determined it to be outside the scope of this specific PR.

### DIFF
--- a/packages/plugins/src/foundry/get-bundle-info.ts
+++ b/packages/plugins/src/foundry/get-bundle-info.ts
@@ -38,6 +38,7 @@ const broadcasting = args[2] === 'true'
       canonicalConfigFolder,
       storageLayout,
       gasEstimates,
+      cachePath,
     } = await getFoundryConfigOptions()
 
     if (!storageLayout || !gasEstimates) {
@@ -72,7 +73,8 @@ const broadcasting = args[2] === 'true'
 
     const getConfigArtifacts = makeGetConfigArtifacts(
       artifactFolder,
-      buildInfoFolder
+      buildInfoFolder,
+      cachePath
     )
 
     const configArtifacts = await getConfigArtifacts(userConfig.contracts)
@@ -102,16 +104,16 @@ const broadcasting = args[2] === 'true'
       writeCanonicalConfig(canonicalConfigFolder, configUri, canonicalConfig)
 
       const ipfsHash = configUri.replace('ipfs://', '')
-      const cachePath = path.resolve('./cache')
+      const artifactCachePath = path.resolve(`${cachePath}/configArtifacts`)
       // Create the canonical config network folder if it doesn't already exist.
-      if (!fs.existsSync(cachePath)) {
-        fs.mkdirSync(cachePath)
+      if (!fs.existsSync(artifactCachePath)) {
+        fs.mkdirSync(artifactCachePath)
       }
 
       // Write the config artifacts to the local file system. It will exist in a JSON file that has the
       // config URI as its name.
       fs.writeFileSync(
-        path.join(cachePath, `${ipfsHash}.json`),
+        path.join(artifactCachePath, `${ipfsHash}.json`),
         JSON.stringify(configArtifacts, null, 2)
       )
     }
--- a/packages/plugins/src/foundry/utils/index.ts
+++ b/packages/plugins/src/foundry/utils/index.ts
@@ -1,5 +1,5 @@
 import * as fs from 'fs'
-import { join } from 'path'
+import path, { join } from 'path'
 import { promisify } from 'util'
 
 import {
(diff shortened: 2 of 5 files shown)


packages/plugins/src/foundry/utils/index.ts
```diff
@@ -89,40 +123,116 @@ export const makeGetConfigArtifacts = (
...
+        async ([referenceName, contractConfig]) => {
+          const artifact = await getContractArtifact(
+            contractConfig.contract,
+            artifactFolder
+          )
+
+          // Just make sure the files are sorted by time
+          const sortedCachedFiles = Object.values(buildInfoCache).sort(
+            (a, b) => b.time - a.time
+          )
```
sam-goldman, line 163: It seems like this can go outside the `map`

packages/plugins/src/foundry/utils/index.ts
```diff
@@ -74,12 +77,43 @@ export const getContractArtifact = async (
...
 
-    const buildInfoPromises = fs
+    // If the required files are missing then clear the cache
+    // We only check the first file because we can safely assume that if any file is missing then
+    // the first one will also be missing due to how the build info files are generated.
+    // The only case where this wouldn't be true is if the user manually deleted the file, which
+    // is not a legitimate use case.
+    if (
+      Object.keys(buildInfoCache).length > 0 &&
+      !fs.existsSync(`${buildInfoPath}/${Object.keys(buildInfoCache)[0]}`)
```
sam-goldman, line 110: I think we should use `path.join` here instead of using a platform-specific delimiter, `/`

packages/plugins/src/foundry/utils/index.ts
```diff
@@ -89,40 +123,116 @@ export const makeGetConfigArtifacts = (
         time: fs.statSync(`${buildInfoPath}/${fileName}`).mtime.getTime(),
```
sam-goldman, line 123: > I think we should use `path.join` here instead of using a platform-specific delimiter, `/`

Ik you didn't touch this line in this PR, but we should probably fix this too

packages/plugins/src/foundry/utils/index.ts
```diff
@@ -89,40 +123,116 @@ export const makeGetConfigArtifacts = (
...
+            artifactFolder
+          )
+
+          // Just make sure the files are sorted by time
+          const sortedCachedFiles = Object.values(buildInfoCache).sort(
+            (a, b) => b.time - a.time
+          )
+          // Look through the cahce for the first build info file that contains the contract
+          for (const file of sortedCachedFiles) {
+            if (file.contracts.includes(artifact.sourceName)) {
```
sam-goldman, line 166: Should we keep track of fully qualified names (source name + contract name) instead of just the source name in `file.contracts`? Then we'd be able to compare this against the artifact's fully qualified name here. It feels a little safer since I'm not confident about how the solidity compiler behaves in edge cases involving multiple contracts in the same source file

packages/plugins/src/foundry/utils/index.ts
```diff
@@ -89,40 +123,116 @@ export const makeGetConfigArtifacts = (
...
+    const localBuildInfoCache = {}
+    await Promise.all(
+      buildInfoFileNames
+        .filter((file) => buildInfoCache[file.name]?.time !== file.time)
+        .map(async (file) => {
+          const buildInfo = JSON.parse(
+            await readFileAsync(join(buildInfoFolder, file.name), 'utf8')
+          )
+
+          // We keep track of the last modified time in each build info file so we can check the file if it has changed.
```
sam-goldman, line 138: I'm pretty sure that each build info file's name is a hash of its contents, so we don't need to worry about a build info file having the same name but different contents. Is there a place in this function where we actually check if the contents have changed? If so, is it necessary? Regardless, it makes sense that we update the file.time

packages/plugins/src/foundry/utils/index.ts
```diff
@@ -74,12 +77,43 @@ export const getContractArtifact = async (
...
+
     const buildInfoPath = join(buildInfoFolder)
 
-    const buildInfoPromises = fs
+    // If the required files are missing then clear the cache
+    // We only check the first file because we can safely assume that if any file is missing then
+    // the first one will also be missing due to how the build info files are generated.
+    // The only case where this wouldn't be true is if the user manually deleted the file, which
+    // is not a legitimate use case.
+    if (
```
sam-goldman, line 108: Is there a place in this function where it'd be convenient to make this check on every build info in the cache, and not just the first one? This seems fine, but it'd be great if we could check this in a less hacky way

sam-goldman (approved):
Left some minor comments, nothing blocking

sam-goldman:
Have you timed this on a repo with a bunch of build info files to sanity check that this improves performance?

packages/plugins/src/foundry/utils/index.ts
RPate97: I don't think this should be necessary here because we're fetching the source name from the contract artifact, so I don't see a strong reason why we would need the fully qualified name here. As long as we have the correct contract artifact, we'll have the correct source name here as well.

This seems related to the other bug you pointed out with how we're fetching contract artifacts where there's some ambiguity if the user has multiple contracts of the same name. That ambiguity does affect this code since we're assuming that the contract artifact is correct, but I don't think we should try to address the ambiguity here. I think we should just resolve the root issue.

packages/plugins/src/foundry/utils/index.ts
RPate97: We pretty much already do that. When we read the files, we detect if one is missing and throw an error recommending the user recompile with force. We could also do the check separately, but would be a wasted operation.

RPate97:
> Have you timed this on a repo with a bunch of build info files to sanity check that this improves performance?

Yes, it performs as expected. It's a few hundred milliseconds slower in general but remains at roughly the same speed regardless of the number of build info files. I checked it with like 20 files.

packages/plugins/src/foundry/get-bundle-info.ts
```diff
@@ -102,16 +104,18 @@ const broadcasting = args[2] === 'true'
       writeCanonicalConfig(canonicalConfigFolder, configUri, canonicalConfig)
 
       const ipfsHash = configUri.replace('ipfs://', '')
-      const cachePath = path.resolve('./cache')
+      const artifactCachePath = path.resolve(
+        '.canonical-configs/configArtifacts'
```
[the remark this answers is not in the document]
RPate97: We'd be putting them in the foundry cache actually since we pull that from the config file. Yeah, it's probably better to put them there since the user shouldn't be committing these to version control anyway.

PR #825, opened 2023-06-23
feat(pg): Support fully qualified names in foundry

RPate97:
## Purpose
Adds support for fully qualified names in foundry


packages/plugins/src/foundry/utils/index.ts
```diff
@@ -45,19 +45,36 @@ export const getContractArtifact = async (
...
+  const [file, contract] = qualifiedSections?.split(':') ?? ['', '']
+  const qualifiedFilePath = join(artifactFilder, file, `${contract}.json`)
+
+  if (await existsAsync(standardFilePath)) {
+    return parseFoundryArtifact(
+      JSON.parse(await readFileAsync(standardFilePath, 'utf8'))
+    )
+  } else if (await existsAsync(qualifiedFilePath)) {
+    return parseFoundryArtifact(
+      JSON.parse(await readFileAsync(qualifiedFilePath, 'utf8'))
```
sam-goldman, line 66: It seems like this doesn't mirror Hardhat's `hre.artifacts.readArtifact` behavior in the following case. If there are two contracts with the name "MyContract", where one has the fully qualified name "MyContract.sol:MyContract", and the other's is "MyRandomFile.sol:MyContract", then it seems like this function will always choose the former contract if the user specifies "MyContract" in their config. However, in this case, Hardhat will actually throw an error that prompts the user to supply a fully qualified name because "MyContract" is ambiguous. I think we should follow this pattern here. If you agree, I think there are a few ways we could tackle this. The main consideration is that we need to retrieve all of the contract names in the artifacts directory, since we need to throw an error in the situation above. Some ideas for getting the contract names:
1. Iterate over every contract artifact file. This is probably the slowest option by far, so it's my least favorite solution.
2. Use the `forge tree` CLI command, then do some light parsing on the output. I looked at all of the CLI commands for cast and forge, and this looked like the only one that we could use for this purpose.
3. Load the `solidity-files-cache.json` auto-generated by Foundry in their cache file. It seems like this contains all of the contract names. Note that it's possible for users to specify `cache=false` in their foundry.toml, which would prevent this file from being created.
4. Use our existing build info cache to keep track of the contract names. You'd know better than me if we can do this.
Option 3 seems like the easiest and most performant. Option 4 might be solid too if we can easily re-use the build info cache. If we choose option 3, we should probably check if the user has `cache=false` in their config and throw an error if so.

RPate97: That is a good catch, however, I would resolve this separately because this bug is not strictly tied to this specific PR and the solution may involve using the build info cache from my other PR. I've created a separate ticket for this.

PR #828, opened 2023-06-29
feat(ct): Add crosschain funding contract

RPate97:
## Purpose
Adds cross-chain funding contracts

### DIFF
--- a/packages/contracts/contracts/ChugSplashFunder.sol
+++ b/packages/contracts/contracts/ChugSplashFunder.sol
@@ -1,0 +1,108 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import { LzApp } from "@layerzerolabs/solidity-examples/contracts/lzApp/LzApp.sol";
+import {
+    ILayerZeroEndpoint
+} from "@layerzerolabs/solidity-examples/contracts/interfaces/ILayerZeroEndpoint.sol";
+import { FunderAction } from "./ChugSplashDataTypes.sol";
+
+contract ChugSplashFunder is LzApp {
+    constructor(address _endpoint) LzApp(_endpoint) {}
+
+    /**
+     * @notice Handles sending a batch of fund actions
+     *
+     * @param actions Set of FunderActions
+     */
+    function sendBatchFundActions(FunderAction[] memory actions) public payable {
+        // Calculate the total cost
+        uint total = 0;
+        uint zroAmount = 0;
+        bytes[] memory adapterParams = new bytes[](actions.length);
+        uint[] memory nativeFees = new uint[](actions.length);
+        for (uint i = 0; i < actions.length; i++) {
+            // Require the airdrop value be above 0 and not being sent to the zero address
+            require(actions[i].airdropAmount > 0, "Airdrop amount must be greater than 0");
+            require(actions[i].airdropAddress != address(0), "Cannot airdrop to address 0");
+
+            // use adapterParams v1 to specify more gas for the destination
+            uint16 version = 2;
+            uint gasForDestinationLzReceive = 200000;
+            adapterParams[i] = abi.encodePacked(
+                version,
+                gasForDestinationLzReceive,
+                // We assume the user checked the airdrop limit offchain
+                actions[i].airdropAmount,
+                actions[i].airdropAddress
+            );
+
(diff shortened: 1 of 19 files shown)


packages/contracts/contracts/ChugSplashFunder.sol
```diff
@@ -0,0 +1,106 @@
...
+            _adapterParams
+        );
+    }
+
+    function _blockingLzReceive(
+        uint16 _srcChainId,
+        bytes memory _srcAddress,
+        uint64 _nonce,
+        bytes memory _payload
+    ) internal override {}
```
RPate97, line 105: Just silences a warning due to not overriding this when inheriting LzApp.

packages/core/src/constants.ts
```diff
@@ -37,3 +37,18 @@ export const CURRENT_CHUGSPLASH_MANAGER_VERSION = {
...
+  },
+  420: {
+    address: '0xae92d5aD7583AD66E49A0c67BAd18F6ba52dDDc1',
+    lzChainId: 10132,
+  },
+}
+
+export const LAYERZERO_RELAYER_ADDRESSES = {
+  1: '0x902F09715B6303d4173037652FA7377e5b98089E',
+}
```
RPate97, line 54: LayerZero does have a package with addresses, but I don't think we should use it for a couple reasons:
1. It's an alpha package that they warn could be unstable
2. It's distributed under a business license which is super weird tbh

packages/core/src/contract-info.ts
```diff
@@ -34,84 +36,105 @@ import {
...
+    {
+      artifact: FunderArtifact,
+      expectedAddress: getFunderAddress(lzEndpointAddress),
+      constructorArgs: [lzEndpointAddress],
+    },
+    {
+      artifact: LZReceiverArtifact,
+      expectedAddress: getLZReceiverAddress(lzEndpointAddress),
+      constructorArgs: [lzEndpointAddress],
+    },
```
RPate97, line 138: LayerZero endpoint contracts aren't consistent across networks, so we have to fetch the correct addresses based on the chain id.

packages/plugins/script/write-constants.ts
```diff
@@ -93,7 +93,7 @@ const writeConstants = async () => {
     },
   }
 
-  const contractInfo = CHUGSPLASH_CONTRACT_INFO.map(
+  const contractInfo = getChugSplashConstants(31337).map(
```
RPate97, line 96: We don't need the funding-related contracts in here, so we use a default chain id and then filter out the funding-related contracts below so they don't even make it into the constants file.

packages/plugins/script/write-constants.ts
```diff
@@ -93,8 +93,13 @@ const writeConstants = async () => {
...
 
-  const contractInfo = CHUGSPLASH_CONTRACT_INFO.map(
-    ({ artifact, constructorArgs, expectedAddress }) => {
+  const contractInfo = getChugSplashConstants(31337)
+    .filter(
+      (el) =>
+        el.artifact.contractName !== 'ChugSplashFunder' &&
+        el.artifact.contractName !== 'ChugSplashLZReceiver'
+    )
+    .map(({ artifact, constructorArgs, expectedAddress }) => {
```
RPate97, line 102: We don't need the funding-related contracts in the constants file since the Foundry plugin doesn't need to interact with them. So we use a default chain id when calling `getChugSplashConstants` and then filter out the funding-related contracts below so they don't even make it into the constants file.

packages/core/src/constants.ts
```diff
@@ -37,3 +37,14 @@ export const CURRENT_CHUGSPLASH_MANAGER_VERSION = {
...
+export const LAYERZERO_ENDPOINT_ADDRESSES = {
+  5: {
+    address: '0xbfD2135BFfbb0B5378b56643c2Df8a87552Bfa23',
+    lzChainId: 10121,
+  },
+  420: {
+    address: '0xae92d5aD7583AD66E49A0c67BAd18F6ba52dDDc1',
+    lzChainId: 10132,
+  },
+}
```
RPate97, line 50: LayerZero does have a package with addresses, but I don't think we should use it for a couple of reasons:
1. It's an alpha package that they warn could be unstable
2. It's distributed under a business license (disallows commercial use) which is super weird tbh

PR #842, opened 2023-07-11
fix(pg): update init tasks and add tests for TypeScript init tasks

sam-goldman:


.circleci/config.yml
```diff
@@ -23,7 +23,7 @@ jobs:
             - v2-cache-yarn-install
       - run:
           name: Install dependencies
-          command: yarn --frozen-lockfile
+          command: yarn --frozen-lockfile --link-duplicates
```
RPate97, line 26: https://github.com/yarnpkg/yarn/issues/5487

PR #861, opened 2023-07-22
fix(core): Complete integration of new config structure

RPate97:


packages/core/src/tasks/index.ts
```diff
@@ -314,9 +314,14 @@ export const proposeAbstractTask = async (
...
+  // Note that calculating this here and passing in a single value works as long as the address
+  // is the same on all networks, but we may need to change this in the future to support chains
+  // which calculate addresses in different ways. I.e ZKSync Era
   const authAddress = getAuthAddress(
-    prevConfig.options.owners,
-    prevConfig.options.threshold,
+    parsedConfig.options.owners,
+    parsedConfig.options.threshold,
     parsedConfig.project
   )
```
RPate97, line 326: The website relies on this address being correct for any new networks. If/when we support deploying on networks with inconsistent addresses, we'll need to update this logic to handle cases where the address differs between networks. 

Also if/when we add support for changing project owners in the config, we'll need to revisit this to make sure the address management strategy on the site will continue to function correctly after making that change.

PR #868, opened 2023-07-26
feat(exec): Track transaction receipts during execution

RPate97:


packages/contracts/contracts/SphinxBalance.sol
```diff
@@ -37,6 +37,7 @@ contract SphinxBalance is Ownable {
         orgId = _orgId;
         usdc = ERC20(_usdc);
         escrow = _escrow;
+        usdc.increaseAllowance(_escrow, type(uint256).max);
```
RPate97, line 40: This makes it so the user only has to perform a single transaction to deposit funds into the balance contract and then the escrow contract automatically has the necessary allowance to withdraw. Without setting this, we need a two-step process to deposit and then set the escrow contract allowance. 

My thought process is that explicitly setting the allowance is unnecessary. If the user does not want to trust us with their funds by maintaining a balance, then they are able to simply deposit exactly the funds their deployment needs and then withdraw the excess from the balance contract later. 

You might ask why not just send the funds directly to the escrow contract to avoid doing this. The reason is doing that would require me to maintain two flows on the website and would also make refunding the excess funds after the deployment is complete more difficult. I.e if we send the funds to their balance contract, then how will they be used in the next deployment if their typical flow involves sending funds directly to the escrow contract? We would probably have to send then funds back to the specific user who deposited them which just gets complicated. All around, I think this makes things a lot cleaner on the website side. 

It's worth noting that while [unlimited erc20 allowances are generally considered to be a bad practice](https://kalis.me/unlimited-erc20-allowances/#:~:text=However%2C%20this%20setup%20comes%20with,%22safely%22%20in%20your%20wallet.), I believe this design is perfectly fine because the allowance is set on the balance contract vs being set on a users wallet. So the only funds at risk due to the allowance are those explicitly transferred to the balance contract, not funds in the users' wallet.

packages/contracts/contracts/SphinxBalanceFactory.sol
```diff
@@ -17,7 +17,9 @@ contract SphinxBalanceFactory {
         string indexed orgIdHash,
         address owner,
         string orgId,
-        address caller
+        address caller,
+        address balance,
+        address escrow
```
RPate97, line 22: Including the addresses here is just convenient.

RPate97, line 22: Including the addresses here is just convenient since importing the necessary functions to calculate these addresses in the website backend was causing issues for me. I think I need to reorganize the project structure such that the website is a git module inside the mono repo, but it's not a high priority right now so I just fixed it by doing this. Also I think this is generally good info to include on the event anyway.

packages/contracts/contracts/SphinxBalance.sol
sam-goldman: Sounds good. Could you update the docs in this contract (and perhaps the others if relevant) to reflect this design decision?

packages/contracts/contracts/SphinxBalance.sol
sam-goldman, line 1: SInce we're not using the `increaseAllowance` and `decreaseAllowance` functions on this contract, how about we remove them? If that sounds good, could you also update the docs and remove the corresponding unit tests?

sam-goldman (approved):

sam-goldman:
Do we use both the `batchTransferFrom` and `transferFrom` functions in the `SphinxEscrow` contract?

packages/contracts/contracts/SphinxBalance.sol
sam-goldman: The more detail the better. Feel free to mention the thing about unlimited ERC20 approvals if you want to

RPate97:
> Do we use both the `batchTransferFrom` and `transferFrom` functions in the `SphinxEscrow` contract?

Yeah I'm currently using both in the website.

PR #988, opened 2023-08-25
fix(core): Combine previous canonical config with new config

RPate97:
## Purpose
Fixes an issue where removing some data from your config file can result in our DB containing a slightly out-of-date canonical config which can result in bugged deployments. 

For example prior to this PR, if I deployed with this config: 
```
const config: UserConfigWithOptions = {
  projectName: 'Foundry Deployment',
  options: {
    orgId: <ORG_ID>,
    owners: [<OWNER>],
    ownerThreshold: 1,
    testnets: ['arbitrum-goerli'],
    mainnets: [],
    proposers: [<PROPOSER>],
  },
  contracts: ...
}
```

And then I deployed with this config: 
```
const config: UserConfigWithOptions = {
  projectName: 'Foundry Deployment',
  options: {
    orgId: <ORG_ID>,
    owners: [<OWNER>],
    ownerThreshold: 1,
    testnets: ['optimism-goerli'],
    mainnets: [],
    proposers: [<PROPOSER>],
  },
  contracts: ...
}
```

The canonical config in the DB would contain chain status' for only the chain status for optimism goerli. So then, if I deployed with this config the deployment would bug out because our system would think the user never deployed on optimism goerli. 
```
const config: UserConfigWithOptions = {
  projectName: 'Foundry Deployment',
  options: {
    orgId: <ORG_ID>,
    owners: [<OWNER>],
    ownerThreshold: 1,
    testnets: ['optimism-goerli', 'arbitrum-goerli'],
    mainnets: [],
    proposers: [<PROPOSER>],
  },
  contracts: ...
}
```

This PR fixes this by making sure the previous canonical config is combined with the new canonical config. Even though we don't recommend doing something like this, we need to handle it unless we actually implement logic for preventing the user from removing stuff from their config which I don't think we should. In the future, if we add more stuff to the canonical config, or change it we'll need to make sure we continue to handle doing the combination.

### DIFF
--- a/packages/core/src/tasks/index.ts
+++ b/packages/core/src/tasks/index.ts
@@ -358,9 +358,15 @@ export const proposeAbstractTask = async (
   const newCanonicalConfig: CanonicalConfig = {
     manager: prevConfig.manager,
     options: parsedConfig.options,
-    contracts: parsedConfig.contracts,
+    contracts: {
+      ...prevConfig.contracts,
+      ...parsedConfig.contracts,
+    },
     projectName: parsedConfig.projectName,
-    chainStates: newChainStates,
+    chainStates: {
+      ...prevConfig.chainStates,
+      ...newChainStates,
+    },
   }
 
   // We calculate the auth address based on the current owners since this is used to store the
--- a/.changeset/rare-sloths-invent.md
+++ b/.changeset/rare-sloths-invent.md
@@ -1,0 +1,5 @@
+---
+'@sphinx-labs/core': patch
+---
+
+Combine previous canonical config with new config


packages/core/src/tasks/index.ts
```diff
@@ -358,9 +358,15 @@ export const proposeAbstractTask = async (
   const newCanonicalConfig: CanonicalConfig = {
     manager: prevConfig.manager,
     options: parsedConfig.options,
```
sam-goldman, line 360: I think we may need to combine the `options.chainIds` too. In the example you showed, it seems like the first chain ID (arbitrum goerli) would be removed from this field

sam-goldman (approved):

packages/core/src/tasks/index.ts
RPate97: Are you sure? The `ConfigOptions` type does not appear to include the `chainIds` field: 
https://github.com/sphinx-labs/sphinx/blob/804309f8cc33d2a9a3fcabff081870daf63bbdb7/packages/core/src/config/types.ts#L120

Only the `ParsedConfigOptions` type includes that field, but it is not used here.

PR #994, opened 2023-08-31
fix(core): Better failure logging

RPate97:


packages/contracts/contracts/SphinxManager.sol
```diff
@@ -938,11 +935,11 @@ contract SphinxManager is
...
     // TODO(docs)
     function _deploymentFailed(DeploymentState storage _deployment, uint256 _actionIndex) private {
-        activeDeploymentId = bytes32(0);
-        _deployment.status = DeploymentStatus.FAILED;
-
         emit DeploymentFailed(activeDeploymentId, _actionIndex);
         registry.announceWithData("DeploymentFailed", abi.encodePacked(activeDeploymentId));
+
+        activeDeploymentId = bytes32(0);
+        _deployment.status = DeploymentStatus.FAILED;
```
RPate97, line 942: We have to update the `activeDeploymentId` after emitting the event b/c it's included in the event.

packages/core/src/fund.ts
```diff
@@ -22,15 +23,25 @@ export const estimateExecutionGas = async (
...
+        try {
+          // We estimate the gas for the contract deployment by calling `estimateGas` on the provider.
+          return await provider.estimateGas({
+            data: action.code,
+          })
+        } catch (e) {
+          // If the estimate fails, we return a default value of 500k gas which is plenty since the actual
+          // deployment will not happen on chain.
+          return BigInt(500_000)
+        }
```
RPate97, line 44: Estimating the gas here throws an error if the constructor reverts, so we catch it here and use a conservative estimate instead. 

Context: We have tooling that attempts to prevent the user from proposing something that will fail during execution. Previously we had similar tooling in the executor as well, where we would check if it seems like a deployment will fail part of the way through and we do not execute it if so. However, since we have an on-chain process for handling if a deployment fails part of the way through, and that process is hit when a post-deployment action fails. It would be better to have consistent behavior where we still attempt the deployment even if a constructor may revert and then run through the same on-chain failure process in all cases.

PR #995, opened 2023-09-01
feat(core): Support manager upgrades

RPate97:


packages/core/src/actions/bundle.ts
```diff
@@ -793,6 +800,26 @@ export const getAuthLeafsForChain = async (
   // because the first two indexes are reserved for the setup and proposal leafs.
   let index = firstProposalOccurred ? 1 : 2
 
+  const managerVersionString = `v${configCache.managerVersion.major}.${configCache.managerVersion.minor}.${configCache.managerVersion.patch}`
```
sam-goldman, line 803: Is the manager version always equal to the auth version?

packages/core/src/actions/bundle.ts
```diff
@@ -793,6 +800,26 @@ export const getAuthLeafsForChain = async (
...
+  if (
+    managerVersionString !== parsedConfig.options.managerVersion &&
+    firstProposalOccurred
```
sam-goldman, line 806: Why is `firstProposalOccurred` in the if-statement clause here?

packages/core/src/languages/solidity/predeploys.ts
```diff
@@ -334,7 +336,10 @@ export const initializeSphinx = async (
     SphinxRegistryABI,
     owner
   )
-  const sphinxManagerV1Address = getSphinxManagerV1Address(Number(chainId))
+  const sphinxManagerV1Address = getSphinxManagerImplAddress(
```
sam-goldman, line 339: nit: I think we can remove V1

packages/core/src/utils.ts
```diff
@@ -104,11 +104,19 @@ import {
...
 import { assertValidConfigOptions, parseConfigOptions } from './config/parse'
 import { SphinxRuntimeEnvironment, FailureAction } from './types'
-import {
-  SUPPORTED_NETWORKS,
-  SupportedChainId,
-  SupportedNetworkName,
-} from './networks'
+import { SUPPORTED_NETWORKS, SupportedChainId } from './networks'
+
+export const parseSemverVersion = (version: string) => {
```
sam-goldman, line 109: It appears this function would return a valid semver object for a string where the "v" is anywhere, e.g. "1.0v.2". This is b/c "replace" just finds the first instance of "v"

packages/core/src/languages/solidity/predeploys.ts
sam-goldman: Looks like there's a few remaining instances of "V1" in the TypeScript/Solidity logic

packages/contracts/contracts/SphinxManager.sol
sam-goldman, line 1: The Natspec for this contract still has: `* @custom:version 1.0.0`. I think it'd be fine to remove this field since it doesn't serve any purpose, and it's one less thing to remember to update

packages/contracts/contracts/SphinxAuth.sol
sam-goldman, line 1: > The Natspec for this contract still has: `* @custom:version 1.0.0`. I think it'd be fine to remove this field since it doesn't serve any purpose, and it's one less thing to remember to update

Same w/ this file

sam-goldman (approved):

packages/core/src/utils.ts
RPate97: I don't think that really needs to be handled by this function b/c the assumption is that this function is getting called with a `ValidManagerVersion` string since we perform validation on the input earlier in the deployment process. I do think the input type should be changed to `ValidManagerVersion` though.

PR #996, opened 2023-09-01
feat: post-deployment actions

sam-goldman:


docs/post-deployment-actions.md
```diff
@@ -0,0 +1,269 @@
...
+postDeploy: [
+  MyContract.increment(),
+  MyContract.setMyValue(2),
+  MyContract.transferOwnership('0x1234...')
+]
+```
+
+The post-deployment actions are executed in the order they're defined in the `postDeploy` array. So, in the example above, the order would be: `increment`, `setMyValue`, then `transferOwnership`.
+
+Like contract deployments, your post-deployment actions are idempotent. This means that they'll only be executed once per chain, even if you re-deploy your config file.
```
RPate97, line 55: Nit: It would be nice to provide a bit more detail here. I.e explaining that if the input arguments change for a specific chain, then we will treat that as a new action and it will get executed.

docs/post-deployment-actions.md
```diff
@@ -0,0 +1,269 @@
...
+
+## Calling external contracts
+
+You may want to call a function on a contract that you're not deploying in your config file. Here's how to do this in your config file:
+
+```ts
+import { Contract, UserSphinxConfig } from '@sphinx-labs/core'
```
RPate97, line 62: We should specifically note that this is a custom contract object that has to be imported from our package. When I was first trying this out, I accidentally imported the `Contract` object from ethers instead.

RPate97 (approved):

docs/post-deployment-actions.md
```diff
@@ -0,0 +1,269 @@
...
+```ts
+const MyOwnableContract = new Contract('{{ MyOwnableContract }}')
+
+const config: UserSphinxConfig = {
+  contracts: {
+    MyOwnableContract: {
+      constructorArgs: {
+        // This contract reference resolves to the address of the
+        // SphinxManager contract:
+        _sphinxManager: '{{ SphinxManager }}'
```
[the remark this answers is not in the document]
sam-goldman: Yep, already accounted for that :)

[Validation](https://github.com/sphinx-labs/sphinx/pull/996/files#diff-437a9444caa71b5e7ac239ee371b3b7435d44bc0a1ff51c82af3a8932a5bfe17R627-R635) and [corresponding test](https://github.com/sphinx-labs/sphinx/blob/9e1dc5dabc029034192270370e67ec3e4e712d01/packages/plugins/test/Validation.spec.ts#L477)

PR #1007, opened 2023-09-04
feat(core): update diff to include constructor arg variables and post-deployment actions

sam-goldman:


packages/core/src/actions/bundle.ts
```diff
@@ -535,6 +536,12 @@ export const makeActionBundleFromConfig = (
     const { isTargetDeployed } = configCache.contractConfigCache[referenceName]
     const { kind, salt, constructorArgs } = contractConfig
 
+    const readableSignature = prettyFunctionCall(
+      referenceName,
+      'constructor',
+      constructorArgs
+    )
```
RPate97, line 543: I'm reverting this because the output is too complicated to render well on the website. I think if we want to include the constructor args, that's reasonable but we should do it in a way where we are only including the arguments for the chain where the deployment failed.

PR #1039, opened 2023-09-13
fix(core): Remove silent failure on external call revert or contract deployment failure

RPate97:


packages/plugins/contracts/test/SphinxManagerInherited.sol
sam-goldman, line 1: Is this file used anywhere?

packages/plugins/contracts/foundry/Sphinx.sol
```diff
@@ -448,21 +448,30 @@ abstract contract Sphinx {
                 manager.setStorage{ gas: bufferedGasLimit }(rawActions, _proofs);
             } else {
                 vm.recordLogs();
-                manager.executeInitialActions{ gas: bufferedGasLimit }(rawActions, _proofs);
+                // manager.executeInitialActions{ gas: bufferedGasLimit }(rawActions, _proofs);
+                (bool success, bytes memory result) = address(manager).call{ gas: bufferedGasLimit }(abi.encodeWithSignature("executeInitialActions((uint8,uint256,bytes)[],bytes32[][])", rawActions, _proofs));
```
sam-goldman, line 452: Why do we use `encodeWithSignature` here?

packages/plugins/test/PostDeploymentActions.spec.ts
```diff
@@ -643,7 +643,7 @@ describe('Post-Deployment Actions', () => {
     })
   })
 
-  const integrations: Array<Integration> = ['hardhat', 'foundry']
+  const integrations: Array<Integration> = ['hardhat']
```
sam-goldman, line 646: Is there a reason you removed 'foundry' here? Without this, the test suite won't be run against the foundry deploy task

sam-goldman (approved):

sam-goldman:
Looks like there are a couple places in the user-facing docs where the manager version is 0.2.3

sam-goldman:
I think you need to update the `prev-build-info.json` with the latest build info file

sam-goldman:
Looks like there's some on and off chain logic that uses `DeploymentStatus.FAILED`, which I think we can remove. We may not be able to remove this member from the DeploymentStatus type though because it could produce a storage layout compatibility error

packages/plugins/contracts/foundry/Sphinx.sol
RPate97: If we just call the function directly, it will revert with the custom error immediately. I switched this to a low-level call so that the script doesn't instantly revert and instead we can decode the action index and output a more helpful error message.

PR #1040, opened 2023-09-13
fix(core): cancel existing deployments via proposals

sam-goldman:


RPate97 (approved):
Gave it a quick look through and everything looks solid to me. I think this is fine to merge once the checks are passing. I'll be writing automated tests for this logic on the website as well, so will let you know if that uncovers any problems.

PR #1156, opened 2023-11-06
a

sam-goldman:


packages/contracts/contracts/core/ManagedService.sol
```diff
@@ -0,0 +1,117 @@
...
+     */
+    event Called(address indexed relayer, address indexed to, bytes32 data);
+
+    /**
+     * @notice A modifer that refunds the caller for the gas spent in the function call.
+     */
+    modifier refund {
+        uint256 start = gasleft();
+        _;
+        uint256 spent = start - gasleft() + 40000;
```
sam-goldman, line 42: How likely is it that at some point, the `ManagedService` won't have enough ETH in its balance to cover a gas amount equal to `start - gasleft() + 40000`? This'd cause the entire tx to revert, including the user's deployment, which seems potentially problematic

packages/contracts/contracts/core/ManagedService.sol
```diff
@@ -0,0 +1,117 @@
...
+     * @param relayer The address of the account that made the call.
+     * @param to      The address of the remote contract.
+     * @param data    A keccak256 hash of the input data.
+     */
+    event Called(address indexed relayer, address indexed to, bytes32 data);
+
+    /**
+     * @notice A modifer that refunds the caller for the gas spent in the function call.
+     */
+    modifier refund {
```
sam-goldman, line 39: nit: is it necessary to use a modifier, since we only attach it to a single function? Merging it into the `exec` function would be slightly cleaner imo because the `exec` function returns data (via `return res;`), but this doesn't actually exit the modifier. This is just a Solidity quirk, but it makes the flow of the function a little opaque

specs/managed-service.md
```diff
@@ -0,0 +1,65 @@
...
+Should execute a low-level call to the `_to` address with `_data`, and refund the caller for the cost of the transaction.
+
+- Must revert if the caller does not have the `RELAYER_ROLE` role.
+- Must revert if the underlying call reverts.
+- Must revert if the balance of the `ManagedService` contract is not sufficient to refund the caller for their gas cost.
+- A successful call:
+  - A call is made to the target address with the requested data.
+  - The cost of the transaction is refunded to the caller.
+  - Emit a `Called` event in the `ManagedService` contract.
+  - The return value of the underlying call is returned encoded as raw bytes.
```
sam-goldman, line 48: nit: Would be slightly easier to read this section if the sentence structures were consistent. E.g.
- A successful call must:
  - Call the target address with the requested data.
  - Refund the caller the cost of the transaction.
etc

Same applies for the corresponding section in the other function sections

specs/managed-service.md
```diff
@@ -0,0 +1,65 @@
...
+
+#### function exec(address _to, bytes calldata _data) public payable returns (bytes memory)
+Should execute a low-level call to the `_to` address with `_data`, and refund the caller for the cost of the transaction.
+
+- Must revert if the caller does not have the `RELAYER_ROLE` role.
+- Must revert if the underlying call reverts.
+- Must revert if the balance of the `ManagedService` contract is not sufficient to refund the caller for their gas cost.
+- A successful call:
+  - A call is made to the target address with the requested data.
+  - The cost of the transaction is refunded to the caller.
```
sam-goldman, line 46: nit: Might wanna mention that this includes a buffer

specs/managed-service.md
```diff
@@ -0,0 +1,65 @@
...
+#### function withdrawTo(uint256 _amount, address _recipient) public
+Should send `_amount` from the `ManagedService` contract and to the `_recipient` address.
+
+- Must revert if the caller does not have either the `RELAYER_ROLE` or the `DEFAULT_ADMIN_ROLE`.
+- Must revert if `_amount` is greater than the balance of the `ManagedService` contract.
+- Must revert if the `_recipient` is the zero address.
+- Must revert if the transfer fails for any reason.
+- A successful call:
+  - `_amount` in native tokens is transferred to the `_recipient` address.
+  - Emit a `Withdew` event in the `ManagedService` contract.
```
sam-goldman, line 59: `Withdew` -> `Withdrew`

specs/managed-service.md
```diff
@@ -0,0 +1,65 @@
...
+
+#### function withdraw(uint256 _amount) external
+Should send `_amount` to the caller. Invariants match the `withdrawTo` function.
+
+## Assumptions
+The `ManagedService` contract on the OpenZeppelin `AccessControl` contract. We test that the interactions with this contract work properly in the [unit tests for the `ManagedService`](TODO(end)). However, we don't thoroughly test the internals of this contract. Instead, we rely on the assumption that it is robust and secure.
```
sam-goldman, line 65: Looks like the first sentence is incomplete

sam-goldman, line 65: I updated this language in the SphinxModuleProxyFactory spec to be a little less choppy. Feel free to use if you want:

> We test that the interactions with these contracts work properly in the [unit tests for the `SphinxModuleProxyFactory`](TODO(end)), but we don't thoroughly test the internals of these external contracts. Instead, we rely on the assumption that they're secure and have been thoroughly tested by their authors.

sam-goldman, line 65: We may want to mention the fact that the owner of the `ManagedService` and the relayers can drain all of the funds in the `ManagedService`. We can just say that we can assume these accounts are secured properly.

specs/managed-service.md
sam-goldman, line 1: It may also be worth mentioning somewhere in the spec that if a relayer's private key is leaked, they cannot execute anything malicious through a user's `SphinxModuleProxy`, since it's only possible to execute transactions that have been explicitly approved by the user

packages/contracts/contracts/core/ManagedService.sol
sam-goldman, line 1: Did you run slither on the latest version of this contract? If not, you can just run `yarn slither` in the contracts package. I've been addressing everything that slither mentions, even green detectors. If a Slither warning isn't a concern, I've been documenting the reason and then using [one of these inline comments](https://github.com/crytic/slither/wiki/Usage#triage-mode) to remove the warning

sam-goldman:
Looks great! Just minor comments

packages/contracts/test/ManagedService.t.sol
sam-goldman, line 1: Does this have full coverage of the `ManagedService` contract? You can check w/ `forge coverage`

packages/contracts/contracts/core/ManagedService.sol
```diff
@@ -0,0 +1,117 @@
...
+    event Withdew(address indexed recipient, uint256 amount);
+
+    /**
+     * @notice Emitted when a call is made to a remote contract.
+     *
+     * @param relayer The address of the account that made the call.
+     * @param to      The address of the remote contract.
+     * @param data    A keccak256 hash of the input data.
+     */
+    event Called(address indexed relayer, address indexed to, bytes32 data);
```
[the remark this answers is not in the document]
RPate97: It's so we can search for the event (and transaction) that corresponds to a specific call based off the target address and input data.

packages/contracts/contracts/core/ManagedService.sol
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    /**
+     * @notice Allows for the relayers to make arbitrary calls using this contract. The relayers will
+     *         be automatically refunded for the gas spent in the call.
+     *
+     * @param _to   The target address.
+     * @param _data The data that will be sent.
+     */
+    function exec(address _to, bytes calldata _data) public payable refund returns (bytes memory) {
```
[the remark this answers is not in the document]
RPate97: I expect to use this contract for a variety of things on the website, not all of which will necessarily be related to doing actual deployments via the module. So I think it's a good idea to include it b/c I expect I may encounter cases where it is useful.

packages/contracts/contracts/core/ManagedService.sol
RPate97: If we run out of funds, then we won't be able to execute their deployment regardless of if we revert here or not. This just makes it an explicit requirement that the managed service must contain the necessary funds for each transaction instead of having it be an implicit requirement that the individual wallets must have funds.

specs/managed-service.md
RPate97: Yeah, we can mention that here, but I think that's more relevant to the SphinxModule spec. I.e we should discuss there what the purpose of the executor address is, and that it should not matter what/who that address is. They cannot do anything that isn't explicitly approved by the user.

packages/contracts/contracts/core/ManagedService.sol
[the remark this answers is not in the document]
RPate97: It's an estimate of the overhead of the refund modifier which is not included in the gas estimation happening here. I estimated it should be about 30k, and then tested it and found that 40k was a nice number to consistently refund a bit more than the actual cost. I was intentionally a little conservative with it.

packages/contracts/contracts/core/ManagedService.sol
RPate97: This is intentional and makes it easier to accurately calculate the gas cost b/c the cost of the return value is included in the estimate calculated in the modifier.  Additionally, the use of a modifier is common practice for this use case, so I expect auditors to be familiar with the pattern or at least be able to find resources online related to it.

packages/contracts/contracts/core/ManagedService.sol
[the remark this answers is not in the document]
sam-goldman: That makes sense, but it looks like the `dataHash` isn't `indexed`. Is it supposed to be?

specs/bundling.md
```diff
@@ -0,0 +1,27 @@
+## Safe Module Merkle Tree Bundling Logic
+
+### Goal
+A function that transforms an arbitrary set of transactions on an arbitrary set of networks into a Merkle Tree which is executable by the SphinxModule. See the SphinxModule Merkle Tree spec for more detail on the exact structure and content of the Merkle tree and its leafs.
```
sam-goldman, line 4: SphinxModule -> SphinxModuleProxy? Likewise for the other places in this spec where it's mentioned

specs/bundling.md
```diff
@@ -0,0 +1,27 @@
...
+
+### Secondary Goals
+- Simple and auditable
+- Should be a function that is agnostic to the source of transactions so it can be used to assemble a Merkle Tree based on transactions generated from any scripting framework (i.e Foundry, Hardhat Ignition, some arbitrary future framework)
```
sam-goldman, line 8: nit: Merkle Tree -> Merkle tree

specs/bundling.md
```diff
@@ -0,0 +1,27 @@
...
+
+### Input
+Accepts an object where the keys are canonical chain ids and the values are deployment data objects.
+
+#### Deployment Object Fields:
+- nonce: The current nonce of the Safe on the target network
```
sam-goldman, line 14: nit: `backticks` for variable names

sam-goldman, line 14: I think the `nonce` is supposed to be the nonce in the `SphinxModuleProxy`, not the Safe

specs/bundling.md
```diff
@@ -0,0 +1,27 @@
...
+- txs: An array of Sphinx Transaction objects
+
+#### Sphinx Transaction Object Fields:
+- to: The destination address for the transaction
+- value: The amount of native gas token to send from the Safe as part of the transaction
```
sam-goldman, line 19: It's implied, but it'd be nice to explicitly say that the value is being transferred to the destination address, i.e. the `to` field

specs/bundling.md
```diff
@@ -0,0 +1,27 @@
...
+- gas: The gas limit for the transaction.
```
sam-goldman, line 20: To be a little more specific, this is the amount of gas that's included in the call from the `SphinxModuleProxy` to the Gnosis Safe

specs/bundling.md
```diff
@@ -0,0 +1,27 @@
...
+- operation: The type of transaction operation. We should use a Typescript enum with the values `Call`(0) and `DelegateCall`(1) which correspond to the `Enum.Operation.Call` and `Enum.Operation.DelegateCall` enum values defined by Safe.
```
sam-goldman, line 21: nit: Typescript -> TypeScript

specs/bundling.md
```diff
@@ -0,0 +1,27 @@
...
+- data: Arbitrary calldata to forward to the safe.
+
+### Output
+A `StandardMerkleTree` from the `@openzeppelin/merkle-tree` library with leafs that match the SphinxModule Merkle tree spec. In addition to that, the `StandardMerkleTree` should contain leafs which are ordered by chain id ascending and index ascending starting with the approval leaf for each chain.
+
+So for example, if we intend to run some set of transactions on Goerli (5), OP Goerli (420), and Polygon Mumbai (80001). Then the leafs are expected to be ordered with the approval leaf for Goerli first, then all of the transaction leafs on Goerli in ascending order, then the approval leaf for OP Goerli, then all the transaction leafs on OP Goerli in ascending order, and so on.
```
sam-goldman, line 27: Example looks good, but it looks like there are a couple grammar errors in this paragraph

specs/bundling.md
```diff
@@ -0,0 +1,27 @@
...
+
+#### Sphinx Transaction Object Fields:
+- to: The destination address for the transaction
+- value: The amount of native gas token to send from the Safe as part of the transaction
+- gas: The gas limit for the transaction.
+- operation: The type of transaction operation. We should use a Typescript enum with the values `Call`(0) and `DelegateCall`(1) which correspond to the `Enum.Operation.Call` and `Enum.Operation.DelegateCall` enum values defined by Safe.
+- data: Arbitrary calldata to forward to the safe.
+
+### Output
+A `StandardMerkleTree` from the `@openzeppelin/merkle-tree` library with leafs that match the SphinxModule Merkle tree spec. In addition to that, the `StandardMerkleTree` should contain leafs which are ordered by chain id ascending and index ascending starting with the approval leaf for each chain.
```
sam-goldman, line 25: nit: chain id -> chain ID

specs/bundling.md
```diff
@@ -0,0 +1,27 @@
...
+- Should be a function that is agnostic to the source of transactions so it can be used to assemble a Merkle Tree based on transactions generated from any scripting framework (i.e Foundry, Hardhat Ignition, some arbitrary future framework)
+
+### Input
+Accepts an object where the keys are canonical chain ids and the values are deployment data objects.
+
+#### Deployment Object Fields:
+- nonce: The current nonce of the Safe on the target network
+- txs: An array of Sphinx Transaction objects
+
+#### Sphinx Transaction Object Fields:
```
sam-goldman, line 17: Would be helpful to have the actual TypeScript names of these data types. Another option is to just document these variable types in TypeScript then link to them in a 'Relevant Files' section like we do in the other specs. That may be preferable b/c we'll probably want to document the TypeScript logic, and this way we don't need to duplicate it

specs/bundling.md
sam-goldman, line 25: > ordered by chain id ascending and index ascending starting with the approval leaf for each chain.

This makes it sound like the order should be:
1. Approval leaf, chain ID X
2. Approval leaf, chain ID Y
3. Execute leafs

but the example in the next paragraph indicates otherwise. It may be clearer to just explain with an example.

sam-goldman, line 25: > ordered by... index ascending

We don't explicitly mention the index anywhere in this spec except here, which is a little confusing

specs/bundling.md
sam-goldman, line 1: I don't mention the term 'bundle'/'bundling' anywhere in the other specs/docs. Instead of 'bundling', i've been saying 'encoding' since imo it's slightly clearer. Also, instead of referring to a Merkle root + leafs + proofs as a 'bundle', I've just been calling it a Merkle tree, since imo it's not necessary to introduce a new term. Not opinionated but we should probably just be consistent

sam-goldman, line 1: Would be great to include a "Relevant Files" section like the other specs

packages/contracts/src/module.ts
sam-goldman, line 1: Would be great if we could unit test this file. My tests thoroughly cover the single chain case (including when there's just a single `APPROVE` leaf), but I don't test it with multiple chains

specs/introduction.md
```diff
@@ -0,0 +1,57 @@
+# Introduction
+
+Sphinx is a protocol that aims to make the smart contract deployment process more secure, transparent, and efficient. This document provides an overview of Sphinx to serve as context for the rest of the specification.
+
+## Goals
+
+* **Secure**: Deployments with Sphinx are tamper-proof. This is crucial because of the mission-critical nature of smart contract deployments, where subtle changes in a deployment can cause significant issues after a system has been deployed.
```
RPate97, line 7: It's unclear what tamper proof actually means in this context. I think we need to give more detail on what the tool actually does to make it clear what the value of this is exactly. I.e we should explain that the tool enables deployments to be approved off-chain and then executed by an arbitrary third party. In that context, being tamper-proof is very important.

I think it's reasonable to include this context either in the main readme or in this document. I would probably do a combination of both.

specs/introduction.md
```diff
@@ -0,0 +1,57 @@
...
+* **Transparent**: Sphinx offers greater transparency by providing a verifiable hash of each deployment. Users and other stakeholders can verify the transactions in a deployment before it's executed.
+* **Efficient**: Developers can approve arbitrarily large deployments across an arbitrary number of chains by signing the deployment's unique identifier using a meta transaction. Once it's approved, Sphinx's backend will trustlessly execute the deployment on every chain. Developers don't need native gas tokens on any chain.
+
+## Architecture Overview
+
+Most teams use a multisig wallet to approve permissioned actions. We wanted teams to be able to approve deployments from their multisig as well, so we built our protocol on top of Gnosis Safe. Our primary on-chain component is a [Safe Module](TODO(end)), which allows us to use our own signature verification mechanism that's designed for deployments (details below).
```
RPate97, line 13: I think the ordering of this section makes it difficult to understand. I also think it could be cleanly split into two sections. I would start with a high level `Architecture Overview` section, then followup with a `Merkle Tree Architecture`.

In the `Architecture Overview`, I would just explain at a high level how we intend to accomplish the stated goals. So something like: 

"To achieve the above goals Sphinx implements a custom transaction verification scheme that allows a set of multisig owners to approve a deployment including an arbitrarily large number of transactions across an arbitrary set of networks using an off chain meta transaction. Once a deployment is approved, it can be executed trustlessly by an external third party."

Then in the same section, I we should explain that we use a Merkle tree as an efficient data structure for storing all of the transactions. Finally, we should end this section by explaining that we implement this verification scheme as a module installed in a Gnosis Safe since that is the industry standard for multi-signature wallets. 

I'd then have a second section titled `Merkle Tree` which focuses specifically on the structure of the Merkle Tree. I think the content you have here related to the Merkle Tree is solid.

RPate97 (commented):
Reviewed `introduction.md`

specs/introduction.md
```diff
@@ -0,0 +1,47 @@
+# Introduction
+
+Sphinx is a protocol that aims to make the smart contract deployment process more secure, transparent, and efficient. This document provides an overview of Sphinx to serve as context for the rest of the specification.
```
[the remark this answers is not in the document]
sam-goldman: Sounds good, I'll make a note to update the main README soon then we can see if we need additional context here

specs/introduction.md
sam-goldman: Will also come back to this after updating the main README

RPate97 (commented):
reviewed `sphinx-module-proxy.md` and also left a minor comment on `SphinxModule.sol`

specs/sphinx-module-proxy.md
```diff
@@ -0,0 +1,252 @@
...
+
+The `data` field of an `EXECUTE` leaf primary contains data to forward to the Gnosis Safe. It consists of the following fields, which are ABI encoded:
+* `address to`: The target address of the transaction. This is _not_ the address of the Gnosis Safe.
+* `uint256 value`: The amount of native gas token to transfer from the Gnosis Safe to the target address. This value is not transferred from the `SphinxModuleProxy`.
+* `uint256 gas`: The amount of gas that's included in the call from the `SphinxModuleProxy` to the Gnosis Safe.
+* `bytes txData`: The transaction's data.
+* `Enum.Operation operation`: The type of transaction to execute in the Gnosis Safe, i.e. `Call` or `DelegateCall`.
+* <span id="requireSuccess">`bool requireSuccess`: If this is `true` and the transaction in the Gnosis Safe fails, the deployment be marked as "failed" and will end immediately. If this is `false`, the deployment will continue regardless of whether the transaction fails.</span>
```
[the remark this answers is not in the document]
sam-goldman: oops

specs/sphinx-module-proxy.md
```diff
@@ -0,0 +1,258 @@
...
+
+### `EXECUTE` Leaf Data
+
+The `data` field of an `EXECUTE` leaf primarily contains data to forward to the Gnosis Safe. It consists of the following fields, which are ABI encoded:
+* `address to`: The target address of the transaction. This is _not_ the address of the Gnosis Safe.
+* `uint256 value`: The amount of native gas token to transfer from the Gnosis Safe to the target address. This value is not transferred from the `SphinxModuleProxy`.
+* `uint256 gas`: The amount of gas that's included in the call from the `SphinxModuleProxy` to the Gnosis Safe.
+* `bytes txData`: The transaction's data.
+* `Enum.Operation operation`: The type of transaction to execute in the Gnosis Safe, i.e. `Call` or `DelegateCall`.
+* `bool requireSuccess`: If this is `true` and the transaction in the Gnosis Safe fails, the deployment be marked as "failed" and will end immediately. If this is `false`, the deployment will continue regardless of whether the transaction fails.
```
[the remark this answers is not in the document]
sam-goldman: I don't think we should talk about the `DeploymentStatus` here b/c it's a lower level data structure that we don't mention until the function-level invariants. The answer to your question is in [this part of the spec](https://github.com/sphinx-labs/sphinx/pull/1156/files/9d10b6a8a9e4992c32aaf6226a96917944e26426#diff-d649d57297a3839c70a33dd38650cc6155217c41ad9f8296bb1740bac6accfd8R189-R193), which says we only set the deployment as `FAILED` if a txn fails _and_ the txn requires a success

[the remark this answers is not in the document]
sam-goldman: We haven't mentioned the `DeploymentStatus` at this point in the spec, so I don't think we should describe this here. I updated the next section, `Deployment Process`, to describe the exact situation where a deployment is marked as `FAILED`:

> 4. **Failed**: A Merkle root will fail if one of the transactions reverts in the Gnosis Safe _and_ the transaction must succeed (i.e. `requireSuccess == true`). Failed Merkle roots can never be re-approved or executed.

IMO this makes it clear that if `requireSuccess == false`, the deployment won't be marked as failed. Does that clear it up for you?

specs/sphinx-module-proxy.md
```diff
@@ -0,0 +1,258 @@
...
+    E -->|Yes| F[Failed]
+    E -->|No| G[Are there any more leaves to execute for the current chain in this deployment?]
+    G -->|Yes| D
+    G -->|No| B
+```
+
+## High-Level Invariants
+
+- There must be at most one active Merkle root in a `SphinxModuleProxy` at a time.
+- A Merkle root must contain at most one deployment per chain.
```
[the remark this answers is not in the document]
sam-goldman: Just going to remove this bullet point b/c I think it's covered by this invariant in the `approve` function:
> - Must revert if the Merkle root has ever been approved before.

specs/merkle-tree.md
```diff
@@ -0,0 +1,158 @@
...
+    E[Hash] --> H["Approve: Chain ID: 5, Index: 0"]
+    F[Hash] --> M["Execute: Chain ID: 420, Index: 3, Data: 0xdd..."]
+    F[Hash] --> N["Execute: Chain ID: 420, Index: 2, Data: 0xee..."]
+    G[Hash] --> O["Execute: Chain ID: 420, Index: 1, Data: 0xff..."]
+    G[Hash] --> L["Approve: Chain ID: 420, Index: 0"]
+```
+
+Sphinx Merkle trees should follow a consistent ordering scheme to ensure they can be executed by a variety of executor implementations without issue. We impose the standard that the leaves should be ordered by index and chain ID ascending.
+
+So for example, if we intended to run some set of transactions on Goerli (5), OP Goerli (420), and Polygon Mumbai (80001). Then the leaves are expected to be ordered with the approval leaf for Goerli first, then all of the transaction leaves on Goerli in ascending order, then the approval leaf for OP Goerli, all the transaction leaves on OP Goerli in ascending order, and so on.
```
sam-goldman, line 125: It's implied here, but I think we should make it explicit that when we refer to the order of leafs, we're talking about the position of the leafs in the Merkle tree

sam-goldman, line 125: It'd be nice to list out the leafs in this example. E.g:
Position 0: Approval leaf on Goerli (chain ID 5)
Position 1: ...

specs/merkle-tree.md
```diff
@@ -0,0 +1,158 @@
...
+
+### 3. Every chain ID and index combination must be unique within the tree
```
sam-goldman, line 127: chain ID and index -> `chainId` and `index` field

specs/merkle-tree.md
```diff
@@ -0,0 +1,158 @@
+# Sphinx Merkle Tree
+Sphinx uses a custom Merkle tree data structure to allow teams to approve arbitrarily large deployments across an arbitrary number of networks via off-chain signatures, and then allow those transactions to be trustlessly executed via their Gnosis Safe by a third party relayer. This document outlines the architecture and content of the Sphinx Merkle tree.
+
+**Vocabulary notes**:
```
sam-goldman, line 4: I think we should add a vocabulary term for the word _index_. We should make it clear that this refers to a field in the Merkle leaf, and _not_ the Merkle leaf's position in the Merkle tree. I think it'd be appropriate to add a link or a footnote explaining why we do this.

specs/merkle-tree.md
```diff
@@ -0,0 +1,158 @@
...
+
+So for example, if we intended to run some set of transactions on Goerli (5), OP Goerli (420), and Polygon Mumbai (80001). Then the leaves are expected to be ordered with the approval leaf for Goerli first, then all of the transaction leaves on Goerli in ascending order, then the approval leaf for OP Goerli, all the transaction leaves on OP Goerli in ascending order, and so on.
+
+### 3. Every chain ID and index combination must be unique within the tree
+Since the Sphinx Merkle tree leaves contain explicit indexes, it is also possible to construct trees that contain multiple leaves with the same index value and chain ID. However, only one of the leaves would be executable on-chain. Therefore, constructing a tree with multiple leaves with the same chain ID and index would make it ambiguous to the relayer which leaf should be executed, and would result in a tree that cannot be fully executed. So we impose the standard that in a given Sphinx Merkle tree there must be at most one leaf with each chain ID and index combination.
```
sam-goldman, line 128: > it is also possible to construct trees that contain multiple leaves with the same index value and chain ID. 

This is confusing b/c it's not possible under this spec

sam-goldman, line 128: I think this paragraph can be condensed into a single sentence. I think it's enough rationale to just say that this ensures there is an unambiguous ordering of leaves on each chain

specs/merkle-tree.md
```diff
@@ -0,0 +1,158 @@
...
+
+### 4. There must be exactly one approval leaf or cancel leaf per chain
+Since Sphinx Merkle trees can contain an arbitrary set of leaves, it is possible to construct a single tree that contains `APPROVE` or `CANCEL` leaves that approve or cancel multiple separate deployments on a single chain. However, this leads to an ambiguous situation for the relayer where it is unclear which deployment should be approved or canceled.
```
sam-goldman, line 131: Same feedback as the last section: it's confusing to say "it's possible to construct..." in this section because it's not supposed to be possible under this spec

specs/merkle-tree.md
```diff
@@ -0,0 +1,158 @@
...
+
+If the `arbitraryChain` field is set to true, then we consider this approval leaf to apply to all chains. Therefore if `arbitraryChain` = true, then there should be exactly one `APPROVE` leaf in the entire tree and no `CANCEL` leaves.
+
+## Sphinx Merkle Tree Generation Standard Implementation
+We provide a standard implementation for generating Sphinx Merkle trees that adhere to the above invariants. This implementation is used by the official Sphinx Foundry plugin.
+
+### Input
+Accepts an object where the keys are canonical chain ids and the values are deployment data objects which contain all of the necessary info to assemble a SphinxMerkleTree. See the [inline documentation](TODO(end)) for exact details on the input data. The Sphinx Merkle tree generation function should be agnostic to the source of the transaction data so it can be used to assemble a tree based on transactions generated from any scripting framework (i.e Foundry, Hardhat Ignition, some arbitrary future framework).
```
sam-goldman, line 141: > Accepts an object where the keys are canonical chain ids and the values are deployment data objects which contain all of the necessary info to assemble a SphinxMerkleTree.

I think we should explicitly mention and link to the `DeploymentData` object. Something like "Accepts a [`DeploymentData`](TODO(end)) object, which maps canonical chain IDs to..."

specs/merkle-tree.md
```diff
@@ -0,0 +1,158 @@
...
+
+### Output
+A Sphinx Merkle tree that follows the above specification and invariants. See the [inline documetation](TODO(end)) for the exact details of the output data structure.
```
sam-goldman, line 144: I think we should explicitly mention and link to the `SphinxMerkleTree` object. Something like "Accepts a [`SphinxMerkleTree`](https://github.com/sphinx-labs/sphinx/pull/1156/TODO(end)) object"

specs/merkle-tree.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Function-Level Invariants
+We export three separate functions `makeSphinxLeaves`, `makeSphinxMerkleTreeFromLeaves`, and `makeSphinxMerkleTree`. We provide three separate functions for convenience and modularity to enable testing. However, for this specification and the below invariants we assume the use of the `makeSphinxMerkleTree`. We *only* recommend using that function when generating Sphinx Merkle trees in production and therefore only consider it to satisfy the invariants listed above.
```
sam-goldman, line 147: > we assume the use of the `makeSphinxMerkleTree`

I think you mean that we just cover this function in the spec, but it's not totally clear

sam-goldman, line 147: > and therefore only consider it to satisfy the invariants listed above.

it's not totally clear what this means

sam-goldman, line 147: Based on this paragraph, it seems these specs only concern `makeSphinxMerkleTree`. We should include its function signature like we do for the other specs

specs/merkle-tree.md
```diff
@@ -0,0 +1,158 @@
...
+
+### It is impossible to specify multiple `NetworkDeploymentData` objects for a given network
+The Merkle tree generation logic relies on the fact that it is not possible to submit multiple `NetworkDeploymentData` objects to guarantee that there is exactly one `APPROVE` or `CANCEL` leaf ([4](#4-there-must-be-exactly-one-approval-leaf-or-cancel-leaf-per-chain-4)).
```
sam-goldman, line 150: I think this can be shortened to: This guarantees that there is exactly one `APPROVE` or `CANCEL` leaf ([4](#4-there-must-be-exactly-one-approval-leaf-or-cancel-leaf-per-chain-4)).

specs/merkle-tree.md
```diff
@@ -0,0 +1,158 @@
...
+
+### The index of all leaves is generated internally and cannot be specified by the input data
+The Merkle tree generation logic relies on calculating the indexes for all leaves internally to guarantee that the leaves are ordered properly ([2](#2-merkle-tree-leaves-must-be-ordered-in-the-tree-by-index-and-chain-id-ascending)) and that there are no index gaps ([1](#1-there-must-be-no-gaps-in-index-in-leaves-that-have-the-same-chain-id)).
```
sam-goldman, line 153: The fact that we generate the indexes internally seems like less of an invariant and more of an implementation detail. Consider rephrasing this to focus on things we can unit test, like leaves that are ordered properly and have no index gaps

specs/merkle-tree.md
```diff
@@ -0,0 +1,158 @@
...
+
+This invariant works together with the above invariant, it is not possible to specify multiple `NetworkDeploymentData` objects for a given network, to ensure each index + chain ID combination is unique ([3](#3-every-chain-id-and-index-combination-must-be-unique-within-the-tree-3)).
+
+## Dependencies
+The Sphinx standard Merkle tree generation logic makes calls to two popular external libraries. We use `ethers` version 6.7.0 to handle ABI encoding tree leaf `data`, and we use `@openzeppelin/merkle-tree` to assemble the Merkle tree. We test that interactions with these libraries work properly in our higher-level unit tests which ensure that Merkle trees generated by this logic have data that is encoded properly and that they are executable on-chain. However, we do not test the internals of these libraries and instead rely on the assumption that they are robust and bug-free.
```
sam-goldman, line 158: popular external libraries -> battle-tested external libraries

sam-goldman, line 158: > our higher-level unit tests which ensure that... they are executable on-chain.

It seems like you're referring to the Merkle tree unit tests, which don't check that the tree is executable on-chain. I.e. the `SphinxModuleProxy` can revert for any number of reasons, which isn't checked by the Merkle tree tests

specs/merkle-tree.md
```diff
@@ -0,0 +1,158 @@
...
+
+### It is impossible to specify multiple `NetworkDeploymentData` objects for a given network
+The Merkle tree generation logic relies on the fact that it is not possible to submit multiple `NetworkDeploymentData` objects to guarantee that there is exactly one `APPROVE` or `CANCEL` leaf ([4](#4-there-must-be-exactly-one-approval-leaf-or-cancel-leaf-per-chain-4)).
+
+### The index of all leaves is generated internally and cannot be specified by the input data
+The Merkle tree generation logic relies on calculating the indexes for all leaves internally to guarantee that the leaves are ordered properly ([2](#2-merkle-tree-leaves-must-be-ordered-in-the-tree-by-index-and-chain-id-ascending)) and that there are no index gaps ([1](#1-there-must-be-no-gaps-in-index-in-leaves-that-have-the-same-chain-id)).
+
+This invariant works together with the above invariant, it is not possible to specify multiple `NetworkDeploymentData` objects for a given network, to ensure each index + chain ID combination is unique ([3](#3-every-chain-id-and-index-combination-must-be-unique-within-the-tree-3)).
```
sam-goldman, line 155: It should be clear that "it is not possible to specify multiple `NetworkDeploymentData` objects for a given network" is a section header. I'd remove this phrase and link to do something like: "This invariant works together with the [`above invariant`](TODO(end)). (Notice that we add a hyperlink to the phrase "above invariant")

sam-goldman, line 155: > index + chain ID

I think we should avoid informal language like this

specs/merkle-tree.md
```diff
@@ -0,0 +1,158 @@
+# Sphinx Merkle Tree
+Sphinx uses a custom Merkle tree data structure to allow teams to approve arbitrarily large deployments across an arbitrary number of networks via off-chain signatures, and then allow those transactions to be trustlessly executed via their Gnosis Safe by a third party relayer. This document outlines the architecture and content of the Sphinx Merkle tree.
+
+**Vocabulary notes**:
+- We use the term `relayer` to refer to the entity that submits the transactions to execute a Sphinx Merkle tree on-chain via a Sphinx Module proxy.
```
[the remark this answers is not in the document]
RPate97: No, the relayer explicitly isn't necessarily a specific account it's the off-chain system that actually executes the deployment. That's what I was trying to get across with these definitions. I will update the relayer definition to make this clearer as I agree "entity" is not clear enough.

specs/merkle-tree.md
```diff
@@ -0,0 +1,158 @@
...
+- We use the term `executor` to refer to the exact account that needs to submit the transactions to the `SphinxModuleProxy` to complete a deployment on-chain.
+
+### Relevant Files
+- Standard Merkle Tree Generation Logic: [`merkle-tree.ts`](TODO(end))
```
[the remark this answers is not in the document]
RPate97: It's a merkle tree that conforms the the invariants listed below, I will add a definition for it.

specs/merkle-tree.md
RPate97: The purpose of this paragraph is to explain _why_ the invariant exists by explaining what you could do if it did not and why that would cause issues. This is the main reason why I include so much detail on the possible Merkle trees you could construct which are not valid. I am trying to preemptively answer the auditor's questions regarding potential situations where the Merkle tree may be ambiguous, but still executable.

This was one of Kelvin's main pieces of feedback on the specs in general. He said they do a good job of stating exactly what the code must do, but they do not explain _why_. This prevents the reader from really understanding the purpose of the invariants and stops them from being able to offer higher-level feedback on the overall design of the system. Focusing only on exactly what the system does and not explaining the why forces them to focus only on the low-level details of the specific invariants. I think it would be a mistake to remove this context and I would encourage you to consider adding more reasoning to the invariants in the specs to make sure auditors understand the purpose of the invariants.

specs/merkle-tree.md
RPate97: See https://github.com/sphinx-labs/sphinx/pull/1156#discussion_r1400142204

specs/merkle-tree.md
RPate97: I'm referring to `SphinxModuleProxy.t.sol` which uses this function to generate Merkle trees and tests out executing them.

specs/merkle-tree.md
```diff
@@ -0,0 +1,142 @@
...
+* `Enum.Operation operation`: The type of transaction to execute in the Gnosis Safe, i.e. `Call` or `DelegateCall`.
+* `bool requireSuccess`: If this is `true` and the transaction in the Gnosis Safe fails, the deployment is marked as "failed" and will end immediately. If this is `false`, the deployment will continue regardless of whether the transaction fails.
+
+### `CANCEL` Leaf Data
+TODO
+
+## Valid Merkle Trees
+There are a wide variety of trees that conform to the basic structure described above, but not all of them will necessarily be executable on-chain for all networks they contain leaves for. For example, it is possible to include two leaves with the same `index` and `chainId`. In that case, both leaves would be valid but only one would be executable.
+
+To resolve these types of problems we outline a set of Merkle tree invariants below that define what a 'valid' Merkle tree looks like. We expect relayers to rely on Merkle trees following the invariants set out here. We assume that relayers will refuse to execute Merkle trees that do not follow these invariants.
```
sam-goldman, line 92: We already discussed this on the call, so here's a brief sketch of how I think we could rework this aspect of the spec. I think we can remove this section, and instead include a high-level rationale at the beginning of the "High-Level Invariants" section. Something like: "The purpose of these invariants is to define an unambiguous, consistent structure for Sphinx Merkle trees. This ensures that it's easy for off-chain tooling to recreate the exact Merkle tree signed by the user. It's worth mentioning that it's possible to construct Merkle trees that do not adhere to this spec, but are executable on-chain. We discuss this possibility in more detail in the [Assumptions section](TODO(end))."

Then, the assumptions section can explain these types of Merkle trees in detail

specs/merkle-tree.md
```diff
@@ -0,0 +1,142 @@
...
+
+## High-Level Merkle Tree Invariants
+
+### 1. Must be executable on all chains for which there is at least one Merkle leaf.
+Assumption: The input Merkle leaf(s) data satisfies the on-chain conditions in the corresponding SphinxModuleProxy contract(s). See the [SphinxModuleProxy specification](https://github.com/sphinx-labs/sphinx/blob/feature/pre-audit/specs/sphinx-module-proxy.md#function-level-invariants) for more information.
+
+### 2. Each leaf on a chain must have an index field with a value that sequentially increments by 1
+To ensure that `EXECUTE` leaves are executed in the correct order, all Merkle tree leaves must have an `index` field that starts at 0 on each chain and sequentially increments by 1.
```
sam-goldman, line 100: I think we should explicitly mention a few more things somewhere in these invariants:
- `APPROVE` and `CANCEL` leaves must always have an index of `0`
- `EXECUTE` leaves must begin with an index of `1`
- `EXECUTE` leaves must always follow an `APPROVE` leaf (never a `CANCEL` leaf)

This is a non-exhaustive list.  There may be other aspects of the TypeScript logic that deserve to be formalized in this spec. I think it's worth spending some time reviewing the TypeScript logic to make sure we aren't excluding important details

specs/merkle-tree.md
```diff
@@ -0,0 +1,142 @@
...
+
+### 3. Every leaf `chainId` and `index` combination must be unique within the tree
+Constructing a tree with multiple leaves with the same chain ID and index would make it ambiguous which leaf should be executed, and would result in a tree that cannot be fully executed. So we impose the restriction that in a valid Merkle tree there must be at most one leaf with each chain ID and index combination which ensures there is unambiguous ordering of leaves within the tree.
+
+### 4. There must be exactly one `APPROVE` leaf or `CANCEL` leaf per chain
+Constructing a tree that contains `APPROVE` or `CANCEL` leaves that approve or cancel multiple separate deployments on a single chain would lead to an ambiguous situation for the relayer where it is unclear which deployment should be approved or canceled. So we impose the restriction that in a valid Merkle tree there must be exactly one `APPROVE` or `CANCEL` leaf per chain.
+
+If the `arbitraryChain` field is set to true, then we consider this approval leaf to apply to all chains. Therefore if `arbitraryChain` = true, there should be exactly one `APPROVE` leaf in the entire tree and no `CANCEL` leaves.
```
sam-goldman, line 108: Auditors will check that this is upheld in the TypeScript logic. Do we enforce this?

specs/merkle-tree.md
```diff
@@ -0,0 +1,142 @@
...
+
+### 5. Merkle tree leaves must be ordered in the tree by the leaf's `index` and `chainId` fields ascending [^2]
+Sphinx Merkle tree leaves use an explicit `index` field in the leaf rather than an implicit index determined by the leaves position in the tree. This is because the Sphinx Merkle tree is intended to be executed across multiple networks and not all leaves will be executed on all networks. We impose the restriction the tree leaves will be ordered by their `index` and `chainId` fields ascending.
+
+So for example, if we intended to run a single transaction on Goerli (5), OP Goerli (420), and Polygon Mumbai (80001). Then the leaves are are expected to be ordered like so:
```
sam-goldman, line 113: What do you mean by a "single transaction"? Do you mean a single `EXECUTE` leaf?

specs/merkle-tree.md
```diff
@@ -0,0 +1,142 @@
...
+1. Approval Leaf on Goerli (Chain ID 5)
+2. Execute Leaf on Goerli
+3. Approval Leaf on OP Goerli (Chain ID 420)
+4. Execute Leaf on OP Goerli
+5. Approval Leaf on Polyon Mumbai (Chain ID 80001)
+6. Execute Leaf on Polygon Mumbai
```
sam-goldman, line 119: For clarity and consistency with other specs, I think we should call these `APPROVE` and `EXECUTE` leaves, including backticks

specs/merkle-tree.md
```diff
@@ -0,0 +1,142 @@
...
+## Merkle Tree Generation Logic
+We provide utility for generating Sphinx Merkle trees. This utility is used by the official Sphinx Foundry plugin. We expect it to generate Merkle trees that adhere to the above invariants.
+
+#### `const makeSphinxMerkleTree = (deploymentData: DeploymentData): SphinxMerkleTree`
+
+#### Input
+Accepts a [DeploymentData](https://github.com/sphinx-labs/sphinx/blob/591ff8a33578ecf61f184f4afc83a551397009f9/packages/contracts/src/merkle-tree.ts#L45C23-L45C23) object where the keys are canonical chain ids and the values are deployment data objects which contain all of the necessary info to assemble a SphinxMerkleTree. The Sphinx Merkle tree generation function should be agnostic to the source of the transaction data so it can be used to assemble a tree based on transactions generated from any scripting framework (i.e Foundry, Hardhat Ignition, some arbitrary future framework).
+
+#### Output
+Outputs a [SphinxMerkleTree](https://github.com/sphinx-labs/sphinx/blob/591ff8a33578ecf61f184f4afc83a551397009f9/packages/contracts/src/merkle-tree.ts#L115) object that follows the above specification and invariants.
```
sam-goldman, line 130: > above specification and invariants

I think this should be a bit more precise. Unless there are specific parts of the spec that you're referring to, I think we should just say "invariants"

specs/merkle-tree.md
```diff
@@ -0,0 +1,142 @@
...
+
+## Dependencies
+The Merkle tree generation logic makes calls to two battle-tested external libraries. We use `ethers` version 6.7.0 to handle ABI encoding tree leaf `data`, and we use `@openzeppelin/merkle-tree` version 1.0.5 to assemble the Merkle tree. We test that interactions with these libraries work properly in our [SphinxModuleProxy unit tests](https://github.com/sphinx-labs/sphinx/blob/feature/pre-audit/packages/contracts/test/SphinxModuleProxy.t.sol) which cover that Merkle trees generated by this logic have data that is encoded properly and that they are executable on-chain. However, we do not test the internals of these libraries and instead rely on the assumption that they are robust and bug-free.
```
sam-goldman, line 133: > The Merkle tree generation logic makes calls to two battle-tested external libraries.

I know I recommended the word 'battle-tested', but this phrasing is a bit awkward. I think it's fine to just remove this word b/c we say we assume they're robust and bug-free later

specs/merkle-tree.md
```diff
@@ -0,0 +1,142 @@
...
+
+## Assumptions
+The Merkle tree generation logic outputs trees which are valid and executable. However, there is no foolproof requirement that it be used. We make the asumption that relayers like the Sphinx DevOps platform will refuse to execute Merkle trees that are not valid according to the invariants defined in this document.
```
sam-goldman, line 136: > The Merkle tree generation logic outputs trees which are valid and executable.

Saying they're "valid" is a bit redundant because, by definition, a valid Merkle tree adheres to this spec (according to the vocabulary section). Saying they're "executable" is a bit confusing because there may be on-chain conditions that prevent the Merkle tree from being executed on-chain, which is an assumption that we discuss shortly after this

sam-goldman, line 136: > However, there is no foolproof requirement that it be used.

This sentence is unclear. Used by who? I think it's fine to remove this sentence because the next one does a good job of describing the assumption that we're making

specs/merkle-tree.md
```diff
@@ -0,0 +1,142 @@
...
+
+## Valid Merkle Trees
+There are a wide variety of trees that conform to the basic structure described above, but not all of them will necessarily be executable on-chain for all networks they contain leaves for. For example, it is possible to include two leaves with the same `index` and `chainId`. In that case, both leaves would be valid but only one would be executable.
+
+To resolve these types of problems we outline a set of Merkle tree invariants below that define what a 'valid' Merkle tree looks like. We expect relayers to rely on Merkle trees following the invariants set out here. We assume that relayers will refuse to execute Merkle trees that do not follow these invariants.
+
+## High-Level Merkle Tree Invariants
+
+### 1. Must be executable on all chains for which there is at least one Merkle leaf.
+Assumption: The input Merkle leaf(s) data satisfies the on-chain conditions in the corresponding SphinxModuleProxy contract(s). See the [SphinxModuleProxy specification](https://github.com/sphinx-labs/sphinx/blob/feature/pre-audit/specs/sphinx-module-proxy.md#function-level-invariants) for more information.
```
sam-goldman, line 97: I think we should link to the assumptions section, where we discuss this in slightly more detail, instead of the SphinxModuleProxy spec

specs/merkle-tree.md
```diff
@@ -0,0 +1,142 @@
...
+
+## Assumptions
+The Merkle tree generation logic outputs trees which are valid and executable. However, there is no foolproof requirement that it be used. We make the asumption that relayers like the Sphinx DevOps platform will refuse to execute Merkle trees that are not valid according to the invariants defined in this document.
+
+Merkle trees generated with this logic are only valid to the extent that the data used to generate them is valid. We make the assumption that the input data satisfies the [on-chain conditions in the corresponding SphinxModuleProxy contract(s)](https://github.com/sphinx-labs/sphinx/blob/feature/pre-audit/specs/sphinx-module-proxy.md#function-level-invariants).
```
sam-goldman, line 138: It'd be nice to provide a brief example to illustrate what this means. E.g. A leaf must be executed on its proper chain, or else the `SphinxModuleProxy` will revert

specs/merkle-tree.md
```diff
@@ -0,0 +1,142 @@
...
+
+## Footnotes
+[^1]: Because the Sphinx Merkle tree contains leaves that are intended to be executed across multiple networks and not all leaves will be executed on all networks, we cannot rely on the position of the leaves within the tree to determine the leaf index. Instead, we use an explicit `index` field on each leaf.
+[^2]: Because Sphinx Merkle tree leaves contain explicit `index` fields, it is technically possible to assemble Merkle trees that have leaves which are ordered in a variety of ways (such as reverse order or simply random). We impose the ordering restriction to help make it easier to understand the content of the Merkle tree. It also benefits offchain relayer infrastructure.
```
sam-goldman, line 142: > it is technically possible to assemble Merkle trees that have leaves which are ordered in a variety of ways (such as reverse order or simply random)

An example of this would be great

specs/merkle-tree.md
```diff
@@ -0,0 +1,166 @@
...
+* `uint256 leafMerkleRootNonce`: See [`APPROVE` leaf data](#approve-leaf-data).
+* `bytes32 merkleRootToCancel`: The Merkle root of the deployment that should be canceled.
+* `address executor`: See [`APPROVE` leaf data](#approve-leaf-data).
+* `string memory uri`: See [`APPROVE` leaf data](#approve-leaf-data).
+
+
+## High-Level Merkle Tree Invariants
+The purpose of these invariants is to define an unambiguous, consistent structure for Sphinx Merkle trees. This ensures that it's easy for off-chain tooling to recreate the exact Merkle tree signed by the user, and to check that a given Merkle tree is valid. It's worth mentioning that it's possible to construct Merkle trees that do not adhere to this spec, but are executable on-chain. We discuss this possibility and the expected executor behavior in more detail in the [Assumptions section](#assumptions).
+
+### 1. Must be executable on all chains for which there is at least one Merkle leaf.
```
sam-goldman, line 101: The auditors would probably find it helpful if we indicate how we test that this is true, since it's not tested in `merkle-tree.ts`

specs/merkle-tree.md
```diff
@@ -0,0 +1,166 @@
...
+
+### `CANCEL` Leaf Data
+
+The `data` field of an `CANCEL` leaf consists of the following fields, which are ABI encoded:
+* `address leafSafeProxy`: See [`APPROVE` leaf data](#approve-leaf-data).
+* `address moduleProxy`: See [`APPROVE` leaf data](#approve-leaf-data).
+* `uint256 leafMerkleRootNonce`: See [`APPROVE` leaf data](#approve-leaf-data).
+* `bytes32 merkleRootToCancel`: The Merkle root of the deployment that should be canceled.
+* `address executor`: See [`APPROVE` leaf data](#approve-leaf-data).
+* `string memory uri`: See [`APPROVE` leaf data](#approve-leaf-data).
```
sam-goldman, line 95: This shouldn't say `memory`

specs/merkle-tree.md
```diff
@@ -0,0 +1,166 @@
...
+
+
+## High-Level Merkle Tree Invariants
+The purpose of these invariants is to define an unambiguous, consistent structure for Sphinx Merkle trees. This ensures that it's easy for off-chain tooling to recreate the exact Merkle tree signed by the user, and to check that a given Merkle tree is valid. It's worth mentioning that it's possible to construct Merkle trees that do not adhere to this spec, but are executable on-chain. We discuss this possibility and the expected executor behavior in more detail in the [Assumptions section](#assumptions).
+
+### 1. Must be executable on all chains for which there is at least one Merkle leaf.
+We assume that the input Merkle leaf(s) data satisfies the on-chain conditions in the corresponding `SphinxModuleProxy` contract(s) and that the executor is not buggy. See the [Assumptions section](#assumptions) for more information.
+
+### 2. There must be exactly one `APPROVE` leaf or `CANCEL` leaf per chain
```
sam-goldman, line 104: Should we mention that these leaves must have an `index` field equal to zero? Might be worth putting it into its own invariant to avoid cluttering this one. It's not technically necessary to do this because of the [note in the 6th invariant](https://github.com/sphinx-labs/sphinx/blob/7519ddf2b0c336db5cc9b910f6e2965c6a169c26/specs/merkle-tree.md?plain=1#L121), but it may be worth making this its own invariant

specs/merkle-tree.md
```diff
@@ -0,0 +1,166 @@
...
+
+### 4. `EXECUTE` leaves must *never* exist in a tree with a `CANCEL` leaf on the same chain
+To ensure that all `EXECUTE` leaves are executable, they must *never* follow a `CANCEL` leaf on the same chain. The presence of a `CANCEL` leaf for a given chain [implies there is no `APPROVE` leaf](#2-there-must-be-exactly-one-approve-leaf-or-cancel-leaf-per-chain), and therefore any `EXECUTE` leaves on that chain [will not be executable](#3-execute-leaves-must-always-follow-an-approve-leaf-on-the-same-chain).
+
+### 5. Every leaf `chainId` and `index` combination must be unique within the tree
+Constructing a tree with multiple leaves with the same chain ID and index would make it ambiguous which leaf should be executed, and would result in a tree that cannot be fully executed. So we impose the restriction that in a valid Merkle tree there must be at most one leaf with each chain ID and index combination which ensures there is an unambiguous ordering of leaves within the tree.
+
+### 6. Each leaf on a chain must have an index field with a value that starts with 0 and sequentially increments by 1
```
sam-goldman, line 118: > Each leaf on a chain must have an index field with a value that starts with 0

This sentence is confusing because each leaf on a chain doesn't have an index field that starts with 0. Rather, the `APPROVE` and `CANCEL` leaves have an index that equals zero

specs/merkle-tree.md
```diff
@@ -0,0 +1,166 @@
...
+So for example, if we intended to run a single transaction (one `EXECUTE` leaf) on Goerli (5), OP Goerli (420), and Polygon Mumbai (80001). Then the leaves are expected to be ordered like so:
+1. `APPROVE` Leaf on Goerli (Chain ID 5)
+2. `EXECUTE` Leaf on Goerli
+3. `APPROVE` Leaf on OP Goerli (Chain ID 420)
+4. `EXECUTE` Leaf on OP Goerli
+5. `APPROVE` Leaf on Polyon Mumbai (Chain ID 80001)
+6. `EXECUTE` Leaf on Polygon Mumbai
+
+## Merkle Tree Generation Logic
+We provide utility for generating Sphinx Merkle trees. This utility is used by the official Sphinx Foundry plugin. We expect it to generate Merkle trees that adhere to the above invariants.
```
sam-goldman, line 135: > We provide utility for generating Sphinx Merkle trees. This utility is used by the official Sphinx Foundry plugin.

nit: This sounds choppy. I think these could be merged into a single sentence

specs/merkle-tree.md
```diff
@@ -0,0 +1,166 @@
...
+
+It is also possible to construct Merkle trees that are invalid according to the invariants defined above, but are still executable on-chain. For example, one could construct a Merkle tree with a different method of determining the position of leaves within the tree (such as reverse or a random order), but that still have valid `index` and `chainId` fields for every leaf. It would be possible to execute a tree like this as long as the leaves are submitted in the correct order according to their `index` field. However, this tree would not be valid according to the invariants defined above.
```
sam-goldman, line 155: I like this explanation, but it's a bit confusing that we motivate this section by saying that a third party could implement their own Merkle tree logic (at the end of paragraph 1), then immediately afterwards (in paragraph 2), we talk about how the Merkle tree won't be executable on-chain. Why would a third party plugin write Merkle tree logic that isn't executable on-chain? I think it'd make more sense to switch paragraphs 2 and 3 so that we start by explaining how a Merkle tree can violate the invariants but still be executable on-chain

sam-goldman, line 155: > It would be possible to execute a tree like this as long as the leaves are submitted in the correct order according to their `index` field. However, this tree would not be valid according to the invariants defined above.

It'd be nice to be explicit about which invariant this violates

specs/merkle-tree.md
```diff
@@ -0,0 +1,166 @@
...
+
+We assume that executors like the Sphinx DevOps platform will use the invariants defined in this document to determine whether or not a given Merkle tree is valid. Furthermore, we assume they will refuse to execute Merkle trees that are not valid according to the invariants defined in this document even if it may theoretically be possible to partially or fully execute the Merkle tree.
+
+### Input Data
+Merkle trees generated with this logic are only valid to the extent that the data used to generate them is valid. For example, you could input a `gas` value for a transaction that is not sufficient which would result in the transaction failing. We assume that the input data satisfies the [on-chain conditions in the corresponding SphinxModuleProxy contract(s)](https://github.com/sphinx-labs/sphinx/blob/feature/pre-audit/specs/sphinx-module-proxy.md#function-level-invariants).
```
sam-goldman, line 160: > We assume that the input data satisfies the [on-chain conditions in the corresponding SphinxModuleProxy contract(s)](https://github.com/sphinx-labs/sphinx/blob/feature/pre-audit/specs/sphinx-module-proxy.md#function-level-invariants).

Not sure what you mean by "on-chain conditions". Did you mean to say "assumptions" instead?

specs/merkle-tree.md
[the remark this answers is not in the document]
RPate97: I think the Grammarly vscode plugin just sucks tbh

specs/merkle-tree.md
RPate97: So since you seem really opinionated on this, I'm fine with splitting it up for clarity. We can say this: 
- EACH `APPROVE` or `CANCEL` leaf on a chain must have an index of 0
- Each sequence of `EXECUTE` leaves on a chain must have the index of the first `EXECUTE` leaf be 1, and the index of the following leaves sequentially increment by 1

However, I just want to point out that the above two invariants is functionally equivalent to the original one you're responding too. A more clear version might make this my point easier to see:
- Each sequence of leaves with the same `chainid` must have the `index` of the first leaf be 0, and the `index` of the following leaves sequentially increment by 1

specs/merkle-tree.md
RPate97: Things like the addresses of the safe and module proxies, the nonce, etc. I will update this.

PR #1212, opened 2023-11-24
fix(mt): Merkle tree generation invariant 3 not checked

RPate97:
## Purpose
- Fixes an [issue where invariant 3 of the Merkle tree specification was not properly checked in the Merkle tree generation logic](https://github.com/spearbit-audits/review-sphinx-labs/pull/4#discussion_r1403916886).


packages/contracts/src/merkle-tree.ts
sam-goldman, line 1: Invariant 3 will be violated if there's a `deploymentData` with two entries:
1. `networkData.type === 'deployment' && networkData.arbitraryChain === false`
2. `networkData.type === 'deployment' && networkData.arbitraryChain === true`

This violates the invariant because two approvals will be added even though the second entry has `arbitraryChain === true`.

packages/contracts/test/mocha/merkle-tree.spec.ts
sam-goldman, line 1: Consider expanding test coverage for invariant 3. [Mario pointed out the lack of coverage for a specific case](https://github.com/spearbit-audits/review-sphinx-labs/pull/4#discussion_r1404394677):

> The test for this invariant does not check for detectedArbitraryApproval then no cancel present.

[The invariant violation I noticed](https://github.com/sphinx-labs/sphinx/pull/1212/files#r1404743146) is another case. These two cases aren't exhaustive though. At this point in the audit, we should really be thorough about testing this

sam-goldman, line 1: Looks great

PR #1215, opened 2023-11-24
fix(ct): use version field in EIP-712 domain separator

sam-goldman:
[https://github.com/spearbit-audits/review-sphinx-labs/pull/2#discussion_r1403285272](https://github.com/spearbit-audits/review-sphinx-labs/pull/2#discussion_r1403285272)

We don't include the following fields in the EIP-712 domain separator:

* _chainId_: A single EIP-712 meta transaction may represent many chains, so this field doesn't make sense in this context.
* _verifyingContract_: A single EIP-712 meta transaction may be executed using `SphinxModuleProxy` contracts that exist at different addresses on different chains, so it doesn't make sense to include a single address here.
* _salt_: There's no need for an arbitrary salt value.


packages/contracts/test/ManagedService.t.sol
sam-goldman, line 1: The changes in this file are just cosmetic (from the linter)

PR #1218, opened 2023-11-24
fix(mt): Check NetworkDeploymentData.tx form

RPate97:
## Purpose
- Fixes an issue where [we did not check the form of `NetworkDeploymentData.tx`](https://github.com/spearbit-audits/review-sphinx-labs/pull/4#discussion_r1404324577).


packages/contracts/src/merkle-tree.ts
sam-goldman, line 1: I know this isn't in scope for this PR, but should we check the `type` field of the network data in `isNetworkDeploymentData` and `isNetworkCancellationData`? (I.e. the field that can equal 'deployment' or 'cancellation'). We check every field except for this one

PR #1224, opened 2023-11-24
fix(ct): remove unnecessary merkle root assignment

sam-goldman:
[https://github.com/spearbit-audits/review-sphinx-labs/pull/2#discussion_r1403572425](https://github.com/spearbit-audits/review-sphinx-labs/pull/2#discussion_r1403572425)


packages/contracts/contracts/core/SphinxModule.sol
```diff
@@ -162,9 +162,8 @@ contract SphinxModule is ReentrancyGuard, Enum, ISphinxModule, Initializable {
         if (numLeaves == 1) {
             emit SphinxMerkleRootCompleted(_root);
             state.status = MerkleRootStatus.COMPLETED;
-            // Set the active Merkle root to be `bytes32(0)` so that a new approval can occur in the
-            // future.
-            activeMerkleRoot = bytes32(0);
+            // We don't need to set the `activeMerkleRoot` to equal `bytes32(0)` because we never
+            // assign it to a non-zero value in this call.
```
RPate97, line 166: Nit: It's implied, but it would be nice to state explicitly this is b/c the only leaf is the `APPROVE` leaf.

sam-goldman: Updated this to be:

```
            // We don't need to set the `activeMerkleRoot` to equal `bytes32(0)` because it already
            // equals `bytes32(0)`. Since there's only an `APPROVE` leaf, the Merkle root was
            // approved and completed in this call.
```

PR #1231, opened 2023-11-24
perf(ct): cache `leavesExecuted` in the `execute` function of the `SphinxModule`

sam-goldman:
[https://github.com/spearbit-audits/review-sphinx-labs/pull/2#discussion_r1403732096](https://github.com/spearbit-audits/review-sphinx-labs/pull/2#discussion_r1403732096)


packages/contracts/contracts/core/SphinxModule.sol
sam-goldman, line 1: The main idea is that we increment a `leavesExecuted` variable, which is stored in memory, instead of `state.leavesExecuted`, which is in storage. We update `state.leavesExecuted` when the deployment fails or after all of the leaves in the `execute` function have been executed.

PR #1232, opened 2023-11-25
fix(ct): make destination address payable in `ManagedService`

sam-goldman:
[https://github.com/spearbit-audits/review-sphinx-labs/pull/3#discussion_r1403286195](https://github.com/spearbit-audits/review-sphinx-labs/pull/3#discussion_r1403286195)


packages/contracts/contracts/core/ManagedService.sol
```diff
@@ -21,9 +21,15 @@ contract ManagedService is AccessControl, ReentrancyGuard {
      *
      * @param relayer  The address of the account that made the call.
      * @param to       The address of the remote contract.
+     * @param value    The value transferred from the caller to the destination address.
      * @param dataHash A keccak256 hash of the input data.
      */
-    event Called(address indexed relayer, address indexed to, bytes32 indexed dataHash);
+    event Called(
```
sam-goldman, line 27: I added an event argument the `msg.value`. Seems like a good idea to include this, although it's not necessary

packages/contracts/contracts/core/ManagedService.sol
```diff
@@ -45,13 +51,13 @@ contract ManagedService is AccessControl, ReentrancyGuard {
...
         bytes calldata _data
     ) public payable nonReentrant returns (bytes memory) {
         require(hasRole(RELAYER_ROLE, msg.sender), "ManagedService: invalid caller");
         require(_to != address(0), "ManagedService: target is address(0)");
 
-        emit Called(msg.sender, _to, keccak256(_data));
+        emit Called(msg.sender, _to, msg.value, keccak256(_data));
 
         // slither-disable-next-line arbitrary-send-eth
         (bool success, bytes memory res) = _to.call{ value: msg.value }(_data);
```
RPate97, line 63: Nit: You might already have thought of this, but we could also just cast `_to` as `payable` within the function. That might make calling this function slightly nicer. It's no big deal though I'm good with either. 
```suggestion
        (bool success, bytes memory res) = payable(_to).call{ value: msg.value }(_data);
```

RPate97 (commented):
Looks good. Personally, I would have cast `_to` as `payable`, but this also works fine.

packages/contracts/contracts/core/ManagedService.sol
sam-goldman: I don't have a preference. I'll just keep it as-is

PR #1235, opened 2023-11-25
docs(mt): Update invariant 9 to reflect `@openzeppelin/merkle-tree` leaf ordering

RPate97:
## Purpose
- Fixes an [issue where the ordering of the leaves within the Merkle tree does not accurately reflect the actual ordering used by `@openzeppelin/merkle-tree`](https://github.com/spearbit-audits/review-sphinx-labs/pull/4#discussion_r1403908147).


specs/merkle-tree.md
```diff
@@ -122,16 +122,14 @@ To ensure that `APPROVE` and `CANCEL` leaves always happen first on a given chai
...
 To ensure that all the `EXECUTE` leaves are executed in the correct order, they must have an `index` field that starts at `1` on each chain and sequentially increments by `1`.
 
-### 9. Merkle tree leaves must be ordered in the tree by the leaf's `index` and `chainId` fields ascending
-Sphinx Merkle tree leaves use an explicit `index` field in the leaf rather than an implicit index determined by the leaf's position in the tree. This is because the Sphinx Merkle tree is intended to be executed across multiple networks, and not all leaves will be executed on all networks. We impose the restriction that tree leaves must be ordered by their `index` and `chainId` fields ascending starting at the leftmost position within the tree.
+### 9. Merkle tree leaves must be ordered in the tree by the leaf's hash descending
+To allow for Sphinx Merkle trees to be reliably reproduced by off-chain tooling, we impose the restriction that Sphinx Merkle trees must have their leaves ordered within the tree by their hash descending.
```
sam-goldman, line 126: Grammarly says this is in the passive voice. Consider using this: 

> We enforce that each leaf in a Sphinx Merkle tree must be ordered by the leaf's hash descending. This ordering ensures the Merkle tree can be reliably reproduced using off-chain tooling.

specs/merkle-tree.md
```diff
@@ -122,16 +122,14 @@ To ensure that `APPROVE` and `CANCEL` leaves always happen first on a given chai
...
 
-So, for example, if we intended to run a single transaction (one `EXECUTE` leaf) on Goerli (5), OP Goerli (420), and Polygon Mumbai (80001), then the leaves are expected to be ordered like so:
-1. `APPROVE` Leaf on Goerli (Chain ID 5)
-2. `EXECUTE` Leaf on Goerli
-3. `APPROVE` Leaf on OP Goerli (Chain ID 420)
-4. `EXECUTE` Leaf on OP Goerli
-5. `APPROVE` Leaf on Polyon Mumbai (Chain ID 80001)
-6. `EXECUTE` Leaf on Polygon Mumbai
+The leaf's hash is defined as the fields of the `SphinxLeaf` encoded as bytes and double-hashed using `keccak256`. For reference, the hash of the [SphinxLeaf](https://github.com/sphinx-labs/sphinx/blob/fdf4670089d026ad0bfd20e0ca2b46a02ee5a332/packages/contracts/contracts/core/SphinxDataTypes.sol#L28) can be calculated in Solidity like so:
```
sam-goldman, line 128: This links to the audit commit. Can you either change it to the `develop` branch or make a note to change this once the audit's finished?

specs/merkle-tree.md
```diff
@@ -122,16 +122,14 @@ To ensure that `APPROVE` and `CANCEL` leaves always happen first on a given chai
...
+
+`keccak256(keccak256(abi.encode(leaf.chainId, leaf.index, leaf.leafType, leaf.data)))`.
```
sam-goldman, line 130: > `keccak256(keccak256(abi.encode(leaf.chainId, leaf.index, leaf.leafType, leaf.data)))`.

I'm not sure this is correct. Here's the hashing algorithm in `_getLeafHash`:

```
keccak256(abi.encodePacked(keccak256(abi.encode(_leaf))));
```

I don't know if the ABI encoded `SphinxLeaf` struct equals its ABI encoded fields. (I'm not familiar with the ABI encoding algorithm). Even if they are equal, it's not obvious that this is the case, so I think we should either change this code snippet or just link to the `_getLeafHash` function

sam-goldman, line 130: > The leaf's hash is defined as the fields of the `SphinxLeaf` encoded as bytes and double-hashed using `keccak256`.

As mentioned above, this may not be true. Also, I think we should say "ABI encoded" because "encoded" is vague

specs/merkle-tree.md
sam-goldman, line 128: SphinxLeaf should be in backticks

specs/merkle-tree.md
```diff
@@ -147,6 +145,8 @@ Outputs a [SphinxMerkleTree](https://github.com/sphinx-labs/sphinx/blob/develop/
...
 
+Specifically, the Sphinx Merkle tree generation logic relies on the `@openzeppelin/merkle-tree` dependency to [sort](https://github.com/OpenZeppelin/merkle-tree/blob/0951d3ca036d44c61e33fd9731750cf0ce2978c3/src/standard.ts#L39-L43) the tree leaves in descending order by their [hash](https://github.com/OpenZeppelin/merkle-tree/blob/0951d3ca036d44c61e33fd9731750cf0ce2978c3/src/standard.ts#L9). This is required to satisfy [invariant 9](#9-merkle-tree-leaves-must-be-ordered-in-the-tree-by-the-leafs-hash-descending). We assume that their hashing and sorting logic is robust and bug-free.
```
sam-goldman, line 148: Do these link to the version of the OpenZeppelin Merkle tree library that we actually use? (1.0.5)

specs/merkle-tree.md
```diff
@@ -147,6 +145,8 @@ Outputs a [SphinxMerkleTree](https://github.com/sphinx-labs/sphinx/blob/develop/
 ## Dependencies
 The Merkle tree generation utility makes calls to external libraries. We use `ethers` version 6.7.0 to handle ABI encoding tree leaf `data` and `@openzeppelin/merkle-tree` version 1.0.5 to assemble the Merkle tree. We test that interactions with these libraries work correctly in our [SphinxModuleProxy tests](https://github.com/sphinx-labs/sphinx/blob/develop/packages/contracts/test/SphinxModuleProxy.t.sol) which test that Merkle trees generated by this logic have data that is encoded correctly and that they are executable on-chain. However, we do not test the internals of these libraries and instead rely on the assumption that they are robust and bug-free.
```
sam-goldman, line 146: I know this is out of scope for the PR, but it'd be nice to link to the version of `@openzeppelin/merkle-tree` that we use so readers doesn't need to dig it up themselves

specs/merkle-tree.md
```diff
@@ -122,16 +122,14 @@ To ensure that `APPROVE` and `CANCEL` leaves always happen first on a given chai
...
+### 9. Merkle tree leaves must be ordered in the tree by the leaf's hash descending
+To allow for Sphinx Merkle trees to be reliably reproduced by off-chain tooling, we impose the restriction that Sphinx Merkle trees must have their leaves ordered within the tree by their hash descending.
 
-So, for example, if we intended to run a single transaction (one `EXECUTE` leaf) on Goerli (5), OP Goerli (420), and Polygon Mumbai (80001), then the leaves are expected to be ordered like so:
-1. `APPROVE` Leaf on Goerli (Chain ID 5)
-2. `EXECUTE` Leaf on Goerli
-3. `APPROVE` Leaf on OP Goerli (Chain ID 420)
-4. `EXECUTE` Leaf on OP Goerli
-5. `APPROVE` Leaf on Polyon Mumbai (Chain ID 80001)
-6. `EXECUTE` Leaf on Polygon Mumbai
```
sam-goldman, line 134: Why remove the example? If anything, it may be more helpful now because the order of the leaves is a little more complicated

specs/merkle-tree.md
sam-goldman, line 1: Could you run this through Grammarly?

specs/merkle-tree.md
RPate97: They are equal, but I agree it would be better to link to the `_getLeafHash` function.

specs/merkle-tree.md
RPate97: Good catch, updated to point to the specific commit

specs/merkle-tree.md
RPate97: I was thinking about including this but decided against it. Since the leaves are ordered according to their hashes and most people can't intuitively tell what the correct order of the hashes should be, the example just looks like it's randomly ordered. So it doesn't really add anything wrt to helping the reader understand the ordering mechanism. I think it's better to focus on directing their attention to an example of the formula used to calculate the hash.

specs/merkle-tree.md
sam-goldman: We don't ABI encode the fields when we call OpenZeppelin's library though. We use a tuple, which is a struct from ethers' POV:

```
  const tree = StandardMerkleTree.of(rawLeafArray, [
    'tuple(uint256, uint256, uint8, bytes)',
  ])
```

RPate97: https://docs.soliditylang.org/en/latest/abi-spec.html#encoding-of-indexed-event-parameters

RPate97: I wasn't able to find conclusive docs for or against this, but I tried out this alternative implementation which did not immediately work.

```
export const makeSphinxMerkleTreeFromLeaves = (
  leaves: SphinxLeaf[]
): SphinxMerkleTree => {
  const rawLeafArray = leaves.map((leaf) => Object.values(leaf))
  const tree = StandardMerkleTree.of(rawLeafArray, [
    'uint256',
    'uint256',
    'uint8',
    'bytes',
  ])

  return {
    root: tree.root,
    leavesWithProofs: leaves.map((leaf) => {
      const leafWithProof = {
        leaf,
        proof: tree.getProof(Object.values(leaf)),
      }
      return leafWithProof
    }),
  }
}
```

So to be safe, I'm rewording this to be very clear that we are ABI encoding a struct.

PR #1268, opened 2023-12-13
fix(ct): check Gnosis Safe compatibility using `VERSION()` function instead of codehash

sam-goldman:


packages/contracts/contracts/core/interfaces/ISphinxModule.sol
```diff
@@ -164,22 +164,11 @@ interface ISphinxModule {
...
-     *         - Gnosis Safe L2 v1.3.0
-     *         - Gnosis Safe L1 v1.4.1
-     *         - Gnosis Safe L2 v1.4.1
-     *         These checks do _not_ ensure that the singleton isn't malicious. The singleton can be
-     *         a metamorphic deployment, allowing somebody malicious to self-destruct the singleton,
-     *         causing the user's Gnosis Safe Proxy to be disabled permanently. However, these
-     *         checks are sufficient to prevent users from accidentally adding the module to an
-     *         incompatible Safe.
+     *         This call will revert if the input Gnosis Safe proxy's singleton has a `VERSION()`
+     *         function that does not equal "1.3.0" or "1.4.1". This prevent users from accidentally
```
RPate97, line 168:
```suggestion
     *         function that does not equal "1.3.0" or "1.4.1". This prevents users from accidentally
```

RPate97 (approved):
Looks good

PR #1299, opened 2024-01-03
fix(pg): stop requiring the user to override the `run()` function in their script

sam-goldman:
Before this PR, the user's `run` function needed to include the `override` keyword. After this PR, this keyword is no longer included.

We should make this change now because it's a breaking change that's necessary to allow the user to use function names other than `run()` in their script, which we plan to support in the future. It's a breaking change because the Solidity compiler will throw this error if the user attempts to override a function that doesn't exist: `Error (7792): Function has override specified but does not override anything.`

### DIFF
--- a/packages/plugins/contracts/foundry/Sphinx.sol
+++ b/packages/plugins/contracts/foundry/Sphinx.sol
@@ -163,9 +163,14 @@ abstract contract Sphinx {
         deploymentInfo.requireSuccess = true;
 
         isCollecting = true;
-        vm.startBroadcast(safe);
-        run();
-        vm.stopBroadcast();
+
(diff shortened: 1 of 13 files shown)


packages/plugins/contracts/foundry/Sphinx.sol
```diff
@@ -257,8 +264,6 @@ abstract contract Sphinx {
         sphinxModifierEnabled = false;
     }
 
-    function run() public virtual;
```
sam-goldman, line 260: This line forced the user to include `override` in their `run` function signature

RPate97 (approved):
Looks good

PR #1300, opened 2024-01-04
feat(core): Allow disabling gas price overrides

RPate97:
## Purpose
Adds an environment variable `SPHINX_INTERNAL__DISABLE_GAS_PRICE_OVERRIDES` which allows disabling the gas price overrides function.


sam-goldman (approved):
Lgtm. Would be nice to just document why this env var exists

PR #1304, opened 2024-01-04
fix(pg): calculate correct Sphinx Module address for non-zero salt nonce

sam-goldman:


packages/plugins/src/foundry/decode.ts
```diff
@@ -328,7 +328,6 @@ export const makeParsedConfig = (
             input.decodedAction = {
               referenceName: contractName,
               functionName: 'deploy',
-              // TODO: We could probably get the constructor args from the init code with some effort since we have the FQN
```
sam-goldman, line 331: I put these TODOs in linear so I removed them here

packages/plugins/src/foundry/decode.ts
```diff
@@ -359,7 +358,6 @@ export const makeParsedConfig = (
               input.decodedAction = {
                 referenceName: fullyQualifiedName.split(':')[1],
                 functionName: 'deploy',
-                // TODO: We could probably get the constructor args from the init code with some effort since we have the FQN
```
sam-goldman, line 362: > I put this TODO in linear so I removed it here

packages/plugins/test/foundry/Sphinx.t.sol
```diff
@@ -0,0 +1,42 @@
...
+        sphinxConfig.saltNonce = 1;
+        IGnosisSafe safeProxy = IGnosisSafe(deploySphinxModuleAndGnosisSafe(sphinxConfig));
+
+        (address[] memory modules, ) = safeProxy.getModulesPaginated(address(0x1), 1);
+        address sphinxModule = modules[0];
+
+        address expectedAddress = this.sphinxModule();
+
+        assertEq(expectedAddress, sphinxModule);
+    }
```
sam-goldman, line 41: This test wouldn't have passed previously

packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -858,7 +834,22 @@ contract SphinxUtils is SphinxConstants, StdUtils {
...
+                // not necessary to include the `saltNonce` here because a new Sphinx Module will be
+                // deployed if the user sets the `sphinxConfig.saltNonce` to a new value and then
+                // deploys a new Gnosis Safe using Sphinx's standard deployment method. A new Sphinx
+                // Module is deployed in this scenario because its address is determined by the
+                // address of the Gnosis Safe. It'd only be necessary to include a `saltNonce` for
+                // the Sphinx Module if a single Gnosis Safe wants to enable multiple Sphinx
+                // Modules, which isn't a feature that we currently support.
+                0
+            )
+        );
```
sam-goldman, line 852: This is the area that was causing the bug

sam-goldman, line 852: This is the area that resolves the bug. We were previously using `_config.saltNonce` to calculate the Sphinx Module's address, which isn't correct because this field is only meant to determine the Gnosis Safe's address. Hard-coding this value to `0` resolves the bug

packages/plugins/script/write-initcode.ts
```diff
@@ -0,0 +1,55 @@
+import { ethers } from 'ethers'
+import { getSphinxConstants, remove0x } from '@sphinx-labs/contracts'
+
+/**
+ * Writes the initcode and expected addresses of all Sphinx system contracts. This allows us to use
+ * the contracts in Forge tests.
```
RPate97, line 6: Nit: This is not directly related to this PR, but it would be nice to document why we do not need to do this in the main library (we deploy the core contracts on local networks using `ensureSphinxAndGnosisSafeDeployed`).

RPate97 (approved):
Looks good

PR #1305, opened 2024-01-05
fix(pg): set simulation account balance to avoid running out of funds

sam-goldman:
The simulation on Arbitrum Sepolia was failing because we weren't setting the caller's balance to be high enough. We set it to the maximum possible ETH value to avoid this issue.


RPate97 (approved):
Looks good

PR #1307, opened 2024-01-05
perf(pg): compile as late as possible in CLI commands

sam-goldman:
Before this PR, the Deploy and Propose CLI commands force re-compiled the user's contracts at the beginning of the command. The problem is that it's common to spend a significant amount of time waiting for re-compilation to occur then run into an error thrown at some point later in the command. For example, this scenario can occur if an error is thrown in the user's script during the transaction collection phase. When this occurs, the user must then sit and wait for re-compilation to occur all over again. If this process repeats multiple times, it becomes quite frustrating to use our plugin.

This PR fixes the issue by force-recompiling as late as possible in these commands. Specifically, we force recompile before calling `getConfigArtifacts`. This PR also moves a bit of logic earlier in these commands so that re-compilation happens as late as possible.

### DIFF
--- a/packages/demo/test/Solc.spec.ts
+++ b/packages/demo/test/Solc.spec.ts
@@ -86,6 +86,8 @@ describe('Solidity Compiler', () => {
     // Generate output directory names. We use separate output directories for each compilation to
     // prevent race conditions.
     const outputDirs = versions.map((version) => `out-${version}`)
+    // Generate separate cache directory names to prevent race conditions.
+    const cacheDirs = versions.map((version) => `cache-${version}`)
 
     const buildPromises = versions.map((version, index) => {
       return spawnAsync(`forge`, [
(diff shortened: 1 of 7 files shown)


packages/plugins/src/cli/propose/index.ts
```diff
@@ -73,24 +74,15 @@ export const buildParsedConfigArray = async (
   scriptPath: string,
   isTestnet: boolean,
   sphinxPluginTypesInterface: ethers.Interface,
-  sphinxContext: SphinxContext,
+  foundryToml: FoundryToml,
+  getConfigArtifacts: GetConfigArtifacts,
```
sam-goldman, line 78: @RPate97 Fyi, I changed the params of this function slightly. Just wanted to lyk because I think you mentioned the tests in the website use this function

RPate97: I was referring to `buildDeploymentWithCompilerConfigs`, thanks though.

packages/demo/test/Solc.spec.ts
sam-goldman, line 1: The updates to this file resolve a race condition that was sometimes causing this test suite to fail in CI

RPate97 (approved):
Looks good

PR #1308, opened 2024-01-05
fix(core): set simulation transaction limit less than block gas limit

sam-goldman:
Prevents errors like this from occurring in the simulation:
```
Simulation failed for polygon_mumbai: InvalidInputError: Transaction gas limit is 20473547 and exceeds block gas limit of 20000000
```


RPate97 (approved):
Looks good

PR #1309, opened 2024-01-06
test(pg): test that the simulation works on all supported live networks

sam-goldman:
This PR adds a test that checks that the simulation works on all supported live networks. I manually checked that this test would have caught [this bug caused by fluctuating gas prices](https://github.com/sphinx-labs/sphinx/pull/1308) and [this bug caused by a high block gas limit](https://github.com/sphinx-labs/sphinx/pulls?q=is%3Apr+is%3Aclosed)

### DIFF
--- a/.changeset/sour-feet-reflect.md
+++ b/.changeset/sour-feet-reflect.md
@@ -1,0 +1,6 @@
+---
+'@sphinx-labs/plugins': patch
+'@sphinx-labs/core': patch
+---
+
+Test that the simulation works on all supported live networks
--- a/packages/core/src/networks.ts
+++ b/packages/core/src/networks.ts
@@ -2,6 +2,8 @@
 // Be careful when importing external dependencies to this file because they may cause issues when this file
 // is imported by the website.
 
+import { isSupportedNetworkName } from './utils'
+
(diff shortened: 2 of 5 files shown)


RPate97 (approved):
Looks good. 

I did notice a different bug related to the simulation that doesn't seem to be picked up by this test suite: [CHU-357](https://linear.app/chugsplash/issue/CHU-357/deploy-task-simulation-fails-due-to-extremely-high-funds-required-for). I think it's worth spending some time investigating why that bug is not detected by this test suite. However, I don't think it's necessary to hold this up on addressing that.

sam-goldman:
> I did notice a different bug related to the simulation that doesn't seem to be picked up by this test suite: [CHU-357](https://linear.app/chugsplash/issue/CHU-357/deploy-task-simulation-fails-due-to-extremely-high-funds-required-for). I think it's worth spending some time investigating why that bug is not detected by this test suite. However, I don't think it's necessary to hold this up on addressing that.

It makes sense that this test suite doesn't detect the error because the test suite only checks the proposal mode, where the caller has an unlimited amount of ETH. Since you already approved this PR and made a ticket for that bug, I'll go ahead and merge this PR and then address the bug separately

PR #1310, opened 2024-01-08
fix(pg): remove logic that fast forwards block number on forked local nodes

sam-goldman:
This PR resolves a bug that was occurring on Anvil nodes that fork live networks. The bug was caused by the fact that we were fast forwarding the block number on these networks, which was causing network timeouts. Now, we only fast forward on non-forked local nodes (i.e. nodes that begin with an entirely fresh state).

### DIFF
--- a/packages/plugins/src/hardhat/simulate.ts
+++ b/packages/plugins/src/hardhat/simulate.ts
@@ -11,8 +11,9 @@ import {
   ExecutionMode,
   MerkleRootStatus,
   SphinxJsonRpcProvider,
-  isLiveNetwork,
   getNetworkNameForChainId,
+  isLiveNetwork,
+  isFork,
 } from '@sphinx-labs/core'
 import { ethers } from 'ethers'
 import {
@@ -89,7 +90,9 @@ export const simulate = async (
     chainId,
   }
 
-  if (!(await isLiveNetwork(provider))) {
+  if (!(await isLiveNetwork(provider)) && !(await isFork(provider))) {
+    // The network is a non-forked local node.
+
     // Fast forward 1000 blocks. This is necessary to prevent the following edge case that occurs
     // when running the simulation against a vanilla Anvil node:
     // 1. We deploy the Gnosis Safe and Sphinx contracts.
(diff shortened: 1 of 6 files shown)


packages/plugins/src/hardhat/simulate.ts
```diff
@@ -89,24 +92,45 @@ export const simulate = async (
...
+    //    Hardhat fork.
+    const blocksToFastForward = getLargestPossibleReorg(chainId)
+    const blocksHex = stripLeadingZero(ethers.toBeHex(blocksToFastForward))
     await provider.send(
-      // The `hardhat_mine` RPC method works on Anvil and Hardhat nodes.
-      'hardhat_mine',
-      // 1000 blocks.
-      ['0x3e8']
+      'hardhat_mine', // The `hardhat_mine` RPC method works on Anvil and Hardhat nodes.
+      [blocksHex]
```
sam-goldman, line 133: This snippet isn't related to the main purpose of the PR, but I was lazy and included it anyways. Mining 1000 blocks resulted in a huge amount of output logged to the Anvil node terminal (~5 lines per block). I resolved this by fast forwarding the minimum number of blocks necessary to prevent this edge case from occurring.

packages/plugins/test/mocha/simulate.spec.ts
```diff
@@ -50,15 +62,37 @@ describe('Simulate', () => {
...
+
+    // Process each batch sequentially
+    for (const batch of batches) {
+      const batchResults = await Promise.allSettled(
+        batch.map((parsedConfig) =>
+          simulate(
+            parsedConfigArray,
+            parsedConfig.chainId,
+            fetchURLForNetwork(toSupportedChainId(Number(parsedConfig.chainId)))
+          )
```
sam-goldman, line 91: We now batch the simulations because our CI process was rate-limited by Alchemy when we ran all of the simulations in parallel

packages/plugins/test/mocha/simulate.spec.ts
```diff
@@ -5,17 +5,27 @@ import {
...
+  getAnvilRpcUrl,
+  killAnvilNodes,
+  makeDeployment,
+  makeStandardDeployment,
+  startForkedAnvilNodes,
+} from './common'
 import { simulate } from '../../src/hardhat/simulate'
 
 chai.use(chaiAsPromised)
+const expect = chai.expect
```
RPate97, line 28: You can just import `expect` alongside chai like this:
```
import chai, { expect } from 'chai'
```

RPate97 (approved):
Looks good

PR #1311, opened 2024-01-08
fix(pg): make block numbers more accurate in simulation

sam-goldman:
This PR does two things:
1. On live networks and forks of live networks, we use the latest block number, which is aligned with Anvil's behavior and it ensures that any recently submitted transactions are included in the simulation. It's worth mentioning that Hardhat warns against using the latest block number because of chain reorgs, but I think it's more important that we align with Anvil's behavior. Otherwise, we risk using an outdated chain state in the simulation.
2. On non-forked local nodes, we only fast forward the necessary number of blocks instead of 1000 blocks.

### DIFF
--- a/packages/core/src/networks.ts
+++ b/packages/core/src/networks.ts
@@ -36,6 +36,27 @@ export type SupportedNetworkName =
   | SupportedTestnetNetworkName
   | SupportedLocalNetworkName
 
+/**
+ * Data returned by the `anvil_metadata` and `hardhat_metadata` RPC methods.
+ *
+ * @param forkedNetwork Info about the network that the local node is forking, if it exists. If the
+ * local node isn't forking a network, this field can be `undefined` or `null` depending on whether
+ * the network is an Anvil or Hardhat node.
+ */
+export type LocalNetworkMetadata = {
+  clientVersion: string
+  chainId: number
+  instanceId: string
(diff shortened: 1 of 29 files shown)


packages/core/src/networks.ts
```diff
@@ -436,3 +436,44 @@ export const toSupportedChainId = (chainId: number): SupportedChainId => {
   }
   return chainId as SupportedChainId
 }
+
+/**
+ * The number of blocks that Hardhat rewinds when forking the given network. Rewinding the block
+ * number protects against chain reorgs. Copied from Hardhat:
+ * https://github.com/NomicFoundation/hardhat/blob/caa504fe0e53c183578f42d66f4740b8ec147051/packages/hardhat-core/src/internal/hardhat-network/provider/utils/reorgs-protection.ts
+ */
+export const getLargestPossibleReorg = (chainIdStr: string): bigint => {
```
RPate97, line 445: Why isn't this imported from Hardhat since we have that installed as a production dependency anyway?

sam-goldman: It's an "internal" directory, so I don't feel confident that the file location and the function signature will be the same in future minor/patch versions of Hardhat

RPate97 (approved):
Looks good

PR #1312, opened 2024-01-08
fix(pg): display error from child process in simulation

sam-goldman:
Previously, if the simulation threw an error, any warnings from the simulation were also displayed in the error message. It wasn't possible to distinguish between a warning and the error that caused the simulation to fail, which made it difficult to understand why the error occurred. This PR fixes this issue by only displaying the error message that caused the simulation to fail.

### DIFF
--- a/packages/plugins/src/hardhat/simulate.ts
+++ b/packages/plugins/src/hardhat/simulate.ts
@@ -152,7 +152,7 @@ export const simulate = async (
   // (although unlikely) that this same situation could happen in a user's test suite. We resolve
   // this by creating the Hardhat node in a child process via `spawnAsync`. This child process exits
   // when `spawnAsync` returns.
-  const { stdout, stderr, code } = await spawnAsync(
+  const { stdout, code } = await spawnAsync(
     'node',
     [hardhatRunnerPath],
     envVars,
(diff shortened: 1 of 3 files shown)


packages/plugins/src/hardhat/simulate.ts
```diff
@@ -161,9 +161,28 @@ export const simulate = async (
...
+        errorMessage += `\n\n${error.stack}`
+      } else {
+        // Display both the error message and the stack trace.
+        errorMessage += `\n\n${error.message}\n\n${error.stack}`
+      }
+    } catch {
+      // Do nothing if an error occurs while attempting to decode the error from the child process.
+      // We'll throw a generic error shortly.
+    }
+    throw new Error(errorMessage)
```
RPate97, line 185: I would include the raw value of stdout in the error message. It's hard to predict what the value of stdout would be in this case, so I'd advise airing on the side of caution by passing it through to the user in case it's useful.

RPate97 (approved):
Looks good, I just left one minor comment.

packages/plugins/src/hardhat/simulate.ts
sam-goldman: Good idea, added that

PR #1317, opened 2024-01-09
fix(ct): allow `storageLayout.types` to be `null`

sam-goldman:


packages/contracts/src/utils.ts
```diff
@@ -244,8 +244,7 @@ const isValidStorageLayout = (storageLayout: any): boolean => {
     storageLayout !== null &&
     typeof storageLayout === 'object' &&
     Array.isArray(storageLayout.storage) &&
-    typeof storageLayout.types === 'object' &&
-    storageLayout.types !== null
+    typeof storageLayout.types === 'object'
```
sam-goldman, line 247: `null` has a type of `object`, so this returns `true` if `storageLayout.types` is `null`

RPate97: Damn

RPate97 (approved):
Looks good

PR #1322, opened 2024-01-09
fix(pg): skip the manual confirmation step in proposal dry run

sam-goldman:
This PR fixes an issue where the user needed to manually confirm the preview when they specified the `--dry-run` flag in the proposal CLI command. This PR also adds a few new tests:
1. Manual confirmation is skipped if the `--dry-run` flag is specified.
2. Manual confirmation is skipped if the `--confirm` flag is specified. 
3. Manual confirmation is required if the neither of these flags are specified.
4. An error is thrown if the user specifies `--dry-run` and `--confirm`. We throw an error because these flags are redundant, which signals user error or a misunderstanding of the commands.

The first three tests are in `propose.spec.ts` and the last test is in `cli.spec.ts`.

This PR also adds some heavyweight proposal-related functions to the `SphinxContext`, particularly  `buildParsedConfigArray`, and `getNetworkGasEstimates`. I originally added these functions because I was planning on mocking them to test that the manual confirmation is skipped. However, I decided not to use these mocks because it was easiest to test that the confirmation is skipped in the existing proposal integration tests, where these functions aren't mocked. I kept these fields on the `SphinxContext` anyways because I figured we'll probably use them soon.


packages/plugins/test/mocha/mock.ts
sam-goldman, line 1: We had a common.ts and a utils.ts that both contained helper functions for our plugin tests. I moved the helper functions in utils.ts into common.ts and added a mock.ts file that contains all of the mock-related functionality.

packages/plugins/test/mocha/cli/deploy.spec.ts
sam-goldman, line 1: This test suite and `propose.spec.ts` now use `makeMockSphinxContextForIntegrationTests` instead of `makeMockSphinxContext`. The latter mocks a lot of functionality used in the propose and deploy CLI commands, so it's not appropriate to use it in integration tests.

packages/plugins/test/mocha/cli/propose.spec.ts
```diff
@@ -167,14 +171,14 @@ describe('Propose CLI command', () => {
...
-    const context = makeMockSphinxContext([
+    const { context, prompt } = makeMockSphinxContextForIntegrationTests([
       'contracts/test/MyContracts.sol:MyContract2',
     ])
     const { proposalRequest, parsedConfigArray, configArtifacts } =
       await propose({
         confirm: true, // Skip preview
         isTestnet,
-        isDryRun: true,
+        isDryRun: false,
```
sam-goldman, line 181: Changed `isDryRun` to `false` here and below because we don't allow `confirm` and `dryRun` to both be `true` anymore

packages/plugins/test/mocha/cli/deploy.spec.ts
sam-goldman, line 1: This test suite and `propose.spec.ts` now use `makeMockSphinxContextForIntegrationTests` instead of `makeMockSphinxContext`. The latter mocks a lot of functionality used in the propose and deploy CLI commands, so it's not appropriate to use it in integration tests.

packages/plugins/src/cli/propose/index.ts
```diff
@@ -314,7 +312,7 @@ export const propose = async (
   )
 
   const { parsedConfigArray, configArtifacts, isEmpty } =
-    await buildParsedConfigArray(
+    await sphinxContext.buildParsedConfigArray(
```
RPate97, line 315: I don't think this should be included in this PR.

packages/plugins/src/cli/propose/index.ts
```diff
@@ -348,7 +346,7 @@ export const propose = async (
   const gasEstimatesPromises = parsedConfigArray
     .filter((parsedConfig) => parsedConfig.actionInputs.length > 0)
     .map((parsedConfig) =>
-      getNetworkGasEstimate(
+      sphinxContext.getNetworkGasEstimate(
```
RPate97, line 349: Same here.

packages/plugins/test/mocha/mock.ts
```diff
@@ -0,0 +1,189 @@
...
+import { propose } from '../../src/cli/propose'
+import { deploy } from '../../src/cli/deploy'
+import { makeSphinxContext } from '../../src/cli/context'
+
+/**
+ * Make a mocked `SphinxContext` object. Use this function if it's safe to assume that all of
+ * `SphinxContext` member functions are mocked. In integration tests, use the
+ * `makeMockSphinxContextForIntegrationTests` function instead.
+ */
+export const makeMockSphinxContext = (
```
RPate97, line 22: I'm confused why there's two functions here, when we are only using one of them. 

It seems like you want to introduce a new pattern where we have this function for unit tests, then we use the `makeMockSphinxContextForIntegrationTests` which we use for larger integration tests? That seems reasonable to me, but I wouldn't recommend making this type of change unless you're going to fully execute on it. I.e add a few new tests that use the new pattern, update the organization of the tests files to clearly differentiate between integration tests and this new type of tests, etc. It's just easy to make this type of change and then never actually come back and finish it which can leave you with a bunch of half finished work that confuses other devs. 

I don't think it's necessary to revert this since it seems like a reasonable change. I would just add documentation that this function isn't actually used currently, and provide an explanation for the intended use case. In the future, I would recommend making this type of change in it's own PR and taking a bit more time to fully think through the change and make sure that the PR includes the full pattern.

RPate97 (approved):
So I think the content of this PR is good, and I'm going to approve this in the spirit of moving forward and learning over time. 

However, this includes a lot of changes that should not be in the same PR. Namely, the mocks for `buildParsedConfigArray` and `getNetworkGasEstimates` along with the introduction of `makeMockSphinxContextForIntegrationTests`. I think all of those should have gone into their own PR aimed at improving our mocking logic and adding a new pattern for lighter-weight unit tests. I think those changes are significant enough changes to warrant their own dedicated thought and effort. As they are now, I think the changes are incomplete and somewhat confusing.

I'm generally in favor of refactoring while making other changes as long as the refactoring is relevant to the change, complete, and minor enough to not slow down the process of reviewing and merging the change itself. In this case, I think we're including too much logic to the point that it outweighs the change itself. I spent more time looking through those changes than I did any of the changes related to the confirm flag.

packages/plugins/test/mocha/cli/propose.spec.ts
sam-goldman: You're right that the restriction doesn't apply here. I'll just keep the check in the `proposeCommandHandler` function since neither of us are opinionated

packages/plugins/test/mocha/mock.ts
sam-goldman: > I'm confused why there's two functions here, when we are only using one of them.

We use `makeMockSphinxContext` [here](https://github.com/sphinx-labs/sphinx/pull/1322/files#diff-af73bf84e06d033e941580bf28757ba49a6cd79dc942568d357a5f77d0e3bfadR181) to make the `SphinxContext` for integration tests. You're right that we don't use it directly in our tests though. 

> I would provide an explanation for the intended use case

I already included an explanation:
```
/**
 * Make a mocked `SphinxContext` object. Use this function if it's safe to assume that all of
 * `SphinxContext` member functions are mocked. In integration tests, use the
 * `makeMockSphinxContextForIntegrationTests` function instead.
 */
```

Are you looking for something more specific?

sam-goldman:
> However, this includes a lot of changes that should not be in the same PR. Namely, the mocks for `buildParsedConfigArray` and `getNetworkGasEstimates` along with the introduction of `makeMockSphinxContextForIntegrationTests`. I think all of those should have gone into their own PR aimed at improving our mocking logic and adding a new pattern for lighter-weight unit tests. I think those changes are significant enough changes to warrant their own dedicated thought and effort. As they are now, I think the changes are incomplete and somewhat confusing.
> 
> I'm generally in favor of refactoring while making other changes as long as the refactoring is relevant to the change, complete, and minor enough to not slow down the process of reviewing and merging the change itself. In this case, I think we're including too much logic to the point that it outweighs the change itself. I spent more time looking through those changes than I did any of the changes related to the confirm flag.

I hear you, my bad

packages/plugins/test/mocha/mock.ts
RPate97: I just would like more detail on exactly what the intended use case for this function is. I don't really understand what you mean by "if it's safe to assume that all of `SphinxContext` member functions are mocked". In what case would that be exactly? 

Is this function intended to be used when testing the deploy/propose tasks themselves, or should it be used only when testing lower-level functions? What specific use case did you have in mind when you created it?

PR #1323, opened 2024-01-09
fix(pg): use default gas estimation overrides on live networks

sam-goldman:
### Problem

Previously, the simulation was setting the transaction gas limit to be extremely high for live network CLI deployments, which resulted in this error: `sender doesn't have enough funds to send txn`. This error was happening because the `getGasPriceOverrides` function set the transaction's gas limit to be extremely high when `isLiveNetwork` was `false`, which always occurs in the simulation due to the fact that the network is a Hardhat node. [Here's the snippet of code that was causing the bug](https://github.com/sphinx-labs/sphinx/blob/main/packages/core/src/utils.ts#L160-L187).

### Solution

Instead of using `isLiveNetwork` to determine when to make the transaction's gas limit extremely high, we now use the `ExecutionMode` enum, which we already use throughout the codebase to distinguish between local node deployments, live network deployments, and proposals. We now set the transaction's gas limit to be extremely high for local node deployments and proposals but not live network deployments.

This PR also updates the live network deployment integration test to use a signer with a small amount of funds. I checked that this would've caught the error. [Here's the snippet in `deploy.spec.ts` that makes this change](https://github.com/sphinx-labs/sphinx/blob/samgoldman12/chu-357-deploy-task-simulation-fails-due-to-extremely-high-funds/packages/plugins/test/mocha/cli/deploy.spec.ts#L252-L263).

### Notes

* This entire PR is basically just refactoring our codebase to pass the `executionMode` into `getGasPriceOverrides`


packages/core/src/utils.ts
```diff
@@ -169,14 +169,17 @@ export const getGasPriceOverrides = async (
...
-    // don't override this on live networks because the signer is the user's wallet, which may have
-    // a limited amount of ETH. It's fine to set a very high gas limit on local networks because we
-    // use an auto-generated wallet to execute the transactions. We set it to 3/4 of the block gas
-    // limit for two reasons:
+    // significantly speeds up deployments because it removes the need for EthersJS to call
+    // `eth_estimateGas`, which is a very slow operation for large transactions. We only make this
+    // optimization in situations where we can safely assume that the caller has an unlimited amount
+    // of ETH. We don't override this on live networks because the signer is the user's wallet,
+    // which may have a limited amount of ETH. We set it to 3/4 of the block gas limit for two
+    // reasons:
```
sam-goldman, line 182: Updated this explanation to clarify when it's safe to make this optimization

packages/core/src/utils.ts
```diff
@@ -169,14 +169,17 @@ export const getGasPriceOverrides = async (
     throw new Error(`Unable to retrieve latest block.`)
   }
 
-  if (!isLiveNetwork_) {
+  if (
+    executionMode === ExecutionMode.LocalNetworkCLI ||
+    executionMode === ExecutionMode.Platform
+  ) {
```
sam-goldman, line 175: This is the line that fixes the bug

packages/core/src/utils.ts
```diff
@@ -158,9 +159,8 @@ export const getGasPriceOverrides = async (
     return overridden
```
RPate97, line 159: We can remove this if statement since this PR addresses the underlying issue.

RPate97 (approved):
Looks good thank you for the extensive description.

RPate97:
It's worth noting this is a breaking change for the website since it modifies the interface of the `getGasPriceOverrides` function, which is used directly by the relayer service. That's fine, though. I just wanted to mention it.

PR #1324, opened 2024-01-09
fix(pg): remove force recompilation in deploy and propose CLI commands

sam-goldman:
Significantly speeds up the deploy and propose CLI commands


RPate97 (approved):
Looks good

PR #1325, opened 2024-01-10
fix(pg): use `CREATE2` salt in quickstart guide to prevent contract address collision

sam-goldman:
This PR fixes the quickstart guide, which broke because it attempted to deploy the sample contract to the same address for everybody.


RPate97 (approved):
Looks good

PR #1334, opened 2024-01-12
maint(core): Simplify network integration

RPate97:
## Purpose
Simplifies the interface for adding support for new networks by refactoring the way we store and fetch network-specific data such as rpc URLs, currencies, etherscan urls, chain IDs, etc. The goal of this is to move to a single source of true for all network data so that we can easily add support for new networks without needing to update a ton of different things throughout the codebase. A secondary goal was to create a pattern that was more compatible with the idea of supporting arbitrary networks.

## Refactor Summary
At a high level, what I've done is moved all network-specific data into a [new array of objects, `SPHINX_NETWORKS`](https://github.com/sphinx-labs/sphinx/pull/1334/files#diff-70d23bf9efc5558c8798ea0e2d9887632be876d6e65c21e61b6542e517e70125R29). I then refactored everything to fetch data from that array. I've also updated the `Network` Solidity enum to be automatically generated from this array in the [write-constants script](https://github.com/sphinx-labs/sphinx/pull/1334/files#diff-6b765f650362aebf5db40be2672febe27d2d30c64ecc1d10136227800b769d3cR76). 

### This accomplishes two things:
1. It makes the `SPHINX_NETWORKS` object the single source of truth for network data. So we can now add support for new networks just by adding a new object to that array. 
2. It changes the network data into something that could be supplied at runtime rather than compile time. Currently, the data is still hardcoded, so the difference isn't super significant. But in the future, if we wanted to, we could supply the network data from an outside source, such as a database with relatively minimal updates. This would be required to support arbitrary networks.

### Types
Unfortunately, this pattern does have one tradeoff, which is that we don't get full-type safety when it comes to network names, chain IDs, etc. Technically, we could still get full type safety by automatically generating TS types using the `SPHINX_NETWORKS` object as I did with the Networks enum. However, I don't think we should do this because it conflicts with the general idea of the network data being supplied at runtime rather than compile time. I only did this with the Network enum because we have to do it with the way the product currently works. In the future, I would consider changing the product to avoid this. We'll have to do that to support arbitrary networks on the website.


packages/contracts/contracts/foundry/SphinxConstants.sol
```diff
@@ -10,3 +10,43 @@ contract SphinxConstants {
   address public constant safeSingletonAddress = 0xd9Db270c1B5E3Bd161E8c8503c55cEABeE709552;
   address public constant sphinxModuleImplAddress = 0x8f4E4d51B8050B0ff713eff1F88f3dD8b5e8a530;
 }
+
+enum Network {
```
RPate97, line 14: This is now automatically generated from the `SPHINX_NETWORKS` object in the `write-constants.ts` script.

packages/contracts/src/networks.ts
```diff
@@ -0,0 +1,480 @@
...
+  //   rpcUrl: () =>
+  //     `https://celo-mainnet.infura.io/v3/${process.env.INFURA_API_KEY}`,
+  //   etherscan: {
+  //     apiURL: 'https://api.celoscan.io/api',
+  //     browserURL: 'https://celoscan.io/',
+  //   },
+  //   currency: 'CELO',
+  //   dripSize: '1',
+  //   requiredEnvVariables: ['INFURA_API_KEY'],
+  // },
```
RPate97, line 317: I was originally going to also add support for all the networks requested by mean finance as part of this PR. I chose not to, though, b/c I don't want this PR to be blocked by all the related tasks (deploying contracts, getting etherscan keys, handling funds on all the networks, etc).

packages/contracts/contracts/foundry/SphinxConstants.sol
```diff
@@ -9,4 +11,175 @@ contract SphinxConstants {
   address public constant safeFactoryAddress = 0xa6B71E26C5e0845f74c812102Ca7114b6a896AB2;
   address public constant safeSingletonAddress = 0xd9Db270c1B5E3Bd161E8c8503c55cEABeE709552;
   address public constant sphinxModuleImplAddress = 0x8f4E4d51B8050B0ff713eff1F88f3dD8b5e8a530;
+
+  uint8 internal constant numSupportedNetworks = 23;
+
+  function getNetworkInfoArray() public pure returns (NetworkInfo[] memory) {
```
RPate97, line 17: We now automatically generate this function and the Network enum from the network array defined in typescript.

packages/core/src/utils.ts
```diff
@@ -647,8 +600,8 @@ export const getNetworkTag = (
   ) {
     return networkName
   } else if (
-    Object.keys(SUPPORTED_NETWORKS).includes(networkName) &&
-    !Object.keys(SUPPORTED_LOCAL_NETWORKS).includes(networkName)
+    SPHINX_NETWORKS.find((n) => n.name === networkName) &&
+    !SPHINX_LOCAL_NETWORKS.find((n) => n.name === networkName)
```
sam-goldman, line 604:
```suggestion
    !SPHINX_LOCAL_NETWORKS.some((n) => n.name === networkName)
```

nit: I think [Array.some](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) is more appropriate here, but `find` works too

sam-goldman (approved):
Great stuff

PR #1335, opened 2024-01-12
feat(pg): support CLI deployments on networks not supported by the DevOps Platform

sam-goldman:
Added comments to this PR to explain various implementation details. Should make it easier to review

One note: if the user deploys the system contracts, we don't attempt to verify them on Etherscan. I chose not to do this because the logic that verifies the system contracts does not handle retries, which means it'd likely halt the verification process for the user. I'll make a Linear ticket for this limitation.


README.md
```diff
@@ -21,6 +21,8 @@ Deployments with Sphinx are a three-step process:
 
 * **Compatible with Forge Scripts**: You can integrate Sphinx by adding a few lines of code to your existing Forge scripts.
 
+* **No Lock-In**: You can execute deployments from your local machine without using the Sphinx DevOps Platform. These deployments are executed in the exact same manner as they would be executed with the DevOps Platform, including identical contract addresses.
```
sam-goldman, line 24: @RPate97 Curious to hear your thoughts on this bullet point

README.md
```diff
@@ -38,7 +40,7 @@ Sphinx is currently invite-only. [Request access on our website.](https://sphinx
 - [CLI Propose Command](https://github.com/sphinx-labs/sphinx/blob/main/docs/cli-propose.md)
 - [Propose in CI](https://github.com/sphinx-labs/sphinx/blob/main/docs/ci-proposals.md)
 - [Configuration Options](https://github.com/sphinx-labs/sphinx/blob/main/docs/configuration-options.md)
-- [Deploy from the CLI](https://github.com/sphinx-labs/sphinx/blob/main/docs/cli-deployments.md)
+- [Deploy from the CLI](https://github.com/sphinx-labs/sphinx/blob/main/docs/cli-deploy.md)
```
sam-goldman, line 43: Changed the file name from 'cli-deployments' to 'cli-deploy' to align with the proposal CLI guide (`cli-propose.md`)

docs/cli-deploy.md
sam-goldman, line 1: Overhauled this doc to match the structure of the propose CLI command and also explain the deploy command's behavior if the system contracts aren't deployed on a network

docs/cli-propose.md
```diff
@@ -37,8 +37,8 @@ pnpm sphinx propose <SCRIPT_PATH> --networks <testnets|mainnets> [options]
 - `--networks <testnets|mainnets>`: **Required**. Choose between proposing on the test networks (`sphinxConfig.testnets`) or the production networks (`sphinxConfig.mainnets`) in your script.
 - `--confirm`: **Optional**. Confirm the proposal without previewing it. Useful for automating proposals.
 - `--dry-run`: **Optional**. Perform a trial run without sending data to Sphinx's backend. Useful for testing and validation.
-- `--silent`: **Optional**. Suppress output to display only error messages. Combine with `--confirm` for silent, confirmed deployments.
-- `--target-contract <TARGET_CONTRACT>`: **Optional**. Specify a contract in multi-contract scripts. Alias: `--tc <TARGET_CONTRACT>`.
+- `--silent`: **Optional**. Suppress output to display only error messages. Must be combined with `--confirm` to confirm the proposal without previewing it.
+- `--target-contract <TARGET_CONTRACT>` (Alias: `--tc <TARGET_CONTRACT>`): **Optional**. Specify a contract in multi-contract scripts.
```
sam-goldman, line 41: Minor changes to match the deploy CLI command options

packages/contracts/contracts/foundry/SphinxPluginTypes.sol
```diff
@@ -228,7 +233,7 @@ contract SphinxPluginTypes {
 
     function sphinxConfigType() external view returns (SphinxConfig memory sphinxConfig) {}
 
-    function leafGasParams() external view returns (SphinxTransaction[] memory txnArray) {}
+    function leafGasParams() external view returns (SphinxTransaction[] memory txnArray, SystemContractInfo[] memory systemContracts) {}
```
sam-goldman, line 236: This is the data structure that describes the input parameters to the `sphinxEstimateMerkleLeafGas` function in `Sphinx.sol`, which estimates the Merkle leaf gas fields. We need to include the system contracts because we can't assume that they've already been deployed on a network when this command is called.

packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -983,4 +980,49 @@ contract SphinxUtils is SphinxConstants, StdUtils {
             revert("Sphinx: Invalid MerkleRootStatus. Should never happen.");
         }
     }
+
+    function create2Deploy(bytes memory _initCodeWithArgs) public returns (address) {
```
sam-goldman, line 984: Moved this function and `deploySphinxSystem` below from `SphinxTestUtils` to `SphinxUtils` because these are both required to deploy the system contracts when estimating the Merkle leaf gas fields

packages/core/src/etherscan.ts
```diff
@@ -116,7 +116,7 @@ export const verifyDeploymentWithRetries = async (
       let success = false
 
       const contractName = fullyQualifiedName.split(':')[1]
-      for (let attempt = 0; attempt < maxAttempts; attempt++) {
+      for (let attempt = 1; attempt <= maxAttempts; attempt++) {
```
sam-goldman, line 119: This line fixes the [off-by-one Etherscan log error](https://linear.app/chugsplash/issue/CHU-372/off-by-one-in-etherscan-logs-in-deploy-cli-command).

packages/core/src/languages/solidity/predeploys.ts
```diff
@@ -32,32 +31,24 @@ import {
...
 
 export const ensureSphinxAndGnosisSafeDeployed = async (
-  provider: SphinxJsonRpcProvider,
+  provider: SphinxJsonRpcProvider | HardhatEthersProvider,
+  wallet: ethers.Wallet,
+  executionMode: ExecutionMode,
   relayers: string[] = [],
   logger?: Logger
 ) => {
-  if (!(await isLiveNetwork(provider))) {
```
sam-goldman, line 39: Context: `ensureSphinxAndGnosisSafeDeployed` is now included in the simulation because otherwise the user would need to deploy the system contracts before they can preview their deployment

Since this function is included in the simulation, this `isLiveNetwork` call would sometimes return the incorrect result. The proper solution is to use the execution mode.

packages/core/src/languages/solidity/predeploys.ts
```diff
@@ -32,32 +31,24 @@ import {
...
-      logger
-    )
-  } else if (!(await allSphinxAndGnosisSafeContractsDeployed(provider))) {
-    throw new Error(`Sphinx is not supported on this network.`)
   }
+
+  await deploySphinxSystem(provider, wallet, relayers, executionMode, logger)
 }
 
-const allSphinxAndGnosisSafeContractsDeployed = async (
```
sam-goldman, line 60: abbreviated this name. it's functionally equivalent

packages/core/src/preview.ts
```diff
@@ -9,7 +9,14 @@ import {
   prettyRawFunctionCall,
 } from './utils'
 
-type PreviewElement = DecodedAction | { to: string; data: string }
+type SystemDeploymentElement = {
```
sam-goldman, line 12: Added a new `PreviewElement` for deploying the system contracts. The system contracts are displayed as a single line in the preview

packages/core/src/preview.ts
```diff
@@ -89,16 +98,23 @@ export const getPreviewString = (
...
-        const functionCallStr = isDecodedAction(element)
-          ? prettyFunctionCall(
-              element.referenceName,
-              element.address,
-              element.functionName,
-              element.variables,
-              5,
-              3
-            )
-          : prettyRawFunctionCall(element.to, element.data)
```
sam-goldman, line 101: Needed to refactor this to account for the new `PreviewElement` type, `SystemDeploymentElement`

packages/plugins/script/write-initcode.ts
sam-goldman, line 1: Refactored this file to reuse the `getSystemContractInfo` TypeScript function and the `SystemContractInfo` Solidity struct, which are both used outside of this file

packages/plugins/src/cli/propose/index.ts
```diff
@@ -106,8 +108,15 @@ export const buildParsedConfigArray: BuildParsedConfigArray = async (
       process.exit(1)
     }
 
+    // Check that the system contracts are deployed.
     const provider = new SphinxJsonRpcProvider(rpcUrl)
-    await ensureSphinxAndGnosisSafeDeployed(provider)
+    // This auto-generated wallet is only used for internal testing purposes.
+    const wallet = new ethers.Wallet(getSphinxWalletPrivateKey(0), provider)
```
sam-goldman, line 114: We pass a wallet into the `ensureSphinxAndGnosisSafeDeployed` function in every case to simplify its logic

packages/plugins/src/hardhat/simulate.ts
```diff
@@ -120,11 +120,11 @@ export const simulate = async (
     //    don't hardcode it, Hardhat uses the default hardfork, which is the behavior we want.
 
     // Fast forward the block number. This is necessary to prevent the following edge case:
-    // 1. We deploy the Gnosis Safe and Sphinx contracts.
+    // 1. We execute some transactions on the network.
     // 2. The simulation uses a block that's multiple confirmations behind the latest block. This is
     //    Hardhat's default behavior, which is meant to protect against chain reorgs on forks of
     //    live networks.
-    // 3. The simulation fails because some of the contracts deployed in step 1 don't exist on the
+    // 3. The simulation fails because the transactions executed in step 1 don't exist on the
```
sam-goldman, line 127: Changed this wording slightly because the provided example isn't relevant anymore. (We now deploy the system contracts in the simulation instead of outside the simulation, so step 1 isn't accurate anymore)

packages/plugins/test/mocha/cli/deploy.spec.ts
```diff
@@ -257,7 +257,7 @@ describe('Deploy CLI command', () => {
       // ensures that the signer can execute a deployment without having a very high balance of ETH.
       await setBalance(
         wallet.address,
-        ethers.toBeHex(ethers.parseEther('0.01')),
+        ethers.toBeHex(ethers.parseEther('1.25')),
```
sam-goldman, line 260: Needed to bump the initial amount of ETH in the signer's wallet because it needs to deploy the System contracts. Does this number sound right to you @RPate97? 1 ETH was too low.

packages/plugins/src/cli/propose/index.ts
```diff
@@ -106,8 +108,15 @@ export const buildParsedConfigArray: BuildParsedConfigArray = async (
...
+    await ensureSphinxAndGnosisSafeDeployed(
+      provider,
+      wallet,
+      ExecutionMode.Platform
+    )
```
RPate97, line 119: This will fail when proposing against a live network since the `fundAccountMaxBalance` function uses the `hardhat_setBalance` rpc method. I actually think we can probably just remove the call to this function entirely since in the propose function we can reasonably assume the contracts are already deployed, but I haven't fully thought that through so use your judgement. 

Please also spend a little bit of time testing the proposal task works on live networks.

packages/core/src/preview.ts
```diff
@@ -73,6 +80,8 @@ export const getPreviewString = (
           )
 
           executingArray.push(green(`${i + 1}. ${actionStr}`))
+        } else if (isSystemDeploymentElement(element)) {
+          executingArray.push(green(`${i + 1}. Sphinx & Gnosis Safe Contracts`))
```
RPate97, line 84: Honestly, we should probably be more aggressive with this since the user may have to pay quite a bit to deploy all the Safe contracts, and since our system may not be compatible with the network at all. I'm fine with just opening a ticket for this, though, if you don't want to address it now.

packages/plugins/src/hardhat/simulate.ts
RPate97: Can we just remove this entirely? I thought the main reason we had it was because the contracts were being deployed prior to the simulation.

packages/plugins/test/mocha/cli/deploy.spec.ts
RPate97: Yeah, I think this is reasonable.

RPate97 (commented):
Looks pretty good, but we need to test proposing on a live network since I noticed at least one thing that would prevent that from working.

packages/plugins/src/hardhat/simulate.ts
sam-goldman: I think we need to keep it. I updated the description, which should hopefully clarify how this could happen. Copying and pasting it here:
```
    // Fast forward the block number. This is necessary to prevent the following edge case:
    // 1. Some transactions are executed on the local network. These transactions could either be
    //    sent by the Sphinx team (during testing) or by the user during local development.
    // 2. The Deploy CLI command is executed, which creates a simulation using a block that's
    //    multiple confirmations behind the latest block. This is Hardhat's default behavior, which
    //    is meant to protect against chain reorgs on forks of live networks.
    // 3. The simulation fails because the transactions executed in step 1 don't exist on the
    //    Hardhat fork.
```

README.md
sam-goldman: Appended a new sentence to this bullet point: " This includes networks that the DevOps Platform doesn't currently support."

I also added a note in the "Supported Networks" section since I agree it's relevant there too

> Also one question: Do you think we should list that this only works for projects with a single owner in the current limitations?

Updated the "Current Limitations" section

packages/plugins/src/cli/propose/index.ts
sam-goldman: Nice catch. I removed this call since you're right that it's not necessary. I checked that we can propose with the latest version of this PR on a live network. Also, I'll make a ticket to add an integration test for a live network proposal because this should've been caught by an automated test

packages/core/src/preview.ts
sam-goldman: Yeah you're right. Added a personal note to create a Linear ticket for this

RPate97 (approved):
Looks good

PR #1338, opened 2024-01-13
perf(pg): optimize deployments in proposal test suite

sam-goldman:
Speeds up the proposal integration test suite by ~40%.

Previously, we were calling the Deploy CLI command (i.e. the `deploy` function) inside the proposal test suite to check that the proposal gas estimates were accurate. The `deploy` function is very slow because it invokes multiple Forge scripts. This PR replaces calls to the `deploy` function with calls to the `runEntireDeploymentProcess` function, which serves the same purpose but doesn't invoke any Forge scripts, so it's significantly faster (~25-40x).


packages/plugins/test/mocha/cli/propose.spec.ts
sam-goldman, line 1: The only thing that changed in this file is the `assertValidGasEstimates`, which now calls `runEntireDeploymentProcess` instead of `deploy`. The rest of the changes just refactor the logic that calls into `assertValidGasEstimates`.

RPate97 (approved):
Very nice

PR #1339, opened 2024-01-13
refactor(pg): rename gnosis safe variable name for clarity

sam-goldman:
In `Sphinx.sol`, we prefix all function names with `sphinx` to avoid name collisions with user-defined functions. However, the name `sphinxSafe()` may be confusing to users because it's a Gnosis Safe. I've renamed it to `safeAddress`.


RPate97 (approved):
Looks good

PR #1340, opened 2024-01-14
fix(pg): check if init code belongs to contract artifact

sam-goldman:
Adds logic that checks if a contract's init code belongs to a contract artifact. This change is "headless", i.e. it doesn't update our contract artifact logic. I figured it'd be easier to review a couple smaller PRs instead of a large one.

We use an "exact match" strategy instead of a string distance algorithm like the Levenshtein distance. I chose this strategy mainly because I think it'll be more reliable. There have been a [few](https://github.com/foundry-rs/foundry/issues/4169#issue-1555995415) [reported](https://github.com/foundry-rs/foundry/issues/4169#issuecomment-1679911149) [cases](https://github.com/foundry-rs/foundry/issues/4169#issuecomment-1745335229) of Foundry's diff score not working. I'm sure we could come up with a more reliable diff score, but if we do, our logic won't be battle tested for a while.

This strategy should work in every scenario except in situations where there's more than one contract with identical creation code. This can only happen if the contracts have identical logic and if the metadata hash has been disabled. I don't think any inference strategy would work in this situation, so I think this is acceptable.

Note: In this PR, the `isInitCodeMatch` function uses the entire contract artifact as an input parameter, which means that our cache would need to store every artifact for every build info file. This'd make the cache enormous. The next PR fixes this issue by revising the `isInitCodeMatch` function slightly to use only the necessary info from the artifact.


packages/plugins/test/mocha/foundry/utils.spec.ts
```diff
@@ -558,4 +579,86 @@ describe('Utils', async () => {
       expect(replaceEnvVariables(input)).to.deep.equal(expected)
     })
   })
+
+  describe('isInitCodeMatch', () => {
```
sam-goldman, line 583: Lmk if you want me to add any test cases

packages/plugins/src/foundry/utils/index.ts
```diff
@@ -553,6 +555,87 @@ export const getConfigArtifactForContractName = (
...
+ * 2. The actual init code does not have library placeholders, whereas the artifact's init code may
+ * contain library placeholders. For more info on library placeholders, see:
+ * https://docs.soliditylang.org/en/v0.8.23/using-the-compiler.html#library-linking
+ *
+ * This function does not need to handle immutable variable references because these references only
+ * exist in the runtime bytecode and not the init code. This is made clear by the Solidity docs,
+ * which only have a `deployedBytecode.immutableReferences` field:
+ * https://docs.soliditylang.org/en/v0.8.23/using-the-compiler.html#library-linking
+ */
+export const isInitCodeMatch = (
```
sam-goldman, line 577: This is the function that checks for a match

RPate97 (approved):
Looks good, nice test suite

PR #1341, opened 2024-01-14
fix(pg): use bytecode inference logic to determine contract artifact

sam-goldman:
This PR updates the `makeGetConfigArtifacts` function to use the `isInitCodeMatch` function instead of the Foundry broadcast file's `contractName` field.

A few notes:
- Run `rm -rf <cacheDir>` in every repo where you've executed a Sphinx deployment (the plugins package, demo package, etc). This removes the `sphinx-cache.json` file, which is necessary because this PR changes the cache file content.
- We test that `getConfigArtifacts` works properly in the existing `artifacts.spec.ts` test suite.
- This PR doesn't remove the `sphinxLabel` code and docs. That'll be in a separate PR.
- A downside to this PR is that we need to cache more information than the fully qualified names in the `sphinx-cache.json` file. Specifically, we need to cache all of the inputs to the `isInitCodeMatch` function. The `isInitCodeMatch` function previously required that we pass in the entire artifact, which meant that we would've needed to store every artifact from every build info file in our cache. This would've made the cache file size enormous, so I changed the `isInitCodeMatch` function to use only the required info from the artifact: the `bytecode`, the constructor fragment (if it exists), and the `linkReferences`. Caching these three fields resulted in a `sphinx-cache` file size of 2.3 MB on the first compiler run in the plugins package (which contains ~100 contracts). The cache file size will scale proportionately with the number of build info files, which means that in the worst case, the cache could probably be hundreds of MBs. Since the vast majority of the file's size comes from the `bytecode`, we can reduce its size even more by storing just the hash of the bytecode and its length. This'd make the file's size comparable to its original structure, which only contained fully qualified names. I figured that this optimization is beyond the scope of this PR, so I'm going to make a ticket for it.


packages/core/src/config/types.ts
```diff
@@ -148,6 +148,7 @@ export type SphinxConfig = {
 export interface RawCreate2ActionInput extends SphinxTransaction {
   contractName: string | null
   create2Address: string
+  initCodeWithArgs: string
```
sam-goldman, line 151: Added this field to the `RawCreate2ActionInput` type because it provides a convenient way to aggregate all of the init code, which we'll pass into `getConfigArtifacts`

packages/plugins/src/foundry/utils/index.ts
```diff
@@ -382,24 +382,34 @@ export const makeGetConfigArtifacts = (
...
 
-    const fullyQualifiedNamePromises = fullyQualifiedNames.map(
-      async (fullyQualifiedName) => {
-        const artifact = await readContractArtifact(
-          fullyQualifiedName,
-          projectRoot,
-          artifactFolder
-        )
-
+    const fullyQualifiedNamePromises = initCodeWithArgsArray.map(
```
sam-goldman, line 385: This section works the same as before, except:
1. We now iterate through the input `initCodeWithArgsArray` instead of the input fully qualified names, and
2. We search for a matching artifact using `isInitCodeMatch` instead of searching for the most recent build info file that includes the fully qualified name.

packages/plugins/src/foundry/utils/index.ts
```diff
@@ -576,10 +546,13 @@ export const getConfigArtifactForContractName = (
  */
 export const isInitCodeMatch = (
   actualInitCodeWithArgs: string,
-  artifact: ContractArtifact
+  artifact: {
+    bytecode: string
+    linkReferences: LinkReferences
+    constructorFragment?: ethers.ConstructorFragment
+  }
```
sam-goldman, line 553: Changed this function to accept a minimal amount of artifact-related info instead of the entire artifact. The functionality remains the same.

packages/plugins/src/foundry/utils/index.ts
```diff
@@ -382,80 +386,48 @@ export const makeGetConfigArtifacts = (
...
         }
-
-        // Throw an error if no build info file is found in the cache for this contract
-        // This should only happen if the user manually deletes a build info file
-        if (existsSync(buildInfoCacheFilePath)) {
-          unlinkSync(buildInfoCacheFilePath)
-        }
-        throw new Error(
-          `Build info cache is outdated, please run 'forge build --force' then try again.`
-        )
```
sam-goldman, line 416: It's necessary to remove this in order to support contracts that are defined as inline bytecode, e.g. a `CREATE3` proxy. If we don't remove these lines, [this test case for `getConfigArtifacts`](https://github.com/sphinx-labs/sphinx/pull/1341/files#diff-36de83659b4192824ef2b7a81238450fd670bbe6aadc70b51e979cd332c7c007R601-R609) will throw an error. If the user deletes a build info file, we'll display a warning in the preview saying that we couldn't identify these contracts, which is reasonable I think. If you think it'll be common for users to run into this issue, we could add something to the troubleshooting guide or the warning in the preview

packages/core/src/utils.ts
```diff
@@ -1499,3 +1500,12 @@ export const zeroOutLibraryReferences = (
...
   return modifiedBytecode
 }
+
+/**
+ * Type guard to check if a value is not undefined.
+ *
+ * @param value The value to check.
+ * @returns true if the value is not undefined, false otherwise.
+ */
+export const isDefined = <T>(value: T | undefined): value is T =>
```
RPate97, line 1510: nice

packages/plugins/src/foundry/utils/index.ts
RPate97: I think this is acceptable.

RPate97 (approved):
Looks good. Thank you for splitting up these PRs, definitely made reviewing easier.

PR #1343, opened 2024-01-15
fix(core): remove `CREATE3` proxies from preview warning

sam-goldman:
Resolves this ticket: ["Skip `CREATE3` proxy warning in preview"](https://linear.app/chugsplash/issue/CHU-385/skip-create3-proxy-warning-in-preview)

I manually checked that this PR fixes the issue in Flood's repo.

The main update in this PR is that we change the parsed config's `unlabeledAddresses` array (which just contains unlabeled addresses) to an `unlabeledContracts` array (which contains the unlabeled addresses and their associated init code). In the preview, we filter out any unlabeled address with init code that matches the `CREATE3` proxy's init code.

Even though we now remove `CREATE3` proxies from the preview warning, we still include them in the [`unlabeledContracts` array in the user's `ExecutionArtifact`](https://github.com/sphinx-labs/sphinx/pull/1343/files#diff-638134161b04c7cde57553b0bab421b2ba1c2e83052b5937fc95890d2a57063cR349). I figured the execution artifact should include every unlabeled contract for completeness, even if it's a `CREATE3` proxy.


docs/deployment-artifacts.md
```diff
@@ -320,7 +320,10 @@ type ExecutionArtifact = {
     isModuleDeployed: boolean
     isExecuting: boolean
   }
-  unlabeledAddresses: Array<string>
+  unlabeledContracts: Array<{
+    address: string
+    initCodeWithArgs: string
+  }>
```
sam-goldman, line 326: The `ExecutionArtifact` now includes each unlabeled contract's `initCodeWithArgs` in addition to its address. This'll make it easier to verify unlabeled contracts after a deployment is completed. This change isn't related to the main focus of the PR, but it was natural to add it because the `ParsedConfig`, which is one of the main inputs to the `ExecutionArtifact`, now contains the same `unlabeledContracts` array. (The `ParsedConfig` needs the `initCodeWithArgs` so that we can use it to filter out the `CREATE3` proxy initcode in the preview).

RPate97 (approved):
Looks good

PR #1344, opened 2024-01-15
feat(pg): Enable deployment artifact cli task

RPate97:
## Purpose 
Adds support for fetching deployment artifacts from the website.


packages/core/src/utils.ts
```diff
@@ -164,10 +164,7 @@ export const getGasPriceOverrides = async (
     throw new Error(`Unable to retrieve latest block.`)
   }
 
-  if (
-    executionMode === ExecutionMode.LocalNetworkCLI ||
-    executionMode === ExecutionMode.Platform
```
RPate97, line 169: Just noticed this bug when testing with the website. We cannot use `ExecutionMode.Platform` b/c it will result in the gas limit being extremely high for the production executor, which will cause transactions to fail due to insufficient funds. 

I'm not sure if there was a specific reason why you included this, though, so let me know if you think some additional changes need to be made related to this.

packages/core/src/utils.ts
```diff
@@ -504,27 +501,33 @@ export const storeCanonicalConfig: StoreCanonicalConfig = async (
...
-    orgId,
-    configData,
-  })
+  } = await axios
+    .post(`${fetchSphinxManagedBaseUrl()}/api/pin`, {
+      apiKey,
+      orgId,
+      configData,
+    })
+    .catch((err) => {
```
RPate97, line 510: Just refactored this to match the pattern I used for the call to fetch the artifacts.

packages/core/src/networks.ts
```diff
@@ -69,7 +69,7 @@ export const fetchCurrencyForNetwork = (chainId: bigint) => {
   const network = SPHINX_NETWORKS.find((n) => n.chainId === chainId)
 
   if (network) {
-    return network.name
+    return network.currency
```
RPate97, line 72: Just noticed this while testing the website.

packages/plugins/src/cli/setup.ts
```diff
@@ -65,58 +69,32 @@ export const makeCLI = (
...
+          })
+          .option('project-name', {
+            describe: 'The name of your project.',
+            type: 'string',
+            demandOption: true,
+          })
+          .option('silent', {
+            describe: 'Silence the output except for error messages.',
+            boolean: true,
+            default: false,
```
sam-goldman, line 93: If you want to add this as a user-facing flag, you should update the deployment artifact docs to include it. Also, I think you should update the usage section above, which is currently:
```
          .usage(
            `Usage: sphinx artifacts --org-id <ORG_ID> --project-name <PROJECT_NAME>`
          )
```

sam-goldman (commented):
Looks great! Just a couple minor things

packages/core/src/utils.ts
RPate97: I'd like to avoid regressions like this, and I think there's a straightforward solution so I've added an environment variable `SPHINX_INTERNAL__HARDCODE_GAS_LIMIT`. It can be used to enable this optimization. I recommend setting it explicitly in test files. I've done so in both `artifacts.spec.ts` and `propose.spec.ts`.

sam-goldman: I don't think that resolves the issue because we need the optimization enabled in the *production* proposal process (i.e. during the user's proposal). The user's proposal process is really slow without the optimization. It seems like your solution would only enable the optimization when we're testing the proposal process.

Two options come to mind:
1. The simplest solution may be to use the original if-statement that existed before this PR, and add an env variable that the executor can specify to *disable* the optimization. (Instead of requiring an environment variable to *enable* the optimization everywhere else).
2. I think the proper but more involved solution is to add another `ExecutionMode` for the production executor. I think this may be a more robust solution because otherwise we're operating under the assumption that the proposal simulation and the production executor share all of the same logic (except for this single if-statement, which would be resolved via env variable). Fwiw, when I was writing the logic for the `Platform` execution mode, I was only thinking about the proposal simulation and not the production executor. There could be other errors that arise because of this oversight on my part

I think option 1 is a reasonable solution for this PR given that you've recently tested that the executor works. If you choose option 1, one of us should make a ticket for option 2. Again, ping me on Telegram if you want me to make that ticket

sam-goldman (commented):
Everything LGTM, but we need to figure out a solution to this gas price overrides issue

sam-goldman (approved):
LGTM!

PR #1345, opened 2024-01-17
fix(core): Fix live network deployment bugs

RPate97:
## Purpose
Fixes a few issues I noticed doing live network testing with the latest changes.
- Fixes some drip sizes which were incorrectly updated to values that don't reflect the real on-chain value
- Implements the recommended gas calculation strategy for linea. See [this article](https://docs.linea.build/build-on-linea/gas-fees) for more information.


packages/contracts/src/networks.ts
```diff
@@ -146,7 +146,7 @@ export const SPHINX_NETWORKS: Array<SupportedNetwork> = [
       envKey: 'POLYGON_ETHERSCAN_API_KEY',
     },
     currency: 'MATIC',
-    dripSize: '1',
+    dripSize: '0.15',
```
RPate97, line 149: I accidentally updated this such that it no longer matches the real on-chain value which causes problems.

RPate97, line 149: I accidentally updated this such that it no longer matches the real on-chain value which causes problems. This change just switches it back to the correct value.

PR #1349, opened 2024-01-18
fix(core): discard git message when user is not in a git repository

sam-goldman:
Fixes an issue where the following line was displayed to the user if they aren't in a git repository:
> "fatal: not a git repository (or any of the parent directories): .git".

I manually checked that there this error message isn't displayed in a repo that doesn't have git


RPate97 (approved):
Looks good

PR #1350, opened 2024-01-18
fix(pg): Resolve linea transaction hang

RPate97:
## Purpose
Sets the gas price to 10x the recommended value on linea goerli to [workaround this bug](https://github.com/ethers-io/ethers.js/issues/4533).

### DIFF
--- a/packages/core/src/utils.ts
+++ b/packages/core/src/utils.ts
@@ -199,13 +199,19 @@ export const getGasPriceOverrides = async (
     case 1442:
     case 1101:
       return overridden
-    // On linea and its testnet, just override the gasPrice
+    // On linea testnet, 10x the gas price because of this bug:
+    // https://github.com/ethers-io/ethers.js/issues/4533
     case 59140:
+      if (gasPrice !== null) {
+        overridden.gasPrice = gasPrice * BigInt(10)
+        return overridden
+      }
+    // On linea, override the gasPrice
     case 59144:
       if (gasPrice !== null) {
         overridden.gasPrice = gasPrice
-        return overridden
       }
+      return overridden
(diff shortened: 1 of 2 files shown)


packages/core/src/utils.ts
```diff
@@ -199,13 +199,19 @@ export const getGasPriceOverrides = async (
     case 1442:
     case 1101:
       return overridden
-    // On linea and its testnet, just override the gasPrice
+    // On linea testnet, 10x the gas price because of this bug:
+    // https://github.com/ethers-io/ethers.js/issues/4533
     case 59140:
+      if (gasPrice !== null) {
```
sam-goldman, line 205: Why do we only override the `gasPrice` when it's not `null`? Some docs would be nice

sam-goldman (approved):
One comment, otherwise LGTM

packages/core/src/utils.ts
RPate97: It just prevents a `gasPrice` value passed in by the caller from being overridden if the `gasPrice` returned by `getFeeData` is `null`. Honestly, this probably isn't necessary given that we generally never pass in our own values for `gasPrice`, `maxFeePerGas`, or `maxPriorityFeePerGas`. Even if we did, I'm not sure I see the use case for including `gasPrice` on a network where it isn't supported. I've mainly continued to do this b/c it was the original pattern implemented by Kelvin for this function. 

If we want to remove this, I would do it for the whole function.

RPate97: I also made a ticket for this.

sam-goldman: Sounds good

PR #1355, opened 2024-01-20
feat(ct): Rearrange contracts for standard forge installation

RPate97:
## Purpose
This PR rearranges the production contracts to support installing them via standard forge installation. 

## Overview
This is the first in a set of PRs related to better support for mono repos. I'm going to use this PR description to provide a high-level description of the entire change. At the bottom, I've included a list of the separate PRs I expect to open related to this.

The general goal of this is to remove the need for the user to define remappings for our Solidity library contracts which point to difficult-to-predict file locations that depend on their particular package manager and/or mono repo structure. This will make it easier to install our library in monorepos, and will make it easy to maintain over time.  

The high-level idea behind this change is to move to a dual installation strategy where we distribute our sphinx CLI as exclusively a typescript CLI. Then the user must install the sphinx library contracts separately using forge install. In practice, the installation and version management of the Sphinx library contracts will be managed internally by the Sphinx CLI. I.e installing them in the first place during setup, ensuring the contract library version installed is compatible with the version of the Sphinx CLI, etc. This is actually very similar to how Foundry manages its own installation process since it deals with a similar problem.

## User installation process
After this is fully implemented, the installation process will look roughly like this:
1. Install the Sphinx CLI `yarn add -D @sphinx-labs/plugins`
4. Initialize sphinx `npx sphinx init`
- During the `init` command, our CLI runs the following command to install the Sphinx Library contracts: `forge install sphinx-labs/sphinx@vx.y.z`
5. Copy a single remapping to the `SphinxPlugin.sol` file into their config file which points to `lib/sphinx/...`
6. Rest of the installation process (i.e other configuration options, integrating with their script, etc)

## Managing Sphinx Library contract versions
Since we're installing the Sphinx Library contracts as a git submodule using forge install instead of packaging them with our CLI, we will have to deal with the possibility of version drift. That is to say, it's possible for the version of the Sphinx Library contracts that is installed to be incompatible with the version of the CLI that is installed. This will typically occur when the user bumps their `@sphinx-labs/plugins` package version, but it can also occur runs `forge update` unnecessarily. To resolve this issue we will implement a function in `Sphinx.sol`, `getSphinxLibraryVersion()`. This function will return the current version of the `contracts` package. Then in our CLI, we'll add checks to detect if the current contract library version is incorrect and automatically synchronize the versions. The source of truth will always be the current version of the Sphinx CLI.

## Expected PRs
1. (This PR) Moving all prod contracts to the contracts package and installing forge-std using the standard forge installation method in the contracts package. This is the bulk of the changes required for monorepo support.
2. Update the release process to make a new GitHub tag whenever we release a new version of the contracts package. This tag needs to be compatible with forge install (no `@`). Also create a constant in the `@sphinx-labs/plugins` package which corresponds to the current plugins package version. This constant must be bumped during the release process by hooking into it using a [custom version script](https://github.com/changesets/action?tab=readme-ov-file#with-version-script). 
3. Add a new command `sphinx install` which installs the required Sphinx contracts library version. Add logic to the propose and deploy task to detect if the Sphinx contracts library is out of date and error if so.
4. Update the `init` command so that it also installs the Sphinx library contracts using forge install and so that it makes sense to use when integrating an existing project. Alternatively, create a new command that just installs the correct Sphinx Library version if the init command is to cumbersome to be used in existing repos.
5. Update docs where necessary


.circleci/config.yml
```diff
@@ -12,6 +12,11 @@ jobs:
           keys:
             - v2-cache-source-{{ .Branch }}-{{ .Revision }}
             - v2-cache-source-{{ .Branch }}
+      - run:
+          name: Initialize submodules
+          command: |
+            git submodule init
+            git submodule update
```
RPate97, line 19: Allows git submodules to be installed during the circle ci checkout step

.circleci/config.yml
```diff
@@ -66,18 +71,11 @@ jobs:
...
-            or:
-              - equal: [ "develop", << pipeline.git.branch >> ]
-              - not:
-                  equal: [ "contracts", <<parameters.package_name>> ]
-          steps:
-            - run:
-                name: Test
-                command: yarn test:coverage
-                no_output_timeout: 20m
-                working_directory: packages/<<parameters.package_name>>
```
RPate97, line 80: Since this PR moves a lot of logic related to our library into the contracts package, including the tests, I decided to reenable the contracts tests. I've created [a ticket to look into specifically disabling the tests for the core contracts](https://linear.app/chugsplash/issue/CHU-414/investigate-disabling-unnecessary-tests-in-the-contracts-package-when).

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "packages/contracts/lib/forge-std"]
+	path = packages/contracts/lib/forge-std
+	url = https://github.com/foundry-rs/forge-std
```
RPate97, line 3: We now install forge-std as a git submodule as recommended by foundry. When the user installs us in their project using `forge install`, our dependency on forge-std will also be installed and our contracts will have everything they need to compile properly.

packages/contracts/contracts/foundry/Sphinx.sol
```diff
@@ -1,13 +1,10 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.0;
 
-import { VmSafe, Vm } from "sphinx-forge-std/Vm.sol";
+import { VmSafe, Vm } from "../../lib/forge-std/src/Vm.sol";
```
RPate97, line 4: Note how we import forge-std using a relative path instead of using a remapping. This is because any remapping we might define related to this would also be required in the user repo. So, I decided not to use remappings at all in the `contracts` package. 

TODO
See this comment for info on how we enforce this to prevent regressions.

packages/contracts/test/mocha/remappings.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * This tests that we have no remapping defined for forge-std. We do this because users install our library in their repo and then import
+ * our contracts in their scripts. So any remapping used in contracts that will be imported by the user would also need to be defined
+ * by them. We chose not to use a remapping for forge-std at all, and we enforce that with the test below.
+ *
+ * In the future if we add additional libraries which are imported by our production contracts, we should add additional tests here
+ * to enforce that they do not rely on remappings which may not be defined in the users project.
+ */
+describe('Remappings', () => {
```
RPate97, line 13: See comment

packages/contracts/test/SphinxTestUtils.sol
```diff
@@ -147,7 +156,7 @@ contract SphinxTestUtils is SphinxConstants, StdCheatsSafe, SphinxUtils, SphinxI
             string(abi.encodePacked(key, ".arguments"))
         );
         string[] memory arguments = argumentsBytes.length == 32 &&
-            bytes32(argumentsBytes) == bytes32(0)
+            toBytes32(argumentsBytes) == bytes32(0)
```
RPate97, line 159: We have to use the above `toBytes32` function to workaround a limitation where bytes cannot be implicitly converted to bytes32. This wasn't an issue previously because were were compiling with `0.8.15`. But now we're compiling everything with `0.8.4`, so we have to implement a workaround. 

Originally, I was just going to have the core contracts compile with `0.8.4` and then compile everything else with `0.8.15` which would have resolved that, but I found things didn't immediately work that way and it seemed relatively insignificant to just implement this workaround instead.

packages/contracts/test/SphinxUtils.t.sol
```diff
@@ -121,10 +121,4 @@ contract SphinxUtils_Test is Test, SphinxUtils {
 
         assertEq(uniqueValues.length, 0, "The returned array should be empty");
     }
-
-
-    function test_networkEnumSize_success() external {
-        uint256 expected = uint8(type(Network).max) + 1;
-        assertEq(numSupportedNetworks, expected);
-    }
```
RPate97, line 129: This is unrelated to the PR. I just realized we don't need to test this anymore since the networks Enum is autogenerated. I noticed and removed it b/c I also encountered an issue with this test related to compiling with `0.8.4`.

packages/plugins/src/cli/setup.ts
```diff
@@ -135,32 +131,9 @@ export const makeCLI = (
...
           .hide('version'),
       async (argv) => {
-        const { pnpm, foundryup, orgId, sphinxApiKey, alchemyApiKey, owner } =
-          argv
-
-        init(pnpm, foundryup, orgId, sphinxApiKey, alchemyApiKey, owner)
-      }
-    )
-    .command(
-      'remappings',
```
RPate97, line 145: We don't really need this anymore because the remapping the user must configure is very simple. So we'll just have them copy paste it from the getting started guide.

packages/contracts/contracts/foundry/SphinxConstants.sol
```diff
@@ -6,11 +6,11 @@ import { NetworkInfo, NetworkType } from "./SphinxPluginTypes.sol";
...
   address public constant compatibilityFallbackHandlerAddress = 0xf48f2B2d2a534e402487b3ee7C18c33Aec0Fe5e4;
   address public constant multiSendAddress = 0xA238CBeb142c10Ef7Ad8442C6D1f9E89e07e7761;
-  address public constant sphinxModuleProxyFactoryAddress = 0x8f3301c9Eada5642B5bB12FD047D3EBb2932E619;
-  address public constant managedServiceAddress = 0xB5E96127D417b1B3ef8438496a38A143167209c7;
+  address public constant sphinxModuleProxyFactoryAddress = 0xb3C4df24Ac0e21F9436F9A6A6BB4F7566872D003;
+  address public constant managedServiceAddress = 0xE1FC160c3037553ec654822653681381a68AB4EF;
   address public constant safeFactoryAddress = 0xa6B71E26C5e0845f74c812102Ca7114b6a896AB2;
   address public constant safeSingletonAddress = 0xd9Db270c1B5E3Bd161E8c8503c55cEABeE709552;
-  address public constant sphinxModuleImplAddress = 0x8f4E4d51B8050B0ff713eff1F88f3dD8b5e8a530;
+  address public constant sphinxModuleImplAddress = 0x3b148245BB8b75cF0FA501688bcE9Cbd0076Af4f;
```
RPate97, line 13: Curious to hear your thoughts on this. These addresses should not change b/c I did not change any of the core contract code. We also can't merge this PR with different addresses for these contracts b/c they will not correspond to the real deployed contracts. 

I was not able to figure out exactly why these changed. Have any idea?

.changeset/eighty-panthers-kneel.md
```diff
@@ -0,0 +1,7 @@
+---
+'@sphinx-labs/contracts': minor
+'@sphinx-labs/plugins': minor
+'@sphinx-labs/core': minor
+---
+
+Support monorepo installation
```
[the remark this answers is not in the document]
sam-goldman: Sounds good. Made a [ticket](https://linear.app/chugsplash/issue/CHU-415/consider-updating-package-versions-for-existing-integrations) to consider updating the package version used by existing integrations

packages/contracts/test/SphinxTestUtils.sol
sam-goldman: Sounds good. Could you include that in the docs for `toBytes32`?

packages/contracts/test/mocha/remappings.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+    // a remapping for forge-std will be included automatically. So we need to set it to specifically
+    // an empty array to get the behavior we would like:
+    // No remappings defined for forge-std.
+    expect(libs.length, 'libs should not be defined').to.eq(0)
+
+    // Expect there to be no remapping for forge-std
+    expect(
+      remappings.some((remapping) => remapping.includes('lib/forge-std')),
+      'remapping detected for forge-std'
+    ).to.eq(false)
```
sam-goldman, line 31: Are there supposed to be any elements in the `remappings` array? If so, could you document why? If not, should this test be modified to check that `remappings.length` equals 0?

packages/contracts/contracts/foundry/Sphinx.sol
```diff
@@ -1,13 +1,19 @@
...
-import { ISphinxModule } from "@sphinx-labs/contracts/contracts/core/interfaces/ISphinxModule.sol";
+// We chose not to use any remappings when importing the standard forge library. This is because when our library is installed in
+// the users project we will be subject to their configured remappings. Bugs can also occur if we rely on the users installation of
+// forge-std which may not be the same exact version our library expects. To resolve both of these issues, we install the version of
+// forge-std we need ourself. We then reference it using a relative import instead of a remapping because that prevents the user from
+// having to define a separate remapping just for our installation of forge-std.
+//
+// Technically, we could use a forge-std remapping in our non-production contracts (scripts, tests, etc). We chose not too because allowing
+// that would increase the chance that we accidentally ship something that may not compile in the users project because we accidentially
+// relied on a remapping in the wrong file.
```
sam-goldman, line 12: It's not clear what you mean here. Does the user compile our non-production contracts (scripts, test, etc)? Regardless, this doesn't seem relevant to `Sphinx.sol`

sam-goldman (commented):
Description was a joy to read. Was concise and gave the right amount of context to make it easy to review the PR

PR looks great. Just had some minor comments and a couple general questions

sam-goldman:
A couple general questions after reading the description:

> 5. Copy a single remapping to the SphinxPlugin.sol file into their config file which points to lib/sphinx/...

Will we require the user to copy and paste the remapping in the quickstart guide, or just the Existing Project guide? For context, the quickstart guide assumes that the user is in a new repository, which currently allows us to create a foundry.toml that includes the correct remappings. This is nice UX because the user doesn't need to manually copy and paste any remappings into their foundry.toml

> ...in our CLI, we'll add checks to detect if the current contract library version is incorrect and automatically synchronize the versions.

When you say "automatically synchronize the versions", do you mean that we'll update the user's dependency version without their explicit permission? If so, is it an accepted practice to change the user's dependency versions without their explicit approval? Maybe it'd be more appropriate to throw an error

sam-goldman:
Also big fan of documenting design decisions in the code

packages/contracts/contracts/foundry/Sphinx.sol
RPate97: The user doesn't compile the non-production contracts which is why it technically is possible to use remappings in them. I've moved this comment to `Sphinx.t.sol` where it is more relevant.

packages/contracts/test/mocha/remappings.spec.ts
RPate97: Yeah, there should be 4 related to the core contracts and their tests.

RPate97:
> Will we require the user to copy and paste the remapping in the quickstart guide, or just the Existing Project guide? For context, the quickstart guide assumes that the user is in a new repository, which currently allows us to create a foundry.toml that includes the correct remappings. This is nice UX because the user doesn't need to manually copy and paste any remappings into their foundry.toml

Just the existing project guide.

RPate97:
> When you say "automatically synchronize the versions", do you mean that we'll update the user's dependency version without their explicit permission? If so, is this a commonly accepted practice? It seems potentially invasive. It's also not clear whether we should update the plugins version, or update the contracts version, or update whichever has the earlier version. Maybe it'd be more appropriate to throw an error.

So the user will have our mono repo installed in their project as a git submodule using forge install. They will also have the `@sphinx-labs/plugins` package installed via their package manager. When the user goes to bump the dependency version for our `plugins` package in their package.json file, the CLI will get updated, but the submodule will not. So we have to detect that it is out of date and somehow ensure it is updated. I was going just to have the CLI detect that and automatically update the submodule so the source of truth is always the version they have in their package.json file. I'd be alright with just erroring and forcing the user to do it themselves if you think that's less invasive though.

Thinking about it more, I think a stronger argument for forcing the user to do it is that when they update the submodule they also need to commit that change in their repo. So forcing them to do it might be better for ensuring that part is done correctly.

sam-goldman:
> I think what I'll do it just make a new CLI command `sphinx install` which handles installing the correct version and prompts the user to commit the changes to version control. Then if we detect version, drift we can just error and prompt them to run that command.

I like the sound of that

sam-goldman (approved):

PR #1356, opened 2024-01-21
fix(ct): Permanently store core contract artifacts

RPate97:
## Purpose
Updates our deployment process and address generation functions to use a static set of contract artifacts which are known to be correct.

## Context
Previously, we relied on the forge compiler to generate the contract artifacts that we use for our live network deployments. This is problematic because it is quite easy to accidentally update settings, which might result in generating artifacts that are slightly different and result in different addresses when the contracts are deployed. 

I noticed this while working on the mono repo support changes, where I noticed that the addresses generated in SphinxConstants changed despite my not making any direct changes to the contracts themselves or significantly modifying the compiler settings. 

I went back through the commit history and noticed several times that the addresses in SphinxConstants have deviated from the correct addresses. So to prevent this from ever happening, I thought it best to retrieve the original artifacts and store them statically in version control instead of relying on the compiler to generate them.


sammyg123456 (approved):
Looks good!

PR #1357, opened 2024-01-21
feat(ct): Add contract library version and release flow

RPate97:
## Purpose
Adds the release process for the Solidity contract library. 

### This includes:
- Updating the changeset script to bump a contracts version value stored in both solidity and typescript in the contracts package.
- Updating the release process to create a new Github tag whenever a release is made for the contracts package.


.circleci/config.yml
```diff
@@ -12,12 +12,12 @@ jobs:
           keys:
             - v2-cache-source-{{ .Branch }}-{{ .Revision }}
             - v2-cache-source-{{ .Branch }}
+      - checkout
       - run:
           name: Initialize submodules
           command: |
             git submodule init
             git submodule update
-      - checkout
```
RPate97, line 20: This isn't directly related to the PR, but was blocking this from building. These need to be swapped b/c `git submodule init` will only work if there is a git repo already setup. This is only the case if CircleCI is able to load up one from the cache. Swapping these makes it always work b/c the repo will be checked out if it's not loaded from cache.

.github/workflows/release.yml
```diff
@@ -49,6 +49,24 @@ jobs:
         id: changesets
         with:
           publish: yarn changeset publish
+          version: yarn changeset:version:custom
```
RPate97, line 52: Overrides the standard `changeset version` command with our own. You can test this out by running it in the root of the mono repo.

.github/workflows/release.yml
```diff
@@ -49,6 +49,24 @@ jobs:
...
+            const publishedPackages = JSON.parse(process.env.PUBLISHED_PACKAGES);
+            const contractsPackage = publishedPackages.find((pkg) => pkg.name === '@sphinx-labs/contracts');
+            if (contractsPackage) {
+              github.rest.git.createRef({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                ref: `refs/tags/v${contractsPackage.version}`,
+                sha: context.sha
+              });
+            }
```
RPate97, line 72: Creates a new GitHub tag in the format `vx.y.z` whenever a new release of the contracts package happens. This is somewhat difficult to fully test locally, but I'm quite confident it will work as I have tested the JS script that creates the tag.

packages/contracts/scripts/write-version.ts
```diff
@@ -0,0 +1,13 @@
...
+/**
+ * Generates a Typescript constant for the contracts package version which is used by the plugins package to determine what
+ * the correct version of the Sphinx library contracts is.
+ */
+const writeVersion = async () => {
+  const file = `export const CONTRACTS_LIRARY_VERSION = 'v${version}'\n`
+  process.stdout.write(file)
+}
+
+writeVersion()
```
RPate97, line 13: This just generates a TS constant for the contracts package version, so we can easily fetch it in the `plugins` package.

packages/contracts/scripts/write-constants.ts
```diff
@@ -57,6 +60,7 @@ const writeConstants = async () => {
     `pragma solidity >=0.6.2 <0.9.0;\n\n` +
     `import { NetworkInfo, NetworkType } from "./SphinxPluginTypes.sol";\n\n` +
     `contract SphinxConstants {\n` +
+    `  string public constant sphinxVersion = 'v${version}';\n` +
```
RPate97, line 63: Including this here makes it easy for us to fetch the current version of the Solidity contracts library.

packages/contracts/src/version.ts
```diff
@@ -0,0 +1 @@
+export const CONTRACTS_LIRARY_VERSION = 'v0.15.2'
```
RPate97, line 1: In the case of version drift, we would expect this value to be different from the value returned by `sphinxLibraryVersion()`.

packages/contracts/contracts/foundry/Sphinx.sol
```diff
@@ -266,6 +266,10 @@ abstract contract Sphinx {
         return sphinxUtils.getGnosisSafeProxyAddress(sphinxConfig);
     }
 
+    function sphinxVersion() public pure returns (string) {
+        return sphinxUtils.getSphinxVersion();
+    }
```
RPate97, line 271: Will be used by the CLI to detect the installed version of the Sphinx library contracts.

packages/contracts/scripts/write-constants.ts
RPate97, line 63: We also store this in Solidity so we don't have to read it from the filesystem.

packages/contracts/scripts/write-version.ts
```diff
@@ -0,0 +1,13 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const { version } = require('../package.json')
```
sam-goldman, line 2: nit: Can we use the standard import syntax here? i.e. `import { version } from '../package.json'`

packages/contracts/scripts/write-constants.ts
```diff
@@ -10,6 +10,9 @@ import {
 } from '../src/addresses'
 import { SPHINX_NETWORKS, SPHINX_LOCAL_NETWORKS } from '../src'
 
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const { version } = require('../package.json')
```
sam-goldman, line 14: nit: Can we use the standard import syntax here? i.e. `import { version } from '../package.json'`

packages/contracts/scripts/write-constants.ts
sam-goldman, line 63: Why is it necessary to store the contracts version in Solidity (in `sphinxVersion` here) and in TypeScript (in `contracts/src/version.ts`)? In other words, can't we just check the contracts version by doing this in the plugins package?

```ts
import { CONTRACTS_LIBRARY_VERSION } from '@sphinx-labs/contracts'

if (expectedContractsLibraryVersion !== CONTRACTS_LIRARY_VERSION) throw new Error(...)
```

packages/contracts/scripts/write-version.ts
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * Generates a Typescript constant for the contracts package version which is used by the plugins package to determine what
+ * the correct version of the Sphinx library contracts is.
+ */
+const writeVersion = async () => {
+  const file = `export const CONTRACTS_LIRARY_VERSION = 'v${version}'\n`
```
sam-goldman, line 9: > CONTRACTS_LIRARY_VERSION

`LIRARY` -> `LIBRARY`

sam-goldman (commented):
It's difficult to review this PR without knowing why it's necessary to store the contracts version in Solidity (in `sphinxVersion`) and in TypeScript (in `contracts/src/version.ts`). Can you explain why both are necessary, and how we plan to use both?

RPate97:
> It's difficult to review this PR without knowing why it's necessary to store the contracts version in Solidity (in `sphinxVersion`) and in TypeScript (in `contracts/src/version.ts`). Can you explain why both are necessary, and how we plan to use both?

So to be able to detect version drift, we need to be able to get the current version of the Sphinx contracts library. We also need to be able to retrieve the version expected by the Sphinx CLI. The CLI will fetch both of these values and compare them to determine if the current installation of the contracts library is out of date. The version expected by the Sphinx CLI is stored in typescript and can simply be used because it comes bundled along with the rest of the CLI. 

The current version of the Sphinx contracts library is different. It does not come bundled with the CLI. It must be stored with the contracts, which are installed via forge install. We need a method for the Sphinx CLI to retrieve it. There are many ways this could be done, but in my opinion, the cleanest way to do it is to retrieve it by calling a Solidity function, which I've implemented in this PR [`sphinxLibraryVersion()`](https://github.com/sphinx-labs/sphinx/pull/1357/files#diff-4e2c1c8b1111f58e9ea0a18c238d2a8e5cafe68c828209d5226bc9e3b9dfc3a2R269). This provides a nice interface for fetching the current version of the Sphinx contracts library that fits into the current pattern of how we retrieve information on the users project (i.e we could add this as a field returned by the `sphinxConfigABIEncoded` function).

For context, with this entire feature I am trying to follow a convention that the installation of the sphinx contracts via forge install should be used _exclusively_ for Solidity code. The reason I am following this convention is that it creates a very clear delineation between what should be packaged with the Sphinx CLI and with the Sphinx library contracts, respectively. I believe this will be important for the maintainability of this setup. Furthermore, this would allow us to make future optimizations to the way we distribute the Sphinx contracts library, such as making releases of the contracts via a separate repository so the user does not have to install our entire mono repo in their project (as will be the case when we first release this feature).

Theoretically, I could implement a solution where we read from the `versions.ts` file in the contracts package and return that in Solidity. But I would rather not do that as it is cumbersome and violates this convention. It's worth noting that I think this setup is only reasonable to do because the whole process for bumping both versions is completely automated as part of our release process.

packages/contracts/contracts/foundry/Sphinx.sol
```diff
@@ -266,6 +266,10 @@ abstract contract Sphinx {
         return sphinxUtils.getGnosisSafeProxyAddress(sphinxConfig);
     }
 
+    function sphinxLibraryVersion() public view returns (string memory) {
+        return sphinxUtils.getSphinxLibraryVersion();
+    }
```
sam-goldman, line 271: Thoughts on adding the library version to the `deploymentInfo` struct that we create in the `sphinxCollect` function in `Sphinx.sol`? This way we don't need to make another `forge script` call in our CLI code, which would be nice because forge script calls take a few seconds each

RPate97: That's basically what I'm planning to do, but it'll come in a separate PR when I implement the logic for checking if the contracts are out of date.

packages/contracts/scripts/write-constants.ts
RPate97: Yes, we can; not sure why I didn't just do that, tbh.

packages/contracts/contracts/foundry/Sphinx.sol
sam-goldman: Sounds good. When you do that, I think it'd be best to remove this function from `Sphinx.sol` to minimize the number of user-facing functions. We should be able to just call `sphinxUtils.getSphinxLibraryVersion()` in the `sphinxCollect` function

sam-goldman (approved):

PR #1359, opened 2024-01-22
feat(pg): Add sphinx install command

RPate97:
## Purpose
Adds the `sphinx install` command to the CLI.


sam-goldman (approved):
LGTM. A test would be nice, but it's not necessary

PR #1360, opened 2024-01-22
feat(pg): Install library during init command

RPate97:
## Purpose
- Updates the `init` command to also install the Sphinx library contracts via `forge install`
- Switches from using a git tag to using a commit hash for the installation process

## Context
I spent some time testing installations using git tags and found foundry was not able to reliably install packages using tags. It seems like the feature is bugged. So I decided to switch to using the commit hash instead which I found to be very reliable.


package.json
```diff
@@ -38,7 +38,7 @@
     "clean": "yarn workspaces run clean",
     "lint": "yarn workspaces run lint",
     "lint:fix": "yarn workspaces run lint:fix",
-    "changeset:version:custom": "yarn changeset version && yarn workspace @sphinx-labs/contracts run generate"
+    "changeset:version:custom": "yarn changeset version && yarn workspace @sphinx-labs/contracts run write-version"
```
RPate97, line 41: I had to make some changes to the release process to handle using the commit hash instead of the version number. The main difference is that we probably shouldn't just regenerate the versions in the standard build command b/c then we'd be frequently switching the hash out for different ones that may not be valid at all.

So instead, I've updated the build process to only run the `write-version` script when we're actually doing a release.

packages/contracts/test/mocha/remappings.spec.ts
```diff
@@ -34,4 +37,34 @@ describe('Remappings', () => {
...
+  /**
+   * We test the init command in the demo package. Unfortunately, it is impratical to test that we can
+   * install the sphinx contracts library and actually use the correct remappings in that test because
+   * the version installed would not be the same version as is being tested (it would be the lasted
+   * released version instead). So we disable installation of the library during the demo tests using
+   * an environment variable.
+   *
+   * Luckily, the only meaningful thing that is not covered by the demo tests is that the recommended
+   * remapping actually works. So we can simply add a test to check that directly.
+   */
```
RPate97, line 50: See comment.

packages/contracts/scripts/write-version.ts
```diff
@@ -1,12 +1,21 @@
...
+import { execSync } from 'child_process'
+
+const getCurrentCommitHash = () => {
+  try {
+    const commitHash = execSync('git rev-parse HEAD').toString().trim()
+    return commitHash
+  } catch (error) {
+    console.error('Error getting current commit hash:', error)
+    return null
+  }
```
sam-goldman, line 10: Optional:

We already have a `getCurrentGitCommitHash` function in the core package (in `utils.ts`). It'd be better to move it to the contracts package instead of essentially duplicating it here. If you do this, you'll also need to move its tests from the core package (in `utils.spec.ts`) to the contracts package. Lastly, since the `getCurrentGitCommitHash` may return `null`, we should throw an error in this file if it returns `null` instead of a string.

sam-goldman (approved):

packages/contracts/scripts/write-version.ts
sam-goldman: To clarify: we should check for the null case regardless of whether or not you merge these functions because they can both return null

PR #1379, opened 2024-01-25
fix(ct): Switch to using branch name in release process

RPate97:
## Purpose
Fixes an issue where it's not possible for the correct commit hash to be installed because of a circular dependency in the publishing process. I resolved this by switching to installing via a git branch which is created after the entire NPM release process finishes avoiding the circular dependency.


.github/workflows/release.yml
```diff
@@ -53,3 +53,20 @@ jobs:
...
+            const publishedPackages = JSON.parse(process.env.PUBLISHED_PACKAGES);
+            const contractsPackage = publishedPackages.find((pkg) => pkg.name === '@sphinx-labs/contracts');
+            if (contractsPackage) {
+              github.rest.git.createRef({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                ref: `refs/heads/v${contractsPackage.version}`,
+                sha: context.sha
+              });
+            }
```
RPate97, line 72: Creates a new branch for the release after the entire NPM release process is finished.

RPate97, line 72: I tested this by creating a temporary test workflow (which I've since removed from this branch). You can see the successful run [here](https://github.com/sphinx-labs/sphinx/actions/runs/7659291853).

packages/demo/test/Solc.spec.ts
```diff
@@ -119,9 +119,15 @@ describe('Solidity Compiler', () => {
...
-        )
+        if (stderr.includes('Unknown version provided')) {
+          console.log(
+            `Detected solc version not currently supported by foundry: ${version}`
+          )
+        } else {
+          errorMessages.push(
+            `Build failed for ${version} with optimizer enabled.\nSTDOUT: ${stdout}\nSTDERR: ${stderr}`
+          )
+        }
```
RPate97, line 130: I was running into an error: Unknown version provided in these tests which I do not think is related to my changes. I believe that solc 0.8.24 was recently released and does not appear to be supported by the foundry yet. So I just added logic to handle this specific situation.

Let me know if you disagree.

sam-goldman: Confirmed that Foundry doesn't support 0.8.24 yet

sam-goldman (approved):
Looks good

PR #1383, opened 2024-01-26
fix(core): Ensure required artifacts always included in output selection

RPate97:
## Purpose
Fixes an issue where its possible for required compiler artifacts to not be output during remote executor compilation due to the `outputSelection` not including them. 

We might want to consider if there are other implications wrt how we handle the build info files. Maybe we should consider adding the use of `--force` back into the deployment process? 

At a higher level, this sort of bug just adds more to the argument that we should:
- Ensure that the simulation process uses exactly the same logic as the production deployment process
- Consider removing the need to recompile in the DevOps platform at all by including all necessary info in the compiler config

### DIFF
--- a/.changeset/lucky-berries-work.md
+++ b/.changeset/lucky-berries-work.md
@@ -1,0 +1,5 @@
+---
+'@sphinx-labs/core': patch
+---
+
+Always include required artifacts in remote compiler output
--- a/packages/core/src/utils.ts
+++ b/packages/core/src/utils.ts
@@ -288,6 +288,27 @@ export const getConfigArtifactsRemote = async (
   // Get the compiler output for each compiler input.
   for (const compilerConfig of compilerConfigs) {
     for (const sphinxInput of compilerConfig.inputs) {
+      // Make sure any contracts we need are included in the output selection as long as they are in the input sources
+      for (const actionInput of compilerConfig.actionInputs) {
+        for (const { fullyQualifiedName } of actionInput.contracts) {
+          // Split the contract's fully qualified name into its source name and contract name.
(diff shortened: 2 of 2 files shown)


sam-goldman (commented):
This seems like a short-term solution that'd work, but it doesn't seem to address the root cause. See [this ticket](https://linear.app/chugsplash/issue/CHU-445/fix-code-in-the-plugin-that-creates-the-compilerconfig) for an explanation of what I think the root cause is.

Also, this PR seems to only be a temporary fix because it only resolves the issue in  `getConfigArtifactsRemote` and not the user's local compilation process. When we remove the remote compilation process in favor of always using the user's local compilation process, the fix in this PR will be removed, which will probably re-introduce the bug.

I'm happy to merge this as a short-term solution, but I just wanted to flag this in case you think it'd be better to implement the solution proposed in that ticket instead

RPate97:
> This seems like a short-term solution that'd work, but it doesn't seem to address the root cause. See [this ticket](https://linear.app/chugsplash/issue/CHU-445/fix-code-in-the-plugin-that-creates-the-compilerconfig) for an explanation of what I think the root cause is.
> 
> Also, this PR seems to only be a temporary fix because it only resolves the issue in `getConfigArtifactsRemote` and not the user's local compilation process. When we remove the remote compilation process in favor of always using the user's local compilation process, the fix in this PR will be removed, which will probably re-introduce the bug.
> 
> I'm happy to merge this as a short-term solution, but I just wanted to flag this in case you think it'd be better to implement the solution proposed in that ticket instead

Totally agree, lets discuss this on our sync.

PR #1384, opened 2024-01-26
fix(pg): Handle prior installation of sphinx library contracts

RPate97:
## Purpose
Updates the `sphinx install` command to handle cases where an old version of the Sphinx library was installed by removing it before then installing the required version.

## Testing
I replicated this issue with the following steps:
1. Run through the quickstart with the currently released packages
2. Create a new branch in the sphinx repo `v0.18.0-test` (or any name that does not include `@`)
3. Push the branch to github
4. Run `forge install sphinx-labs/sphinx@v0.18.0-test`

Note that this issue only occurs if the branch you want to install was not on GitHub when you first installed the sphinx library during the quickstart. I tested this fix quite extensively and confirmed the `install` command works in the following scenarios in addition to resolving the bug itself:
- the library is not installed at all
- the library is installed, but the `lib` or `lib/sphinx` folder has been deleted
- there are staged or unstaged changes in the users repo

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,6 +12,10 @@ jobs:
           keys:
             - v2-cache-source-{{ .Branch }}-{{ .Revision }}
             - v2-cache-source-{{ .Branch }}
+      - run:
+          name: Git reset
+          command: |
+            git reset --hard
       - checkout
       - run:
           name: Initialize submodules
--- a/packages/plugins/src/cli/install/index.ts
+++ b/packages/plugins/src/cli/install/index.ts
@@ -2,8 +2,9 @@ import { execSync } from 'child_process'
 
 import { CONTRACTS_LIBRARY_VERSION } from '@sphinx-labs/contracts'
 import { spawnAsync } from '@sphinx-labs/core'
+import ora from 'ora'
 
-export const handleInstall = async () => {
+const installWithUpdate = async () => {
   const args = [
     'install',
     `sphinx-labs/sphinx@${CONTRACTS_LIBRARY_VERSION}`,
@@ -12,14 +13,66 @@ export const handleInstall = async () => {
     '--no-commit',
   ]
 
-  const dependencyInstall = await spawnAsync('forge', args)
-  if (dependencyInstall.code !== 0) {
+  // Check if the library is installed
+  const submoduleStatus = await spawnAsync('git', ['submodule', 'status'])
+  if (submoduleStatus.code !== 0) {
     // The `stdout` contains the trace of the error.
-    console.log(dependencyInstall.stdout)
+    console.log(submoduleStatus.stdout)
     // The `stderr` contains the error message.
-    console.log(dependencyInstall.stderr)
+    console.log(submoduleStatus.stderr)
     process.exit(1)
   }
+
+  // If the library is already installed, then we need to update it by fetching the latest branches
+  // and installing the correct version.
+  if (submoduleStatus.stdout.includes('lib/sphinx')) {
+    // submodule status will indicate that our library is installed even if the `lib/sphinx` has been deleted
+    // which will cause the update to fail. So we have to run `forge install` to ensure the current version of the
+    // library is in the file system.
+    const installExistingDeps = await spawnAsync('forge', ['install'])
+    if (installExistingDeps.code !== 0) {
+      // The `stdout` contains the trace of the error.
+      console.log(installExistingDeps.stdout)
+      // The `stderr` contains the error message.
+      console.log(installExistingDeps.stderr)
+      process.exit(1)
+    }
+
+    // Run forge update to fetch the latest branches
+    const update = await spawnAsync('forge', ['update', 'sphinx-labs/sphinx'])
+    if (update.code !== 0) {
+      // The `stdout` contains the trace of the error.
+      console.log(update.stdout)
+      // The `stderr` contains the error message.
+      console.log(update.stderr)
+      process.exit(1)
+    }
+
+    // Install the correct version
+    const install = await spawnAsync('forge', args)
+    if (install.code !== 0) {
+      // The `stdout` contains the trace of the error.
+      console.log(install.stdout)
+      // The `stderr` contains the error message.
+      console.log(install.stderr)
+      process.exit(1)
+    }
+  } else {
+    const install = await spawnAsync('forge', args)
+    if (install.code !== 0) {
+      // The `stdout` contains the trace of the error.
+      console.log(install.stdout)
+      // The `stderr` contains the error message.
+      console.log(install.stderr)
+      process.exit(1)
+    }
+  }
+}
+
+export const handleInstall = async (spinner: ora.Ora) => {
+  spinner.start('Installing Sphinx Solidity library...')
+
+  await installWithUpdate()
 
   // Foundry doesn't consistently install our subdependency, so we make sure it's installed ourselves.
   // We should test this out later and remove if possible.
@@ -27,4 +80,5 @@ export const handleInstall = async () => {
     'cd lib/sphinx/packages/contracts && forge install --no-commit && cd ../../ && git restore packages/contracts/foundry.toml',
     { stdio: 'ignore' }
   )
+  spinner.succeed('Successfully installed Sphinx Solidity library')
 }
--- a/.changeset/weak-readers-admire.md
+++ b/.changeset/weak-readers-admire.md
@@ -1,0 +1,5 @@
+---
+'@sphinx-labs/plugins': patch
+---
+
+Add handling for previously installed incorrect version
--- a/packages/plugins/src/cli/setup.ts
+++ b/packages/plugins/src/cli/setup.ts
@@ -1,5 +1,6 @@
 import yargs from 'yargs'
 import { hideBin } from 'yargs/helpers'
+import ora from 'ora'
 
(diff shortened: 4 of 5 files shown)


.circleci/config.yml
```diff
@@ -12,6 +12,10 @@ jobs:
           keys:
             - v2-cache-source-{{ .Branch }}-{{ .Revision }}
             - v2-cache-source-{{ .Branch }}
+      - run:
+          name: Git reset
+          command: |
+            git reset --hard
```
RPate97, line 18: Solves an unrelated issue with the submodule installation due to caching in circle ci.

packages/plugins/src/cli/install/index.ts
```diff
@@ -14,12 +14,35 @@ export const handleInstall = async () => {
...
-    // The `stdout` contains the trace of the error.
-    console.log(dependencyInstall.stdout)
-    // The `stderr` contains the error message.
-    console.log(dependencyInstall.stderr)
-    process.exit(1)
+    // If installation fails due to missing tage, then run forge update and attempt again
+    if (
+      dependencyInstall.stderr.includes(
+        `Tag: "${CONTRACTS_LIBRARY_VERSION}" not found for repo "https://github.com/sphinx-labs/sphinx"`
+      )
```
RPate97, line 21: This error occurs because the local installation of the git repo does not include the target tag. Running `forge update sphinx-labs/sphinx` fetches the latest tags which then allows us to successfully run `forge install` with the target version.

packages/plugins/src/cli/install/index.ts
```diff
@@ -14,12 +14,35 @@ export const handleInstall = async () => {
...
   const dependencyInstall = await spawnAsync('forge', args)
   if (dependencyInstall.code !== 0) {
-    // The `stdout` contains the trace of the error.
-    console.log(dependencyInstall.stdout)
-    // The `stderr` contains the error message.
-    console.log(dependencyInstall.stderr)
-    process.exit(1)
+    // If installation fails due to missing tage, then run forge update and attempt again
+    if (
+      dependencyInstall.stderr.includes(
```
sam-goldman, line 19: I think relying on an error message thrown by Foundry to determine when we should run `forge update` is brittle. If they ever change the error message, this code will break. Is there another way to check whether the tag is missing locally? Or, maybe we could just always run `forge update`?

sam-goldman (commented):
I wasn't able to reproduce the error either, but I checked that this code works in a fresh repo and when upgrading from a previous `@sphinx-labs/plugins` version

packages/plugins/src/cli/install/index.ts
RPate97: I generally agree and spent time looking into this. I feel very strongly that we should just check the error message in this case for the following reasons:
- `forge update` errors if you do not have the dependency installed
- forge does not provide a reliable way to determine what dependencies are installed
- checking for the existence and/or contents of the `.gitmodules` file is not a reliable way to determine if the project installed (because there may be multiple .gitmodules files which are for different nested repos depending on the users own use of git submodules)
- checking for the existence of the library in the users project is not a reliable way to determine if it's installed

So the only way to determine if the user already has the library installed would be to search the git commit history. All around, I think pretty much anything we implement to try and determine if the library is already installed will be far more error-prone than just using the error message.

RPate97: I don't think there is another practical solution for this, mainly because Forge does not provide a reliable way to detect what dependencies are installed. If we wanted to detect this another way we'd have to do one of the following:
- Search for the `sphinx` folder in the users `lib` folder. This is not reliable because the user may remove the `lib/sphinx` folder, or it maybe removed automatically by forge (due to a failed dependency installation for example)
- Search for the `.gitmodules` file and check if it contains the sphinx library. This is not reliable because there may be multiple `.gitmodules` files in the project due to the use of submodels outside of just how foundry uses them.
- Search through the git commit history for evidence of the library being installed. I did not try to implement this, but it seems like it would be very error prone as well.

So I don't think there's an alternative to checking the error message that is not even more fragile. I think the only real alternative would be to simply remove this command and instead ask the user to install the library themselves, but I don't think we should do that either because humans are also not that reliable.

sam-goldman: Why not always run `forge install` then `forge update sphinx-labs/sphinx`?

packages/plugins/src/cli/install/index.ts
```diff
@@ -14,17 +15,63 @@ export const handleInstall = async () => {
...
-    console.log(dependencyInstall.stderr)
-    process.exit(1)
+    // If installation fails due to missing tag, then run forge update and attempt again
+    if (
+      dependencyInstall.stderr.includes(
+        `Tag: "${CONTRACTS_LIBRARY_VERSION}" not found for repo "https://github.com/sphinx-labs/sphinx"`
+      )
+    ) {
+      // Run forge install since installing with the incorrect version deletes the previous installation
+      const install = await spawnAsync('forge', ['install'])
```
sam-goldman, line 25: Running `forge install` without any arguments installs all existing dependencies, not just Sphinx. Is that the intended effect? If not, would it make more sense to run this command with our package as an argument?

packages/plugins/src/cli/install/index.ts
sam-goldman: To be clear, I'm asking why we can't always run these commands instead of just running them if an error occurs during the initial `forge install` command

packages/plugins/src/cli/install/index.ts
```diff
@@ -14,17 +15,63 @@ export const handleInstall = async () => {
...
-    // The `stderr` contains the error message.
-    console.log(dependencyInstall.stderr)
-    process.exit(1)
+    // If installation fails due to missing tag, then run forge update and attempt again
+    if (
+      dependencyInstall.stderr.includes(
+        `Tag: "${CONTRACTS_LIBRARY_VERSION}" not found for repo "https://github.com/sphinx-labs/sphinx"`
+      )
+    ) {
+      // Run forge install since installing with the incorrect version deletes the previous installation
```
sam-goldman, line 24: I don't understand this. What do you mean by "installing with the incorrect version"? Also, could you elaborate on why it's necessary to run `forge install` in the docs?

packages/plugins/src/cli/install/index.ts
RPate97: `forge update sphinx-labs/sphinx` does not work if you do not have the library installed. We could use it, but then we would need to catch the error and match on the message for that instead.

packages/plugins/src/cli/install/index.ts
```diff
@@ -14,17 +15,63 @@ export const handleInstall = async () => {
...
+
+  await installWithUpdate()
 
   // Foundry doesn't consistently install our subdependency, so we make sure it's installed ourselves.
   // We should test this out later and remove if possible.
   execSync(
     'cd lib/sphinx/packages/contracts && forge install --no-commit && cd ../../ && git restore packages/contracts/foundry.toml',
     { stdio: 'ignore' }
   )
+  spinner.start('Successfully installed Sphinx Solidity library')
```
sam-goldman, line 76: This should be `spinner.succeed` I think

sam-goldman, line 76: When I attempt to run `npx sphinx install` in the demo package, the process never exits. I don't know whether that's caused by the fact that we don't call `spinner.success` or if it's caused by the `execSync` call above

packages/plugins/src/cli/install/index.ts
RPate97: That is the intended effect.

packages/plugins/src/cli/install/index.ts
sam-goldman: Why would we need to catch the error message? I.e. can't we do this?
1. Run `forge install`
2. Run `forge update sphinx-labs/sphinx`, wrapping it in a try/catch. Don't do anything if it fails.
3. Run `forge install sphinx-labs/sphinx@${CONTRACTS_LIBRARY_VERSION} --no-commit`

I understand that you've tested this extensively, and I'm sure this approach works now. My concern is that i'll break in the future.

RPate97: `forge update `sphinx-labs/sphinx` could fail for many reasons. It is unsafe to catch and do nothing.

packages/plugins/src/cli/install/index.ts
RPate97: Good catch, this was related to the fact that we weren't calling `spinner.success`

packages/plugins/src/cli/install/index.ts
sam-goldman: What about using `git submodule status` to check if the sphinx dependency is already installed?

RPate97: I'm ok with using that if you're more comfortable with it. I've updated this PR to work that way.

RPate97: My general opinion on `git submodule status` is that it's probably also subject to bugs related to the user modifying their `.gitmodules` file manually and/or having multiple `.gitmodules` files in the same repo. However, I don't think it's any less likely to produce bugs than forge updating their error message, so I'm cool with it.

sam-goldman (approved):
Looks good!

PR #1387, opened 2024-01-28
feat(pg): Support new networks

RPate97:
## Purpose
Adds logic to automatically update drippie configuration when necessary. 
Adds support for:
- Celo + testnet
- Kava + testnet
- Scroll + testnet
- EVMOS + testnet
- Moonbeam, Moonriver, Moonbase Alpha (testnet)
- OKX Chain

Note that I am intentionally not adding support for:
- OKX Chain testnet: I was not able to find a reliable RPC endpoint for this network with support for all the rpc methods we need. We'd have to run our own node to support this network.
- Heco or it's testnet: I was not able to find a way to bridge to this network or an exchange which I can legally access and that would sell me their token in a reasonable amount of time. I was also not able to find any node providers for this network. Honestly, this network seems defunct. Their git repo hasn't seen a single commit in 2 years and their TVL is lower than the [amount their bridge was hacked for in November](https://hacken.io/insights/heco-bridge-hack-explained/). I think the hack may have killed them.
- Rootstock or its testnet: This is due to ethers [failing to reliably estimate the gas limit for action execution transactions](https://linear.app/chugsplash/issue/CHU-465/handle-gas-limit-estimation-bug-on-gnosis-rootstock-and-fuse). We could totally support them, but it would require more effort to investigate the issue and implement a solution that works reliably. 
- Fuse sparknet (testnet): Ethers does not accurately estimate gas limit which causes transactions to revert. This may be related to the issue with Rootstock.

## Todo before release
- [x] Find node providers for all networks
- [x] Get reasonable quantity of tokens on all networks
- [x] Determine which networks are supported etherscan or blockscout and get API keys 
- [x] Deploy and verify contracts on all networks
- [x] Test the deploy command on all networks
- [x] Test deploying via the website on all networks

## Notes
I've created a [ticket to address some issues I noticed with the deploy command](https://linear.app/chugsplash/issue/CHU-447/implement-better-timeout-and-retry-logic-in-deploy-task) separately.


packages/contracts/src/networks.ts
```diff
@@ -346,199 +390,262 @@ export const SPHINX_NETWORKS: Array<SupportedNetwork> = [
...
+      apiURL: 'https://api-moonbase.moonscan.io/api',
+      browserURL: 'https://moonbase.moonscan.io/',
+      envKey: 'MOON_ETHERSCAN_API_KEY',
+      supported: true,
+    },
+    currency: 'DEV',
+    dripSize: '0.05',
+    requiredEnvVariables: ['MOONBASE_ALPHA_RPC_URL'],
+    networkType: 'Testnet',
+    dripVersion: 1, // todo - redeploy this
```
RPate97, line 483: Moonbase alpha is down right now. I will come back to this when it comes back online.

packages/contracts/src/networks.ts
```diff
@@ -6,11 +6,13 @@ export type SupportedNetwork = {
     apiURL: string
     browserURL: string
     envKey: string
+    supported: boolean
   }
   currency: string
   dripSize: string
   requiredEnvVariables: Array<string>
   networkType: NetworkType
+  dripVersion: number
```
RPate97, line 15: Added a `dripVersion` field which makes it easy for us to update the drip sizes on different networks. We just have to update the `dripSize`, bump the `dripVersion`, and then we can run the deploy script and it will archive any previous Drippie configurations and replace them with the new version

packages/core/hardhat.config.ts
```diff
@@ -167,7 +77,7 @@ task('deploy-system')
         )
       }
       const provider = new SphinxJsonRpcProvider(hre.network.config.url)
-      const signer = await hre.ethers.provider.getSigner()
+      const signer = new Wallet(process.env.PRIVATE_KEY!).connect(provider)
```
RPate97, line 80: This is necessary because sending transactions on OKX Chain only works with the SphinxJsonRPCProvider. So we can't just use the signer provided by the HRE here.

packages/core/src/etherscan.ts
```diff
@@ -35,7 +35,6 @@ export const getChainConfig = (chainId: number): ChainConfig => {
   const chainConfig = [
     // custom chains has higher precedence than builtin chains
     ...[...customChains].reverse(), // the last entry has higher precedence
-    ...builtinChains,
```
RPate97, line 38: I dropped the use of `buildinChains` because we always rely on the block explorer info we define ourselves in the `SPHINX_NETWORKS` array.

packages/core/src/etherscan.ts
```diff
@@ -311,6 +310,12 @@ export const etherscanVerifySphinxSystem = async (
       let buildInfo: BuildInfo
       if (type === SystemContractType.SPHINX) {
         buildInfo = sphinxBuildInfo
+        // We have to remove `.Darwin.appleclang` to make etherscan verify our contracts
+        // This could also be an issue when verifying end user contracts, but we haven't seen it so far
+        buildInfo.solcLongVersion = buildInfo.solcLongVersion.replace(
+          '.Darwin.appleclang',
+          ''
+        )
```
RPate97, line 318: See the comment; we should keep an eye out for this being a problem for users deploying via the DevOps platform. So far, I haven't encountered this issue and I've been deploying using the same machine I used to compile our core contracts.

packages/core/src/etherscan.ts
```diff
@@ -311,6 +310,14 @@ export const etherscanVerifySphinxSystem = async (
...
       if (type === SystemContractType.SPHINX) {
         buildInfo = sphinxBuildInfo
+        // We have to remove `.Darwin.appleclang` to make etherscan verify our contracts which is included in the
+        // compiler version for our contracts. It's unclear why this hasn't been an issue historically.
+        // This could also be an issue when verifying end user contracts, but we haven't seen it yet presumably
+        // because we always recompile the contracts. This could be an argument for continuing to do that.
+        buildInfo.solcLongVersion = buildInfo.solcLongVersion.replace(
+          '.Darwin.appleclang',
+          ''
+        )
```
RPate97, line 320: See the comment; we should keep an eye out for this being a problem for users deploying via the DevOps platform especially when we move to only compiling one time.

It's currently impossible for this issue to happen on the DevOps platform since we recompile the contracts ourselves.

packages/core/src/languages/solidity/predeploys.ts
```diff
@@ -175,6 +220,21 @@ export const deploySphinxSystem = async (
...
+
+    // Cancel any out of date drips
+    await cancelPreviousDripVersions(
+      Drippie,
+      provider,
+      owner,
+      executionMode,
+      baseDripName,
+      currentDripVersion
+    )
```
RPate97, line 236: Handles cancelling the previous drip versions if necessary, the latest version is added below.

packages/core/src/provider.ts
```diff
@@ -12,4 +12,16 @@ export class SphinxJsonRpcProvider extends JsonRpcProvider {
...
+  // For shame!
+  // See this issue for more info: https://github.com/NomicFoundation/hardhat/issues/4010
+  // So on OKT Chain we have to call eth_estimateGas ourselves to handle this....
+  estimateGas(_tx: TransactionRequest): Promise<bigint> {
+    if (_tx.chainId === 66) {
+      return this.send('eth_estimateGas', [_tx])
+    }
+
+    return super.estimateGas(_tx)
+  }
```
RPate97, line 26: See comment; it's necessary to override `estimateGas` on OKT Chain due to the fact that they use an outdated geth client. This workaround resolves the issue, but it means we always have to use `SphinxJsonRPCProvider` to interact with that network. I think that is acceptable as we already are very good about using the custom provider. After I implemented this workaround, everything worked properly on the network.

README.md
```diff
@@ -21,6 +21,10 @@ Deployments with Sphinx are a three-step process:
 
 * **Compatible with Forge Scripts**: You can integrate Sphinx by adding a few lines of code to your existing Forge scripts.
 
+* **Deployment Artifact Management**: Generates and stores deployment artifacts which can be retrieved at any time.
+
+* **Contract Verification**: Automatically verifies contracts on block explorers with no configuration necessary.
+
```
RPate97, line 27: Just thought it would be nice to include these.

packages/contracts/src/networks.ts
```diff
@@ -6,11 +6,16 @@ export type SupportedNetwork = {
...
+    blockExplorer: BlockExplorer
   }
   currency: string
   dripSize: string
   requiredEnvVariables: Array<string>
   networkType: NetworkType
+  dripVersion: number
+  decimals: number
+  queryFilterBlockLimit: number
+  legacyTx: boolean
```
RPate97, line 18: Added a few new fields to the networks info.

The `dripVersion` field makes it easy for us to update the drip sizes on different networks. We just have to update the dripSize, bump the dripVersion, and then we can run the deploy script and it will archive any previous Drippie configurations and replace them with the new version. 

The `decimals` field is used whenever we need to handle currency conversions (mainly on the website).

The `queryFilterBlockLimit` is also just used by the website; it's for cases where networks limit the block range you can use when querying events.

The `legacyTx` field allows us to support networks that do not use eip-1559 transactions where we must use a `--legacy` flag when calling into forge to collect the transactions.

packages/core/src/etherscan.ts
```diff
@@ -259,14 +259,6 @@ export const attemptVerification = async (
   }
 }
 
-export const isSupportedNetworkOnEtherscan = (chainId: number): boolean => {
-  const chainConfig = [...customChains, ...builtinChains].find(
-    (config) => config.chainId === chainId
-  )
-
-  return chainConfig !== undefined
-}
```
RPate97, line 268: We had two copies of this function, so I deleted this in favor of just using `isVerificationSupportedForNetwork`.

packages/core/src/languages/solidity/predeploys.ts
```diff
@@ -44,6 +48,48 @@ export const ensureSphinxAndGnosisSafeDeployed = async (
   await deploySphinxSystem(provider, wallet, relayers, executionMode, logger)
 }
 
+export const cancelPreviousDripVersions = async (
```
RPate97, line 51: Handles archiving any previously created Drippie drips before the new version is created.

packages/contracts/src/networks.ts
```diff
@@ -346,199 +459,306 @@ export const SPHINX_NETWORKS: Array<SupportedNetwork> = [
...
+      browserURL: 'https://moonbase.moonscan.io/',
+      envKey: 'MOONBEAM_ETHERSCAN_API_KEY',
+      blockExplorer: 'Etherscan',
+    },
+    currency: 'GLMR',
+    dripSize: '0.05',
+    requiredEnvVariables: ['MOONBASE_ALPHA_RPC_URL'],
+    queryFilterBlockLimit: 500,
+    networkType: 'Testnet',
+    dripVersion: 1, // todo - redeploy this
```
RPate97, line 566: Will do this and the one below before merging.

sam-goldman (commented):
Looks great, just some minor comments

sam-goldman:
Just realized that this PR doesn't update the `sphinxConfig.testnets/mainnets` field in the [configuration options doc](https://github.com/sphinx-labs/sphinx/blob/main/docs/configuration-options.md#devops-platform-options)

packages/contracts/src/networks.ts
RPate97: before is fine

README.md
```diff
@@ -71,13 +75,13 @@ A demo of the DevOps Platform is on the [website's landing page](https://sphinx.
 
 - Sphinx supports `CREATE2` and `CREATE3` but not the `CREATE` opcode, i.e. `new MyContract(...)`.
 - You cannot send ETH as part of a deployment.
-- You can only use the [Deploy CLI Command](https://github.com/sphinx-labs/sphinx/blob/main/docs/cli-deploy.md) on live networks if you're the only owner of the Gnosis Safe. (Deployments with the DevOps Platform support an arbitrary number of owners).
+- You can only use the [Deploy CLI Command](https://github.com/sphinx-labs/sphinx/blob/main/docs/cli-deploy.md) on live networks if your Gnosis Safe has a single owner. (Deployments with the DevOps Platform support an arbitrary number of owners).
+- Contract verification is only supported on networks supported by Blockscout or Etherscan.
```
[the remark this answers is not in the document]
RPate97: I can make a ticket

packages/contracts/src/networks.ts
```diff
@@ -302,25 +400,35 @@ export const SPHINX_NETWORKS: Array<SupportedNetwork> = [
...
     chainId: BigInt(4002),
     rpcUrl: () => process.env.FANTOM_TESTNET_RPC_URL!,
     etherscan: {
-      apiURL: 'https://api-testnet.ftmscan.com/api',
-      browserURL: 'https://testnet.ftmscan.com',
-      envKey: 'FANTOM_ETHERSCAN_API_KEY',
+      apiURL: '',
+      browserURL: '',
+      envKey: '',
+      blockExplorer: 'Unsupported',
```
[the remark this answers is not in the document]
RPate97: No we do not, our verification logic does not work on the fantom testnet for some reason. I haven't prioritized investigating it.

packages/core/src/languages/solidity/predeploys.ts
```diff
@@ -44,6 +48,48 @@ export const ensureSphinxAndGnosisSafeDeployed = async (
...
+          ).wait()
+        }
+
+        await (
+          await Drippie.status(
+            previousDripName,
+            3,
+            await getGasPriceOverrides(provider, wallet, executionMode)
+          )
+        ).wait()
```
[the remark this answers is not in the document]
RPate97: Yeah, it's intentional Drippie just requires pausing the drip before archiving it which is what we're doing here.

packages/core/src/languages/solidity/predeploys.ts
```diff
@@ -175,21 +221,39 @@ export const deploySphinxSystem = async (
   for (const relayer of relayers) {
     const chainId = (await provider.getNetwork()).chainId
 
+    const currentDripVersion = fetchDripVersionForNetwork(chainId)
+    const baseDripName = `sphinx_fund_${relayer}`
+    const dripName =
+      baseDripName + (currentDripVersion > 0 ? `_${currentDripVersion}` : '')
```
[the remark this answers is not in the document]
RPate97: Yeah, this is intentional; it's because the first version I deployed did not include a version number. This logic just replicates that so we get a consistent setup on all networks.

packages/core/src/languages/solidity/predeploys.ts
```diff
@@ -175,21 +221,39 @@ export const deploySphinxSystem = async (
...
+      logger
+    )
+
     const reentrant = false
     const interval = 30
     const dripcheck = getCheckBalanceLowAddress()
     const checkparams = AbiCoder.defaultAbiCoder().encode(
       ['address', 'uint256'],
-      [relayer, ethers.parseEther(fetchDripSizeForNetwork(chainId))]
+      [relayer, ethers.parseEther(fetchDripSizeForNetwork(chainId)) * BigInt(5)]
```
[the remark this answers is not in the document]
RPate97: Here we're setting the maximum balance allowed in the recipient wallet. So if they have a balance greater than this, and you attempt to trigger another drip, it will fail due to them having a large balance. 

Initially, I just used a limit equal to 1 * dripSize, and this is how most of the networks are set up. However, it's possible we might have to drip multiple times to get enough funds to execute a very large deployment. This hasn't been an issue so far, so I haven't actually redeployed on all of the networks to adjust this. But I will if we run into problems related to this.

RPate97:
> Just realized that this PR doesn't update the `sphinxConfig.testnets/mainnets` field in the [configuration options doc](https://github.com/sphinx-labs/sphinx/blob/main/docs/configuration-options.md#devops-platform-options)

good catch

sam-goldman (approved):

PR #1388, opened 2024-01-29
fix(core): merge compiler inputs based on build info ID instead of solc version

sam-goldman:
Previously, we were combining contract sources from different build info files as long as the build info files had the same `solcLongVersion`. This was problematic because different build info files can contain different compiler settings. This PR fixes this issue by only combining sources from the _same_ build info file. We identify each build info file according to its `id` field.

I manually tested that this fixes the issue in the "Steps to reproduce" section of the [ticket for this issue](https://linear.app/chugsplash/issue/CHU-445/fix-code-in-the-plugin-that-creates-the-sphinxinputs-array).

### DIFF
--- a/packages/core/src/utils.ts
+++ b/packages/core/src/utils.ts
@@ -334,7 +334,9 @@ export const getConfigArtifactsRemote = async (
         const [sourceName, contractName] = fullyQualifiedName.split(':')
 
         const buildInfo = solcArray.find(
-          (e) => e.output.contracts[sourceName][contractName]
+          // We use the optional chaining operator so that this line doesn't throw an error if
+          // `contracts[sourceName]` is undefined.
+          (e) => e.output.contracts[sourceName]?.[contractName]
         )
         if (!buildInfo) {
           throw new Error(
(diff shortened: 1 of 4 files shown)


sam-goldman:
@RPate97 It looks like the monorepo is failing to build for some reason unrelated to this PR. I created a [test repo](https://github.com/sphinx-labs/sphinx/tree/temp) that's a clone of `develop`, which is also failing to build. Want me to make an issue for this?

packages/core/src/utils.ts
```diff
@@ -334,7 +334,9 @@ export const getConfigArtifactsRemote = async (
         const [sourceName, contractName] = fullyQualifiedName.split(':')
 
         const buildInfo = solcArray.find(
-          (e) => e.output.contracts[sourceName][contractName]
+          // We use the optional chaining operator so that this line doesn't throw an error if
+          // `contracts[sourceName]` is undefined.
+          (e) => e.output.contracts[sourceName]?.[contractName]
```
sam-goldman, line 339: This fixes an unrelated bug that was causing Mean Finance's project to fail.

RPate97 (approved):
Short and sweet, looks good to me.

PR #1393, opened 2024-01-31
feat(pg): Support Rootstock

RPate97:
## Purpose
Adds support for Rootstock. I did not add support for the Rootstock testnet because I was not able to find a reliable node provider that supported all the RPC methods we require.

### DIFF
--- a/packages/contracts/src/networks.ts
+++ b/packages/contracts/src/networks.ts
@@ -16,6 +16,7 @@ export type SupportedNetwork = {
   decimals: number
   queryFilterBlockLimit: number
   legacyTx: boolean
+  actionGasLimitBuffer: boolean
 }
 
 export type SupportedLocalNetwork = {
@@ -23,6 +24,7 @@ export type SupportedLocalNetwork = {
   chainId: bigint
   networkType: NetworkType
   legacyTx: false
+  actionGasLimitBuffer: false
 }
 
 export const SPHINX_LOCAL_NETWORKS: Array<SupportedLocalNetwork> = [
@@ -31,6 +33,7 @@ export const SPHINX_LOCAL_NETWORKS: Array<SupportedLocalNetwork> = [
     chainId: BigInt(31337),
     networkType: 'Local',
     legacyTx: false,
+    actionGasLimitBuffer: false,
   },
 ]
 
@@ -57,6 +60,7 @@ export const SPHINX_NETWORKS: Array<SupportedNetwork> = [
     decimals: 18,
     queryFilterBlockLimit: 2000,
     legacyTx: false,
+    actionGasLimitBuffer: false,
   },
   {
     name: 'sepolia',
@@ -77,6 +81,7 @@ export const SPHINX_NETWORKS: Array<SupportedNetwork> = [
     decimals: 18,
     queryFilterBlockLimit: 2000,
     legacyTx: false,
+    actionGasLimitBuffer: false,
   },
   {
     name: 'optimism',
@@ -97,6 +102,7 @@ export const SPHINX_NETWORKS: Array<SupportedNetwork> = [
     decimals: 18,
     queryFilterBlockLimit: 2000,
     legacyTx: false,
+    actionGasLimitBuffer: false,
   },
   {
     name: 'optimism_sepolia',
@@ -117,6 +123,7 @@ export const SPHINX_NETWORKS: Array<SupportedNetwork> = [
     decimals: 18,
     queryFilterBlockLimit: 2000,
     legacyTx: false,
+    actionGasLimitBuffer: false,
   },
   {
     name: 'arbitrum',
@@ -137,6 +144,7 @@ export const SPHINX_NETWORKS: Array<SupportedNetwork> = [
     decimals: 18,
     queryFilterBlockLimit: 2000,
     legacyTx: false,
+    actionGasLimitBuffer: false,
   },
   {
     name: 'arbitrum_sepolia',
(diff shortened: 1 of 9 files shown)


packages/core/src/utils.ts
```diff
@@ -228,7 +228,7 @@ export const getGasPriceOverrides = async (
     // On Gnosis, set the gas limit artificially high (since ethers does not seem to always estimate it proplerly especially for contract deployments)
     case 100:
     case 10200:
-      overridden.gasLimit = getMaxGasLimit(block.gasLimit)
+      overridden.gasLimit = (block.gasLimit / BigInt(4)) * BigInt(3)
```
RPate97, line 231: Just fixed this since it was convenient, don't want us to forget.

packages/contracts/src/networks.ts
```diff
@@ -722,43 +759,26 @@ export const SPHINX_NETWORKS: Array<SupportedNetwork> = [
...
+    rpcUrl: () => process.env.ROOTSTOCK_RPC_URL!,
+    etherscan: {
+      apiURL: 'https://rootstock.blockscout.com/api',
+      browserURL: 'https://rootstock.blockscout.com/',
+      envKey: 'ROOTSTOCK_ETHERSCAN_API_KEY',
+      blockExplorer: 'Blockscout',
+    },
+    currency: 'RBTC',
+    dripSize: '0.001',
+    requiredEnvVariables: ['ROOTSTOCK_RPC_URL'],
```
sam-goldman, line 776: General question: does the `requiredEnvVariables` field not need to include the etherscan api key?

packages/contracts/src/networks.ts
```diff
@@ -722,43 +759,26 @@ export const SPHINX_NETWORKS: Array<SupportedNetwork> = [
...
+    dripVersion: 2,
+    networkType: 'Mainnet',
+    legacyTx: true,
+    decimals: 8,
+    queryFilterBlockLimit: 2000,
```
sam-goldman, line 781: Another general question: does the `queryFilterBlockLimit` apply to all networks, even if we're using a paid RPC url? I'm surprised that node providers on Ethereum restrict this to 2000 blocks

packages/core/src/actions/execute.ts
```diff
@@ -417,25 +415,95 @@ export const executeActionsViaManagedService: ExecuteActions = async (
...
+  ])
+  const managedServiceExecData = managedService.interface.encodeFunctionData(
+    'exec',
+    [moduleAddress, executionData]
+  )
+
+  const overrides: ethers.TransactionRequest = {}
+  if (
+    shouldBufferExecuteActionsGasLimit((await provider.getNetwork()).chainId)
+  ) {
```
sam-goldman, line 435: This is optional, but both `executeActionsViaSigner` and `executeActionsViaManagedService` call `await provider.getNetwork` in every call, which is inefficient. I think we already have the `chainId` available as a variable in the functions that call into these `executeActions` functions, so it'd be easy to pass it in as a parameter, which'd save 1 RPC call per batch of execution actions

packages/core/src/actions/execute.ts
```diff
@@ -417,25 +415,95 @@ export const executeActionsViaManagedService: ExecuteActions = async (
...
+    const minimumActionGas = estimateGasViaManagedService(moduleAddress, batch)
+    const gasEstimate = await signer.estimateGas({
+      to: getManagedServiceAddress(),
+      data: managedServiceExecData,
+    })
+
+    const block = await provider.getBlock('latest')
```
sam-goldman, line 442: > This is optional, but both executeActionsViaSigner and executeActionsViaManagedService call await provider.getNetwork in every call, which is inefficient. I think we already have the chainId available as a variable in the functions that call into these executeActions functions, so it'd be easy to pass it in as a parameter, which'd save 1 RPC call per batch of execution actions

Likewise for `provider.getBlock`. (We already have the `blockGasLimit` on the functions that call these execution functions)

packages/core/src/utils.ts
sam-goldman: Thanks. Just so we have some written record of why we set this to 3/4 of the block gas limit instead of 1/2: 

Say our batching logic (which uses heuristics) results in a batch that's 50% of the block gas limit. If the heuristic underestimates the gas used (i.e. the actual gas used is greater than 50%), we'd underfund the transaction by setting the `gasLimit` to be 50% of the block gas limit here.

sam-goldman (commented):
Looks good, just a few minor comments. Also, it looks like you need to include Rootsock in the "Supported Networks" section of the main README and in the [configuration options doc](https://github.com/sphinx-labs/sphinx/blob/main/docs/configuration-options.md#devops-platform-options)

packages/contracts/src/networks.ts
RPate97: There are a few different networks where this is an issue, but it isn't an issue on all of them. However, previously I just had a hardcoded limit to 2k on all networks. So I just set this to 2k for all networks to maintain the current behavior.

packages/contracts/src/networks.ts
RPate97: The pattern is intended to support that use case, but I haven't implemented any of the logic for it. Even if we included the Etherscan API key in this array though, we'd still need to have the field defined separately in `etherscan.envKey` because we need to know specifically what env variable to read from to get the key.

sam-goldman (approved):
Looks good!

PR #1401, opened 2024-02-01
refactor(pg): create decoded actions when making parsed config

sam-goldman:
This PR is a necessary first step towards migrating from Foundry's broadcast file to the state diff cheatcode.

Before this PR, we used Foundry's broadcast file to create the `decodedActions`, which we use in the preview. Specifically, we used the broadcast file's `functionName`, `contractName`, and `callArguments` to construct the `decodedActions`. Since the state diff does not contain any of this decoded information, we must decode the user's transactions ourselves.

This PR implements the code that decodes the transactions by using the `configArtifacts` and a bunch of EthersJS methods. This PR doesn't change the decoded actions themselves.

Everything in this PR is pretty extensively tested


packages/contracts/src/utils.ts
```diff
@@ -154,6 +154,10 @@ export const add0x = (str: string): string => {
  * @param types The types of the variables in the `Result`. This can be retrieved from an
  * `ethers.Interface` object.
  * @param values The `Result` to convert.
+ *
+ * @returns The converted result. The returned object/array will not contain strings instead of
+ * BigInt values to ensure that `JSON.stringify` and `JSON.parse` can be called on it without
+ * causing an error.
```
sam-goldman, line 160: We now use this function (`recursivelyConvertResult`) to decode constructor and function arguments so that we can display them in the preview. This PR makes a slight revision to this function so that it never returns BigInts, which makes it safe to `JSON.stringify` the arguments in the preview

packages/plugins/src/cli/propose/index.ts
```diff
@@ -189,16 +190,18 @@ export const buildParsedConfigArray: BuildParsedConfigArray = async (
     // transactions broadcasted in the user's script for this network. We return an empty array in
     // this case.
     const actionInputs = collectionDryRun
-      ? convertFoundryDryRunToActionInputs(
-          deploymentInfo,
-          collectionDryRun,
-          collectionDryRunPath
-        )
+      ? convertFoundryDryRunToActionInputs(deploymentInfo, collectionDryRun)
```
sam-goldman, line 193: We now use the dry run path as an input to `makeParsedConfig`. (Likewise for the `deploy` function)

packages/plugins/src/foundry/decode.ts
```diff
@@ -280,11 +257,18 @@ export const makeParsedConfig = (
     unlabeledContracts.push(...unlabeledAdditionalContracts)
 
     if (isRawCreate2ActionInput(input)) {
-      const fullyQualifiedName = findFullyQualifiedName(
+      const fullyQualifiedName = findFullyQualifiedNameForInitCode(
```
sam-goldman, line 260: Renamed this function to avoid an ambiguous name due to the presence of a new function, `findFullyQualifiedNameForInitCode`

packages/plugins/src/foundry/decode.ts
```diff
@@ -382,3 +382,96 @@ const parseAdditionalContracts = (
...
+      variables,
+      address: create2Address,
+    }
+  } else {
+    return {
+      referenceName: create2Address,
+      functionName: 'deploy',
+      variables: [],
+      address: create2Address,
+    }
```
sam-goldman, line 429: These are functionally equivalent to the previous `decodedActions`

packages/plugins/src/foundry/decode.ts
```diff
@@ -382,3 +382,96 @@ const parseAdditionalContracts = (
...
+        ? `Very large calldata. View it in Foundry's dry run file: ${dryRunPath}`
+        : data,
+    ]
+    return {
+      referenceName: to,
+      functionName: 'call',
+      variables,
+      address: '',
+    }
+  }
```
sam-goldman, line 476: These are also functionally equivalent to the decoded actions before this PR

packages/plugins/src/foundry/decode.ts
```diff
@@ -382,3 +382,96 @@ const parseAdditionalContracts = (
...
+      data.length > 1000
+        ? `Very large calldata. View it in Foundry's dry run file: ${dryRunPath}`
+        : data,
+    ]
+    return {
+      referenceName: to,
+      functionName: 'call',
+      variables,
+      address: '',
+    }
```
sam-goldman, line 475: It's rather confusing that function call actions with no fully qualified name have a `referenceName` field equal to the `to` address, and an `address` field equal to an empty string. However, this is consistent with the previous implementation for the decoded actions, and changing this is out of scope for this PR

packages/plugins/src/foundry/utils/index.ts
```diff
@@ -1168,3 +1168,26 @@ export const findFullyQualifiedName = (
   // If we make it to this point, we couldn't find a fully qualified name for this init code.
   return undefined
 }
+
+/**
+ * Returns the fully qualified name that corresponds to a given address. Returns `undefined` if no
+ * fully qualified name exists for the address. This function will only find a fully qualified name
+ * if it corresponds to a contract deployed in one of the actions.
+ */
+export const findFullyQualifiedNameForAddress = (
```
sam-goldman, line 1177: This is the un-implemented function that's causing CI to fail. I figured there's no point implementing this function because the `RawActionInput` will have a different structure when I make the PR that transitions to the state diff. In other words, I'd need to completely redo this function in the next PR. (The biggest difference is that the `RawActionInput` probably wont' have an `additionalContracts` array in the state diff).

RPate97:
> @RPate97 You can review this PR even though CI is failing. It's failing because there's a helper function that I don't want to bother implementing until I've written the state diff logic in the next PR. See [this comment ](https://github.com/sphinx-labs/sphinx/pull/1401/files#r1474964071) for more details

Please use a feature branch. Thanks.

sam-goldman:
> Please use a feature branch. Thanks. Happy to review as is though, just don’t want to merge to develop.

Fixed

packages/plugins/src/foundry/decode.ts
RPate97: Yeah, I agree this is weird. It was just implemented this way because it didn't require refactoring the rendering function. I've created a ticket for it.

RPate97 (approved):
Looks good, just left a couple minor comments.

packages/plugins/src/foundry/decode.ts
sam-goldman: There's an [existing ticket](https://linear.app/chugsplash/issue/CHU-324/improve-preview) for various preview improvements. I think it's already in there

packages/contracts/test/SphinxInitCode.sol
```diff
@@ -4,22 +4,64 @@ pragma solidity >=0.6.2 <0.9.0;
 import { SystemContractInfo } from "../contracts/foundry/SphinxPluginTypes.sol";
 
 contract SphinxInitCode {
-  function getSystemContractInfo() public pure returns (SystemContractInfo[] memory) {
-    SystemContractInfo[] memory contracts = new SystemContractInfo[](14);
```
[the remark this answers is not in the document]
sam-goldman: It's because I linted it in this PR

PR #1403, opened 2024-02-04
feat(pg): migrate to the state diff cheatcode

sam-goldman:
Unfortunately, there wasn't a straightforward way to split this PR into smaller PRs. I tried my best to add comments and docs to make the review process easier. I don't recommend trying to review it all in one sitting


packages/contracts/contracts/foundry/Sphinx.sol
```diff
@@ -68,8 +69,6 @@ abstract contract Sphinx {
 
     SphinxUtils private sphinxUtils;
 
-    bool private isCollecting;
```
sam-goldman, line 71: We previously used this variable to determine whether to prank or broadcast in the `sphinx` modifier. We now always prank, so this variable isn't necessary anymore

packages/contracts/contracts/foundry/Sphinx.sol
```diff
@@ -96,20 +95,25 @@ abstract contract Sphinx {
         vm.makePersistent(address(sphinxUtils));
     }
 
-    function sphinxCollectProposal(string memory _deploymentInfoPath) external {
+    function sphinxCollectProposal(
+        string memory _deploymentInfoPath,
+        string memory _systemContractsFilePath
```
sam-goldman, line 100: The main update to the collection functions is that they contain the Merkle leaf gas estimation code, which requires that the Sphinx system contracts and the user's Gnosis Safe are deployed. (Explanation in the docs below)

packages/contracts/contracts/foundry/Sphinx.sol
```diff
@@ -124,14 +127,30 @@ abstract contract Sphinx {
...
     function sphinxCollect(
         ExecutionMode _executionMode,
-        address _executor
+        address _executor,
+        string memory _systemContractsFilePath
     ) private returns (DeploymentInfo memory) {
+        SystemContractInfo[] memory systemContracts = abi.decode(
+            vm.parseBytes(vm.readFile(_systemContractsFilePath)),
+            (SystemContractInfo[])
+        );
```
sam-goldman, line 146: If you're wondering why we read/write these to/from the file system, see [these docs](https://github.com/sphinx-labs/sphinx/pull/1403/files#diff-a10a154e6c3c80ec7f19d201fb8f80e05be0abc3fe778cfeca1675cfe95cbd19R1117-R1124). (This code was previously in `Sphinx.sol:sphinxEstimateMerkleLeafGas`

packages/contracts/contracts/foundry/Sphinx.sol
```diff
@@ -303,63 +340,53 @@ abstract contract Sphinx {
...
+            //    inherently a source of variation because there's a delay between the simulation
+            //    and execution.
+            // 2. Foundry's simulation is treated as a single transaction, which means SLOADs are
+            //    more likely to be "warm" (i.e. cheaper) than the production environment, where
+            //    transactions may be split between batches.
+            //
+            // We chose to multiply the gas by 1.3 because multiplying it by a higher number could
+            // make a very large transaction unexecutable on-chain. Since the 1.3x multiplier
+            // doesn't impact small transactions very much, we add a constant amount of 20k too.
+            gasEstimates[i] = 20_000 + ((startGas - finalGas) * 13) / 10;
```
sam-goldman, line 385: This code was previously in the TypeScript `sphinxEstimateMerkleLeafGas` function, which no longer exists. I figured it makes the most sense to put it here.

packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -389,7 +391,7 @@ contract SphinxUtils is SphinxConstants, StdUtils {
         // library for details.
         bytes memory proxyBytecode = hex"67363d3d37363d34f03d5260086018f3";
 
-        address proxy = computeCreate2Address(_salt, keccak256(proxyBytecode), _deployer);
+        address proxy = vm.computeCreate2Address(_salt, keccak256(proxyBytecode), _deployer);
```
sam-goldman, line 394: I had to prepend `vm` because Forge [recently deprecated the standard `computeCreate2Address` function](https://github.com/foundry-rs/forge-std/blob/master/src/StdUtils.sol#L106-L114)

packages/contracts/contracts/foundry/SphinxConstants.sol
```diff
@@ -7,6 +7,7 @@ contract SphinxConstants {
   string public constant sphinxLibraryVersion = 'v0.17.1';
   address public constant compatibilityFallbackHandlerAddress = 0xf48f2B2d2a534e402487b3ee7C18c33Aec0Fe5e4;
   address public constant multiSendAddress = 0xA238CBeb142c10Ef7Ad8442C6D1f9E89e07e7761;
+  address public constant createCallAddress = 0x7cbB62EaA69F79e6873cD1ecB2392971036cFAa4;
```
sam-goldman, line 10: This is the address of the `CreateCall` contract (written by Gnosis Safe) that we'll use to execute `CREATE` deployments from the Gnosis Safe.

packages/contracts/test/TestUtils.t.sol
```diff
@@ -70,6 +70,15 @@ import {
...
 
+struct SphinxTransaction {
+    address to;
+    uint256 value;
+    bytes txData;
+    IEnum.GnosisSafeOperation operation;
+    uint256 gas;
+    bool requireSuccess;
+}
+
```
sam-goldman, line 81: Moved this from `SphinxPluginTypes.sol` because we now only use it in tests and not in the plugin. (It was previously used in `Sphinx.sol:sphinxEstimateMerkleLeafGas`. We now use the `GnosisSafeTransaction` struct instead because it contains fewer fields. Was necessary to merge the collection and gas estimation logic)

packages/core/src/actions/types.ts
```diff
@@ -6,16 +6,6 @@ import { SphinxPreview } from '../preview'
...
-/**
- * Possible action types.
- */
-export const SphinxActionType = {
-  SET_STORAGE: BigInt(0),
-  DEPLOY_CONTRACT: BigInt(1),
-  CALL: BigInt(2),
-  CREATE: BigInt(3),
-}
-
```
sam-goldman, line 18: Super old field that was still existed in the `ActionInput` TypeScript types. Removed it because it's not necessary anymore

packages/core/src/config/types.ts
```diff
@@ -64,9 +64,10 @@ export type ParsedVariable =
       [name: string]: ParsedVariable
     }
 
-export type RawActionInput = RawFunctionCallActionInput | RawCreate2ActionInput
```
sam-goldman, line 67: The `RawActionInput` types are no longer necessary, so I removed them. Previously, we used them as an intermediate data type between the collection and gas estimation phase

packages/core/src/preview.ts
```diff
@@ -208,15 +208,15 @@ export const getPreview = (
         executing.push({
           referenceName: 'GnosisSafe',
           functionName: 'deploy',
-          variables: [],
+          variables: {},
```
sam-goldman, line 211: Changed this in a bunch of places to b/c the default structure of the `decodedActions.variables` is an object instead of an array as of the [previous PR](https://github.com/sphinx-labs/sphinx/pull/1401). (This is because we use the `recursivelyConvertResult` function)

packages/core/src/utils.ts
```diff
@@ -988,24 +947,32 @@ export const getSphinxWalletPrivateKey = (walletIndex: number): string => {
...
-  // The caller of the transactions on the Gnosis Safe will be the Gnosis Safe itself. This is
-  // necessary to prevent the calls from reverting.
-  const safeSigner = await getImpersonatedSigner(safeAddress, provider)
+  // The caller of the transactions on the Gnosis Safe will be the Sphinx Module. This is necessary
+  // to prevent the calls from reverting. An alternative approach is to call the Gnosis Safe from
+  // the Gnosis Safe itself. However, this would increment its nonce in the `addOwnerWithThreshold`
+  // calls that occur later in this function, which would mess up the addresses of contracts
+  // deployed via the Gnosis Safe. We can't reset the Gnosis Safe's nonce via `hardhat_setNonce`
+  // because Hardhat throws an error if we attempt to set a contract's nonce lower than its current
+  // nonce.
```
sam-goldman, line 960: Had to change the implementation of this function and `removeSphinxWalletsFromGnosisSafeOwners` to avoid this edge case

packages/plugins/src/cli/deploy.ts
```diff
@@ -185,17 +176,8 @@ export const deploy = async (
...
-  // Gnosis Safe as the `msg.sender` to ensure that it's the caller for all transactions. We need
-  // to do this even though we also broadcast from the Safe's address in the script.
-  // Specifically, this is necessary if the user is deploying a contract via CREATE2 that uses a
-  // linked library. In this scenario, the caller that deploys the library would be Foundry's
-  // default sender if we don't set this environment variable. Note that `FOUNDRY_SENDER` has
-  // priority over the `--sender` flag and the `DAPP_SENDER` environment variable.
-  const dateBeforeForgeScriptCollect = new Date()
-  const spawnOutput = await spawnAsync('forge', forgeScriptCollectArgs, {
-    FOUNDRY_SENDER: safeAddress,
-  })
```
sam-goldman, line 198: Had to remove the `FOUNDRY_SENDER` here and in the Propose command to avoid an edge case that occurred because I merged the collection and gas estimation steps. It's fine to remove the `FOUNDRY_SENDER` because we don't support linked libraries anyways.

If you're curious, the edge case is that Foundry throws an error if we attempt to deploy a contract at the same address as the `FOUNDRY_SENDER`. In the context of this Forge script call, we were setting the `FOUNDRY_SENDER` to be the Gnosis Safe, and we're also deploying the Gnosis Safe in `Sphinx.sol:sphinxCollect`.

had to remove the `FOUNDRY_SENDER` b/c of a bug where the `createProxyWithNonce`
  // would fail b/c a deployed address can't be the sender. was happening on `createProxyWithNonce`,
  // where we deploy the safe.

packages/plugins/src/cli/deploy.ts
```diff
@@ -214,51 +196,7 @@ export const deploy = async (
...
-    deploymentInfo.chainId,
-    `sphinxCollectDeployment`,
-    dateBeforeForgeScriptCollect
-  )
-
-  // Check if the dry run file exists. If it doesn't, this must mean that the deployment is empty.
-  // We return early in this case.
-  if (!dryRunFile) {
-    spinner.info(`Nothing to deploy. Exiting early.`)
-    return {}
```
sam-goldman, line 235: Moved this later in the function because we need the `actionInputs` to determine whether the deployment is empty.

packages/plugins/src/cli/deploy.ts
```diff
@@ -276,20 +214,24 @@ export const deploy = async (
...
   const isSystemDeployed = await checkSystemDeployed(provider)
   const parsedConfig = makeParsedConfig(
     deploymentInfo,
-    actionInputs,
-    gasEstimates,
     isSystemDeployed,
     configArtifacts,
-    dryRunFile.libraries,
-    dryRunPath
+    [] // We don't currently support linked libraries.
```
sam-goldman, line 227: We don't have the `libraries` available to us anymore since we don't use the broadcast file, which is why I hardcoded an empty array here

packages/plugins/src/cli/propose/index.ts
```diff
@@ -136,24 +134,8 @@ export const buildParsedConfigArray: BuildParsedConfigArray = async (
...
-    // user defining it in their `.env` file.
-    // It's worth mentioning that we can't run a single Forge script for all networks using
-    // cheatcodes like `vm.createSelectFork`. This is because we use the `FOUNDRY_SENDER`.
-    // Specifically, the state of the Safe on the first fork is persisted across all forks
-    // when using `FOUNDRY_SENDER`. This is problematic if the Safe doesn't have the same
-    // state across networks. This is a Foundry quirk; it may be a bug.
-    const dateBeforeForgeScript = new Date()
-    const spawnOutput = await spawnAsync('forge', forgeScriptCollectArgs, {
-      FOUNDRY_SENDER: safeAddress,
-    })
```
sam-goldman, line 156: Removes the `FOUNDRY_SENDER`. See [this comment](https://github.com/sphinx-labs/sphinx/pull/1403/files#r1477154007) from the Deploy command for an explanation

packages/plugins/src/foundry/decode.ts
```diff
@@ -92,127 +95,21 @@ export const decodeDeploymentInfo = (
...
-      } else {
-        throw new Error(`Unknown transaction type: ${transactionType}.`)
-      }
-    }
-  }
-
-  return actionInputs
-}
-
 export const makeParsedConfig = (
```
sam-goldman, line 111: I recommend just reading the new `makeParsedConfig` function instead of trying to decipher the diff between the old and new version

packages/plugins/src/foundry/decode.ts
```diff
@@ -347,46 +325,8 @@ export const makeParsedConfig = (
...
-  return {
-    parsedContracts,
-    unlabeledAdditionalContracts: unlabeled,
-  }
-}
-
-export const makeCreate2DecodedAction = (
-  create2Address: string,
+export const makeContractDecodedAction = (
+  contractAddress: string,
```
sam-goldman, line 329: Renamed this because we use it for `CREATE` deployments now too

packages/plugins/src/foundry/decode.ts
```diff
@@ -451,11 +390,7 @@ export const makeFunctionCallDecodedAction = (
     const functionName = decoded ? decoded.functionName : 'call'
     const variables = decoded
       ? decoded.variables
-      : [
-          data.length > 1000
-            ? `Very large calldata. View it in Foundry's dry run file: ${dryRunPath}`
-            : data,
-        ]
+      : [data.length > 1000 ? `Calldata is too large to display.` : data]
```
sam-goldman, line 393: We don't execute a dry run anymore, so there's no dry run path to include here

packages/plugins/src/foundry/utils/index.ts
```diff
@@ -1176,18 +1097,138 @@ export const findFullyQualifiedNameForInitCode = (
  */
 export const findFullyQualifiedNameForAddress = (
```
sam-goldman, line 1098: I defined this function in the previous PR but left it empty because I needed the `accountAccesses` to write it

packages/plugins/test/mocha/common.ts
sam-goldman, line 1: The changes to this file stem from the PR's removal of the `RawActionInput` type, which were previously used to define create2 deployment actions for our TypeScript tests. We now use `AccountAccess` objects for this purpose

packages/plugins/test/mocha/foundry/decode.spec.ts
```diff
@@ -34,7 +34,7 @@ describe('Decoded Actions', () => {
     },
   }
 
-  describe('makeCreate2DecodedAction', () => {
+  describe('makeContractDecodedAction', () => {
```
sam-goldman, line 37: This function applies to both `create` and `create2` deployments, which is why I changed the function's name

packages/plugins/test/mocha/common.ts
```diff
@@ -10,7 +10,6 @@ import {
   ConfigArtifacts,
   DeploymentInfo,
   ExecutionMode,
-  RawActionInput,
```
sam-goldman, line 13: The changes to this file stem from the PR's removal of the `RawActionInput` type, which were previously used to define create2 deployment actions for our TypeScript tests. We now use `AccountAccess` objects for this purpose

sam-goldman:
@RPate97 CI is failing for some reason related to the `forge-std` dependency, but this PR is ready for review

packages/plugins/src/foundry/decode.ts
RPate97: I'd be ok with removing this and just displaying the full calldata no matter what. Since we're only ever going to display the call data if we aren't able to find the artifact for the contract, I don't think it'll be very common for anyone to hit this anyway.

packages/plugins/src/foundry/utils/index.ts
```diff
@@ -1176,18 +1097,163 @@ export const findFullyQualifiedNameForInitCode = (
...
+    if (accountAccess.accessor === safeAddress) {
+      if (BigInt(accountAccess.value) > BigInt(0)) {
+        throw new Error(
+          `Detected value transfer in script:\n` +
+            `To: ${accountAccess.account}\n` +
+            `Amount: ${ethers.parseEther(accountAccess.value.toString())}\n` +
+            `Sphinx does not support transferring native gas tokens during deployments.\n` +
+            `Let us know if you want this feature!`
+        )
+      }
```
RPate97, line 1164: Why isn't this allowed? I don't see any reason why we can't support sending value from the safe to some other address as long as the safe is funded. I thought the limitation in the readme was wrt to airdropping tokens to an unfunded safe.

RPate97 (approved):
Looks pretty good, mostly just left comments and a few questions. I'm going to approve this for the sake of time, but I do strongly recommend addressing the comments I left related to organizing the AccountAccess array so it's easier to reason about.

packages/contracts/contracts/foundry/Sphinx.sol
sam-goldman: From the `write-initcode.ts` docs:
```
Including it in the production Foundry plugin would *significantly* slow down the user's compilation process if they're using Yul (i.e. `viaIR`) with the optimizer enabled.
```

If I remember correctly, it took minutes to compile the user's script. It was significant enough to warrant a messier pattern like this. 

In the future, I think we should move this to the TypeScript code that's executed before we collect the user's transactions

packages/contracts/contracts/foundry/Sphinx.sol
sam-goldman: > Hypothetically, if this code did not work and we erroneously underestimated the gas cost for a transaction. Would that be caught during the current simulation step (which uses real providers)? Could that be a good reason to maintain the use of real providers in the simulation?

Great point. I added that situation to the [relevant ticket](https://linear.app/chugsplash/issue/CHU-479/investigate-how-to-make-the-simulation-more-reliable) for making the simulation more reliable.

> It's worth noting that an alternative way to cover this edge case could be with a simulation using real providers that happens in our backend.

Yeah, I think that would work. I'd need to hear more about the benefit of using our backend though, since I don't think I've encountered a situation where a public node provider underestimates the gas

sam-goldman: Oh, you meant that the Hardhat simulation would be more reliable with the real node providers in our backend, didn't you

packages/core/src/config/types.ts
sam-goldman: Got it, good to know. I'll make a ticket for updating Mean Finance's integration (and anybody else's)

packages/plugins/src/foundry/decode.ts
sam-goldman: I'll leave it as is to speed up this PR, but feel free to add that to the [Improve Preview ticket](https://linear.app/chugsplash/issue/CHU-324/improve-preview)

.changeset/large-turkeys-explain.md
```diff
@@ -0,0 +1,8 @@
+---
+'@sphinx-labs/plugins': minor
+'@sphinx-labs/contracts': patch
+'@sphinx-labs/core': patch
+'@sphinx-labs/demo': patch
+---
+
+Add support for `CREATE` opcode deployments
```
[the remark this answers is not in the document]
sam-goldman: Sounds good. We could revisit [this ticket](https://linear.app/chugsplash/issue/CHU-351/make-doc-explaining-what-causes-a-breaking-change-to-the-website) as a first step

packages/core/src/config/types.ts
```diff
@@ -142,17 +145,22 @@ export type SphinxConfig = {
   saltNonce: string
 }
 
-export interface RawCreate2ActionInput extends SphinxTransaction {
+export enum ActionInputType {
+  CREATE,
+  CREATE2,
```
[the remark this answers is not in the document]
sam-goldman: The main benefit of having an `ActionInputType` for `CREATE2` is that it lets us display these deployments in a readable format in the preview. Otherwise, it'd look something like this:
```
(0x4e59...).call(0x00000....)
```

This is currently what `CREATE3` deployments look like, which is why I actually think we should have an `ActionInputType` for these types of deployments too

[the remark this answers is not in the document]
sam-goldman: I understand these are breaking changes and it's not clear how we could scale this pattern to many different action types, but my main point is that they serve a purpose

packages/contracts/contracts/foundry/Sphinx.sol
RPate97: Yeah I just meant doing essentially the same thing we currently do in the proposal task, but handle it in our backend so the proposal is faster and more reliable.

packages/contracts/contracts/foundry/Sphinx.sol
```diff
@@ -124,14 +127,30 @@ abstract contract Sphinx {
...
+        SystemContractInfo[] memory systemContracts = abi.decode(
+            vm.parseBytes(vm.readFile(_systemContractsFilePath)),
+            (SystemContractInfo[])
+        );
+
+        // Deploy the Sphinx system contracts. This is necessary because several Sphinx and Gnosis
+        // Safe contracts are required to deploy a Gnosis Safe, which itself must be deployed
+        // because we're going to call the Gnosis Safe to estimate the gas. Also, deploying the
+        // Gnosis Safe ensures that its nonce is treated like a contract instead of an EOA.
+        sphinxUtils.deploySphinxSystem(systemContracts);
```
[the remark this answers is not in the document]
sam-goldman: You're right - we now just use this in `sphinxCollectDeployment`

packages/contracts/contracts/foundry/Sphinx.sol
```diff
@@ -161,15 +180,40 @@ abstract contract Sphinx {
...
-        isCollecting = true;
+        // Deploy the Gnosis Safe if it's not already deployed. This is necessary because we're
+        // going to call the Gnosis Safe to estimate the gas.
+        if (address(safe).code.length == 0) {
+            // Deploy the Gnosis Safe and Sphinx Module. It's not strictly necessary to prank the
+            // Managed Service contract, but this replicates the prod environment for the DevOps
+            // Platform, so we do it anyways.
+            vm.startPrank(constants.managedServiceAddress());
+            _sphinxDeployModuleAndGnosisSafe();
+            vm.stopPrank();
```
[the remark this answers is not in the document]
sam-goldman: Removed the prank on the ManagedService

packages/contracts/contracts/foundry/Sphinx.sol
```diff
@@ -161,15 +180,40 @@ abstract contract Sphinx {
...
+
+        vm.startStateDiffRecording();
         // Delegatecall the `run()` function on this contract to collect the transactions. This
         // pattern gives us flexibility to support function names other than `run()` in the future.
         (bool success, ) = address(this).delegatecall(abi.encodeWithSignature("run()"));
         // Throw an error if the deployment script fails. The error message in the user's script is
         // displayed by Foundry's stack trace, so it'd be redundant to include the data returned by
         // the delegatecall in our error message.
         require(success, "Sphinx: Deployment script failed.");
+        deploymentInfo.accountAccesses = vm.stopAndReturnStateDiff();
```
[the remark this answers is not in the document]
sam-goldman: Updated in this [PR](https://github.com/sphinx-labs/sphinx/pull/1415/files)

packages/plugins/src/foundry/utils/index.ts
sam-goldman: You're right - fixed in [this PR](https://github.com/sphinx-labs/sphinx/pull/1415/files)

PR #1404, opened 2024-02-04
chore(deps): bump forge-std and use Sphinx fork of Foundry

sam-goldman:
The new `forge-std` version contains the `AccountAccess` structs, which are used in the state diff.


RPate97 (approved):
Looks good

PR #1406, opened 2024-02-04
feat(pg): throw error if Sphinx Foundry fork is not installed

sam-goldman:


packages/plugins/src/foundry/utils/index.ts
```diff
@@ -1191,3 +1191,25 @@ export const findFullyQualifiedNameForAddress = (
...
+
+  const match = stdout.match(/\((\w+)\s/)
+
+  if (match === null) {
+    throw new Error(`Foundry is not installed.`)
+  }
+
+  const foundryVersion = match[1]
+
+  if (foundryVersion !== 'fcb7c1e') {
```
RPate97, line 1208: I know you haven't requested a review on this, but we should discuss this further. I don't know if it's practical to use the commit hash to use this because it'll make it quite difficult for us to update the fork with changes from upstream.

RPate97: I think we might want to do some type of empirical check instead, like running script that tests if the state diff works the way that we expect it to work on our fork.

PR #1408, opened 2024-02-04
docs: fix npm sphinx installation command

sam-goldman:
I verified that these steps work on my machine

### DIFF
--- a/docs/troubleshooting-guide.md
+++ b/docs/troubleshooting-guide.md
@@ -6,6 +6,7 @@ This guide covers some common issues you might encounter using Sphinx. If your q
 
 - [Slow compilation speed](#slow-compilation-speed)
 - [Errors](#errors)
+  - [Installing Sphinx's Foundry fork](#installing-sphinxs-foundry-fork)
   - [`Ineffective mark-compacts near heap limit allocation failed`](#ineffective-mark-compacts-near-heap-limit-allocation-failed)
   - [`EvmError: MemoryLimitOOG`](#evmerror-memorylimitoog)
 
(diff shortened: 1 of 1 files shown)


docs/troubleshooting-guide.md
```diff
@@ -28,6 +29,31 @@ If you don't include this prefix in a `forge` command, Foundry will continue usi
 
 ## Errors
 
+### Installing Sphinx's Foundry fork
+
+If you're getting an error when installing Sphinx's Foundry fork, we recommend removing your existing Foundry installation and then re-installing Foundry.
+
+Here are the steps to do this:
```
RPate97, line 36:
```suggestion
Here are the steps to do this on MacOS:
```

RPate97, line 36:
```suggestion
Here are the steps to do this on macOS. These commands may be slightly different if you are using Linux or Windows.
```

docs/troubleshooting-guide.md
```diff
@@ -28,6 +29,31 @@ If you don't include this prefix in a `forge` command, Foundry will continue usi
...
+2. Install Foundryup:
+```
+curl -L https://foundry.paradigm.xyz | bash
+```
+
+3. Follow the instructions on-screen from the previous command to make `foundryup` available in your CLI.
+
+4. Install Sphinx's fork of Foundry:
+```
+foundryup --repo sphinx-labs/foundry --branch prerelease-v0.0.1
```
RPate97, line 52: If we end up doing a lot of different updates to our foundry fork, we'll probably want a setup similar to what we have for our contract library. Since our plan is for this to be a short-term measure, though, I think it's fine not to implement that now. But if we end up keeping this pattern for a while, we'll have to do it eventually.

RPate97 (approved):
Looks good, just left some minor comments.

docs/troubleshooting-guide.md
sam-goldman: @RPate97 I added this:

> Note: The following instructions will be slightly different if you're not using MacOS or Linux.

ChatGPT said the commands should be fine for Linux, and Foundry doesn't mention anything about using different commands for Linux. Lmk if you disagree, but otherwise I'll merge this

RPate97: I'd be amazed if this worked out of the box on all linux distros. I don't think ChatGPT is capable of assessing that to be honest. That being said, I don't think we have to necessarily call out that the user may need to do some troubleshooting on Linux. It's reasonable to assume a Linux user is capable of that.

PR #1410, opened 2024-02-04
maint: re-enable test that use live network rpc urls

sam-goldman:


packages/core/test/convert.spec.ts
```diff
@@ -88,9 +89,7 @@ describe('Convert EthersJS Objects', () => {
         throw new Error(`Unsupported chain ID: ${chainId}`)
       }
 
-      try {
-        fetchURLForNetwork(BigInt(chainId))
-      } catch {
+      if (!isLiveNetworkRpcApiKeyDefined(BigInt(chainId))) {
```
sam-goldman, line 92: These lines serve the same purpose except `isLiveNetworkRpcApiKeyDefined` doesn't indiscriminately catch errors, which was an issue caused by the `try/catch`

packages/core/test/convert.spec.ts
```diff
@@ -111,8 +110,8 @@ describe('Convert EthersJS Objects', () => {
     )
   })
 
-  describe('convertEthersTransactionReceipt', () => {
-    it('succeeds on anvil', async () => {
+  describe('succeeds on anvil', () => {
+    it('convertEthersTransactionReceipt', async () => {
```
sam-goldman, line 114: I switched the structure of the tests in this file slightly to make it easier to skip the live network tests in CI. The tests themselves remain unchanged

packages/core/test/convert.spec.ts
```diff
@@ -131,10 +130,45 @@ describe('Convert EthersJS Objects', () => {
...
+    before(function () {
+      // Skip the tests if the environment variable `CIRCLE_BRANCH` is defined and does not equal
+      // 'develop', which enforces that these tests only run in CI when the source branch is
+      // 'develop'. These tests will also run on local machines because the `CIRCLE_BRANCH`
+      // environment variable isn't defined.
+      const CIRCLE_BRANCH = process.env.CIRCLE_BRANCH
+      if (typeof CIRCLE_BRANCH === 'string' && CIRCLE_BRANCH !== 'develop') {
+        console.log('Skipping tests since this is not the develop branch')
+        this.skip()
+      }
```
sam-goldman, line 165: This snippet is the only functional difference in this file

packages/core/test/convert.spec.ts
```diff
@@ -145,9 +179,11 @@ describe('Convert EthersJS Objects', () => {
         const rpcUrl = fetchURLForNetwork(chainId)
         const provider = new SphinxJsonRpcProvider(rpcUrl)
 
+        let timeoutId: NodeJS.Timeout
```
sam-goldman, line 182: The `timeoutId` prevents the tests from stalling

packages/plugins/test/mocha/common.ts
```diff
@@ -339,53 +340,64 @@ export const makeDeployment = async (
   const networkNames = mainnets.concat(testnets)
 
   const collectedPromises = networkNames.map(async (networkName) => {
-    const chainId = fetchChainIdForNetwork(networkName)
```
sam-goldman, line 342: The diff below looks messy, but I just wrapped this code block in a try/catch so that we can identify which network caused an error (if one occurs)

packages/plugins/test/mocha/common.ts
```diff
@@ -407,7 +419,7 @@ export const makeDeployment = async (
 
   const parsedConfigArray = collected.map(
     ({ actionInputs, deploymentInfo }) => {
-      const gasEstimatesArray = actionInputs.map(() => (5_000_000).toString())
+      const gasEstimatesArray = actionInputs.map(() => (2_000_000).toString())
```
sam-goldman, line 422: Necessary to lower this from 5 million to 2 million to make the Rootstock simulation pass (due to the low block gas limit)

packages/plugins/test/mocha/simulate.spec.ts
```diff
@@ -1,118 +1,132 @@
-// import chai, { expect } from 'chai'
```
sam-goldman, line 1: This file is unchanged except there's some code that skips running these tests if on a feature branch in CI

RPate97 (approved):
Looks good, pending rebase. 

he one thing that is mildly concerning about this is that we are currently using public rpc nodes for some of these tests. I think this is probably ok since we're not running these tests regularly, but we should be careful not to abuse them. For example, f we were to have these tests run periodically (i.e scheduled once per day), it would probably be irresponsible to use public nodes.

sam-goldman:
> Looks good, pending rebase.
> 
> he one thing that is mildly concerning about this is that we are currently using public rpc nodes for some of these tests. I think this is probably ok since we're not running these tests regularly, but we should be careful not to abuse them. For example, f we were to have these tests run periodically (i.e scheduled once per day), it would probably be irresponsible to use public nodes.

Agreed, made a [ticket](https://linear.app/chugsplash/issue/CHU-508/use-private-rpc-endpoints-in-ci) for that

PR #1415, opened 2024-02-05
refactor(pg): parse `AccountAccess` structs

sam-goldman:
This PR updates the `AccountAccess` code according to [this comment](https://github.com/sphinx-labs/sphinx/pull/1403#discussion_r1477470848):
> I would consider organizing [the AccountAccesses] upfront so we don't have to reason so much about the contents of this array.
> 
> I would organize them into an array of GnosisSafeAccess structs like the one below. Each GnosisSafeAccess struct would represent an account access that we know is going to be executed during the deployment process (it's a create or call access, comes from the safe address, etc). We'd store that initial AccountAccess in the access field. The nestedAccesses field would then store all of the accesses that are nested within that call. So we have them available to us to search for nested contract deployments and whatnot.
> 
> I think this would help us reason about this array much more easily, and prevent us from having to deal with filtering the array multiple times later.

The plugins test suite is failing because the 'Proposes large deployment' test case is failing with the following error:
```
    RangeError: Maximum call stack size exceeded
      at String.match (<anonymous>)
      at _getBytes (/Users/samgoldman/dev/sphinx/node_modules/ethers/src.ts/utils/data.ts:34:45)
      at getBytesCopy (/Users/samgoldman/dev/sphinx/node_modules/ethers/src.ts/utils/data.ts:66:12)
      at new Reader (/Users/samgoldman/dev/sphinx/node_modules/ethers/src.ts/abi/coders/abstract-coder.ts:424:34)
      at AbiCoder.decode (/Users/samgoldman/dev/sphinx/node_modules/ethers/src.ts/abi/abi-coder.ts:209:29)
      at decodeDeploymentInfo (src/foundry/decode.ts:48:59)
      at Object.buildParsedConfigArray (src/cli/propose/index.ts:142:48)
      at processTicksAndRejections (node:internal/process/task_queues:95:5)
      at async propose (src/cli/propose/index.ts:246:5)
      at async Context.<anonymous> (test/mocha/cli/propose.spec.ts:337:7)
```
This error happens when we ABI decode the `DeploymentInfo` object in the TypeScript `decodeDeploymentInfo` function. It's caused by the fact that the `deploymentInfo.accountAccesses` is now an array of arrays instead of a flat array, which increases its ABI encoded size by ~13.5x.

I'll fix this error in the next PR.


packages/contracts/contracts/foundry/Sphinx.sol
```diff
@@ -95,13 +96,15 @@ abstract contract Sphinx {
     }
 
     function sphinxCollectProposal(
-        string memory _deploymentInfoPath
+        string memory _deploymentInfoPath,
+        string memory _systemContractsFilePath
```
sam-goldman, line 100: Proposals no longer load and deploy the Sphinx system contracs

packages/contracts/contracts/foundry/Sphinx.sol
```diff
@@ -125,6 +128,19 @@ abstract contract Sphinx {
             revert("Incorrect execution type.");
         }
 
+        DeploymentInfo memory deploymentInfo = sphinxCollect(
+            _executionMode,
+            deployer,
+            _systemContractsFilePath
+        );
+        vm.writeFile(_deploymentInfoPath, vm.toString(abi.encode(deploymentInfo)));
```
sam-goldman, line 136: Now, we only deploy the system contracts in `sphinxCollectDeployment`

packages/contracts/contracts/foundry/Sphinx.sol
```diff
@@ -136,10 +135,7 @@ abstract contract Sphinx {
         // Gnosis Safe ensures that its nonce is treated like a contract instead of an EOA.
         sphinxUtils.deploySphinxSystem(systemContracts);
 
-        DeploymentInfo memory deploymentInfo = sphinxCollect(
-            _executionMode,
-            deployer
-        );
+        DeploymentInfo memory deploymentInfo = sphinxCollect(_executionMode, deployer);
```
sam-goldman, line 138: linted

sam-goldman, line 138: linted

packages/contracts/contracts/foundry/Sphinx.sol
```diff
@@ -94,9 +95,7 @@ abstract contract Sphinx {
         vm.makePersistent(address(sphinxUtils));
     }
 
-    function sphinxCollectProposal(
-        string memory _deploymentInfoPath
-    ) external {
+    function sphinxCollectProposal(string memory _deploymentInfoPath) external {
```
sam-goldman, line 98: linted

packages/contracts/contracts/foundry/Sphinx.sol
sam-goldman, line 138: linted

packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -886,65 +887,108 @@ contract SphinxUtils is SphinxConstants, StdUtils {
...
+        for (uint256 i = _rootIdx + 1; i < _accesses.length; i++) {
+            Vm.AccountAccess memory access = _accesses[i];
+            if (!isRootAccountAccess(access, _safeAddress)) {
+                count += 1;
+            }
+        }
+        return count;
+    }
+
+    function parseAccountAccesses(
```
sam-goldman, line 930: This is the main function that parses the account accesses. I added some tests in `SphinxUtils.t.sol` for edge cases. We already have the necessary coverage for normal deployments in `deploy.spec.ts`

packages/contracts/test/SphinxInitCode.sol
```diff
@@ -4,22 +4,64 @@ pragma solidity >=0.6.2 <0.9.0;
 import { SystemContractInfo } from "../contracts/foundry/SphinxPluginTypes.sol";
 
 contract SphinxInitCode {
-  function getSystemContractInfo() public pure returns (SystemContractInfo[] memory) {
```
sam-goldman, line 7: linted

packages/plugins/src/foundry/utils/index.ts
```diff
@@ -1171,28 +1187,30 @@ export const assertValidAccountAccesses = (
...
 
-  const { create2Address } = decodeDeterministicDeploymentProxyData(
-    accountAccess.data
-  )
+  const { create2Address } = decodeDeterministicDeploymentProxyData(root.data)
 
   // The next `AccountAccess` must be a `Create` kind, and must be sent from the Deterministic
   // Deployment Proxy to the expected `CREATE2` address.
   return (
-    nextAccountAccess !== undefined &&
```
sam-goldman, line 1195: not necessary anymore b/c `nextAccountAccess` was previously possibly undefined, which isn't the case anymore

sam-goldman, line 1195: not necessary anymore b/c `nextAccountAccess` was previously possibly undefined, which isn't the case anymore

sam-goldman, line 1195: not necessary anymore b/c `nextAccountAccess` was previously possibly undefined, which isn't the case anymore

README.md
```diff
@@ -74,7 +74,6 @@ A demo of the DevOps Platform is on the [website's landing page](https://sphinx.
 ## Current Limitations
 
 - You cannot deploy [libraries](https://docs.soliditylang.org/en/v0.8.24/contracts.html#libraries).
-- You cannot send ETH as part of a deployment.
```
sam-goldman, line 77: Context: In the previous PR, you left a [comment](https://github.com/sphinx-labs/sphinx/pull/1403#discussion_r1477545328) saying that we should allow transferring ETH from the Safe (as long as the Safe is funded). We now allow this.

I think we should remove this sentence from the "Current Limitations" section because we lifted the restriction. I figured it'd be somewhat arbitrary to say "We don't airdrop tokens to your Safe" because it's a random feature that nobody's requested

RPate97: I agree, no one expects airdrops to be supported.

RPate97 (approved):
Looks good

PR #1417, opened 2024-02-06
fix(pg): Empirically check if Foundry fork installed

RPate97:
## Purpose
Checks if the foundry fork is installed by testing the functionality our fork fixes.


packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -47,6 +52,47 @@ contract SphinxUtils is SphinxConstants, StdUtils {
...
 
+    function checkAccesses(Vm.AccountAccess[] memory accountAccesses) internal returns (bool) {
+        // Deploy the contract using create to get the deployed code
+        SphinxForkCheck forkCheck = new SphinxForkCheck();
+
+        // If there aren't two calls, (one to the deployment proxy, and another to deploy the contract)
+        // then return false.
+        if (accountAccesses.length < 2) {
+            return false;
+        }
```
RPate97, line 63: If/when a real user encounters this, they are basically guaranteed to hit this first check. I only do the ones below to be extra cautious.

packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -3,6 +3,7 @@ pragma solidity ^0.8.0;
 
 import { Vm } from "../../lib/forge-std/src/Vm.sol";
 import { StdUtils } from "../../lib/forge-std/src/StdUtils.sol";
+import { console } from "../../lib/forge-std/src/console.sol";
```
sam-goldman, line 6: Should be removed I think

packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -47,6 +52,49 @@ contract SphinxUtils is SphinxConstants, StdUtils {
         return _data[_start:_end];
     }
 
+    function checkAccesses(Vm.AccountAccess[] memory accountAccesses) internal returns (bool) {
+        // Deploy the contract using create to get the deployed code
+        SphinxForkCheck forkCheck = new SphinxForkCheck();
```
sam-goldman, line 57: Can replace this with `type(SphinxForkCheck).runtimeCode`

packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -47,6 +52,49 @@ contract SphinxUtils is SphinxConstants, StdUtils {
...
+        // If the second access does not record deploying the contract to the correct address, then return false.
+        address expectedAddress = vm.computeCreate2Address(
+            0,
+            keccak256(type(SphinxForkCheck).creationCode),
+            DETERMINISTIC_DEPLOYMENT_PROXY
+        );
+        if (accountAccesses[1].account != expectedAddress) {
+            return false;
+        }
+
```
sam-goldman, line 79: Should we also check that `accountAccesses[1].accessor` is the `DETERMINISTIC_DEPLOYMENT_PROXY`? Not sure if this is important

packages/plugins/src/foundry/utils/index.ts
```diff
@@ -1203,13 +1206,16 @@ export const assertSphinxFoundryForkInstalled = async (
     throw new Error(`Foundry is not installed.`)
   }
```
sam-goldman, line 1207: I don't think we need to check this anymore. Previously, this function used `forge --version` to get the commit hash, which isn't used anymore

packages/contracts/test/SphinxUtils.t.sol
```diff
@@ -14,6 +15,154 @@ import {
 contract SphinxUtils_Test is Test, SphinxUtils {
     function setUp() public {}
 
+    function test_checkFork_passes() external {
+        assertEq(checkFork(), true);
+    }
+
+    function test_sphinxCheckAccesses_fails_did_not_use_default_factory() external {
+        // Deploy the contract so we can get the correct deployed code
+        SphinxForkCheck forkCheck = new SphinxForkCheck();
```
sam-goldman, line 24: We can also use `type(SphinxForkCheck).runtimeCode` here

packages/contracts/test/SphinxUtils.t.sol
```diff
@@ -14,6 +15,154 @@ import {
...
+            data: type(SphinxForkCheck).creationCode,
+            reverted: false,
+            storageAccesses: new Vm.StorageAccess[](0)
+        });
+
+        bool passedCheck = checkAccesses(accountAccesses);
+        assertEq(passedCheck, false);
+    }
+
+    function test_sphinxCheckAccesses_fails_incorrect_deployed_code() external {
```
sam-goldman, line 65: The name of this function indicates that it's supposed to be testing for incorrect deployed code, but it looks like it's actually testing that an incorrect `accountAccesses[1].account` results in `checkAccesses` returning `false`

packages/contracts/test/SphinxUtils.t.sol
```diff
@@ -14,6 +15,154 @@ import {
...
+        // Deploy the contract so we can get the correct deployed code
+        SphinxForkCheck forkCheck = new SphinxForkCheck();
+        address expectedAddress = vm.computeCreate2Address(
+            0,
+            keccak256(type(SphinxForkCheck).creationCode),
+            DETERMINISTIC_DEPLOYMENT_PROXY
+        );
+
+        Vm.AccountAccess[] memory accountAccesses = new Vm.AccountAccess[](2);
+        accountAccesses[0] = VmSafe.AccountAccess({
```
sam-goldman, line 32: In the future, it'd be helpful if you could document what's supposed to be different about test cases like this that involve boilerplate. Takes some effort for me to figure out that e.g. `accountAccesses[0].account` is the variable that we're testing here

packages/contracts/test/SphinxUtils.t.sol
```diff
@@ -14,6 +15,154 @@ import {
...
+            data: type(SphinxForkCheck).creationCode,
+            reverted: false,
+            storageAccesses: new Vm.StorageAccess[](0)
+        });
+
+        bool passedCheck = checkAccesses(accountAccesses);
+        assertEq(passedCheck, false);
+    }
+
+    function test_sphinxCheckAccesses_fails_incorrect_address() external {
```
sam-goldman, line 108: It's not clear what this function is testing. Which variable is different?

sam-goldman (commented):
I'm running into a couple issues

First, I'm on the `prerelease-v0.0.1` fork, and I'm running into this error when running `deploy.spec.ts`:
> Error: Detected invalid Foundry version. Please use Sphinx's fork of Foundry by
running the command:
foundryup --repo sphinx-labs/foundry --branch sphinx-patch-v0.1.0

Second, I tried updating to the latest fork, and got this error:
> \> foundryup --repo sphinx-labs/foundry --branch sphinx-patch-v0.1.0
>
> fatal: couldn't find remote ref sphinx-patch-v0.1.0
foundryup: command failed: git fetch origin sphinx-patch-v0.1.0:remotes/origin/sphinx-patch-v0.1.0

sam-goldman:
The empirical check looks solid, but I don't have enough context about the Foundry bugs to know whether this function covers everything we need to check. I trust that you've covered the things we need to check

sam-goldman:
Also, it looks like `always_use_create_2_factory = true` isn't specified in the plugins foundry.toml. Not sure if that's causing the CI tests to fail, but figured I'd lyk in case

sam-goldman:
Last comment: what happens if the user doesn't have `always_use_create_2_factory = true` in their `foundry.toml`? Will this check throw an error telling them to update their Foundry branch? If so, we should probably check `always_use_create_2_factory` separately (e.g. in `getFoundryToml`)

packages/contracts/test/SphinxUtils.t.sol
RPate97: So this is the check this function is designed to test: 
```
// If the deployed code at the calculated address is incorrect, then return false.
// This confirms the deterministic deployment proxy is in fact being used for the
// internal simulation.
if (keccak256(address(expectedAddress).code) != keccak256(type(SphinxForkCheck).runtimeCode)) {
   return false;
}
```

Note that this check *does not* check if the `deployedCode` field is correct in the account access object, but rather it checks that the code at the expected address is correct. This confirms that the deterministic deployment proxy is actually being used by the internal simulation. 

Since we do not actually deploy anything at the expected address in this test, it the deployed code will not be correct and the above check will cause `checkAccesses` to return false.

I do think the name of the test function is misleading though, so I’ll update it.

RPate97:
> Last comment: what happens if the user doesn't have `always_use_create_2_factory = true` in their `foundry.toml`? Will this check throw an error telling them to update their Foundry branch? If so, we should probably check `always_use_create_2_factory` separately (e.g. in `getFoundryToml`)

This check will fail. I agree we should check this separately. I was planning to do that in another PR after this.

packages/contracts/contracts/foundry/SphinxUtils.sol
RPate97: Yeah, we can there’s no harm. For context in a real world scenario where our fork is not installed on the users machine, the `checkFork` function is pretty much guaranteed to return false due to the first check (that the length is not correct). These additional checks are really just sanity checks that cover the possibility that someone might make a change to Foundry that causes the first check to no longer return false if our fork is not installed. In that case, we want to make sure we have sufficient coverage to confirm that Foundry is operating the way we expect it to operate. 

Technically it would be better to check as many things as possible in the `checkAccesses` field, but then we’d need tests and whatnot for all of the checks which would make this take a lot longer. I just added the set of checks that are sufficient for me to feel comfortable that if they all pass it’s extremely unlikely to be for any reason other than our fork being installed.

packages/contracts/test/SphinxUtils.t.sol
RPate97: See inline docs

packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -49,6 +53,54 @@ contract SphinxUtils is SphinxConstants, StdUtils {
...
+        return true;
+    }
+
+    function checkFork(address safeAddress) public returns (bool) {
+        vm.startPrank(safeAddress);
+        vm.startStateDiffRecording();
+        new SphinxForkCheck{ salt: 0 }();
+        Vm.AccountAccess[] memory accountAccesses = vm.stopAndReturnStateDiff();
+
+        return checkAccesses(accountAccesses);
```
RPate97, line 101: We have to prank the safes address here or we get an error like the one in the screenshot below. I believe this is probably a bug in foundry triggered by the fact that I’m calling into `SphinxUtils` from `Sphinx` and then starting the state diff recording within `SphinxUtils`. I don’t think this will cause other problems for us, but just wanted to point it out in case you run into it.

sam-goldman (commented):
Looks great, but I switched to the default foundry fork via `foundryup`, then ran the first test in `deploy.spec.ts`, which led to this error:

```
The application panicked (crashed).
Message:  missing CALL account accesses
Location: /Users/runner/work/foundry/foundry/crates/cheatcodes/src/inspector.rs:1002

This is a bug. Consider reporting it at https://github.com/foundry-rs/foundry

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
                                ⋮ 8 frames hidden ⋮                               
   9: core::option::expect_failed::h055bb418ffad3409
      at <unknown source file>:<unknown line>
  10: revm::inspector::_::<impl revm::inspector::Inspector<DB> for &mut T>::call_end::h2425ef58cd8ebae6
      at <unknown source file>:<unknown line>
  11: <revm::evm_impl::EVMImpl<GSPEC,DB> as revm_interpreter::host::Host>::call::h02c194a8ddae1580
      at <unknown source file>:<unknown line>
  12: revm_interpreter::instructions::host::call_inner::h067e5fb825047732
      at <unknown source file>:<unknown line>
  13: revm::inspector::instruction::inspector_instruction::{{closure}}::h024bb4888e7a193b
      at <unknown source file>:<unknown line>
  14: revm::evm_impl::EVMImpl<GSPEC,DB>::run_interpreter::h0750fad5de3dce64
      at <unknown source file>:<unknown line>
  15: <revm::evm_impl::EVMImpl<GSPEC,DB> as revm_interpreter::host::Host>::call::h02c194a8ddae1580
      at <unknown source file>:<unknown line>
  16: revm::evm_impl::EVMImpl<GSPEC,DB>::transact_preverified_inner::h394d2d5cb636b951
      at <unknown source file>:<unknown line>
  17: <revm::evm_impl::EVMImpl<GSPEC,DB> as revm::evm_impl::Transact<<DB as revm_primitives::db::Database>::Error>>::transact::h9abb5af983c736ca
      at <unknown source file>:<unknown line>
  18: foundry_evm::executors::Executor::call_raw::h9e316397606b377f
      at <unknown source file>:<unknown line>
  19: forge::cmd::script::runner::ScriptRunner::call::h1b1e1904b4771d40
      at <unknown source file>:<unknown line>
  20: forge::cmd::script::executor::<impl forge::cmd::script::ScriptArgs>::execute::{{closure}}::h147130fcddcb3302
      at <unknown source file>:<unknown line>
  21: forge::cmd::script::cmd::<impl forge::cmd::script::ScriptArgs>::run_script::{{closure}}::h8a3fd927481694ac
      at <unknown source file>:<unknown line>
  22: foundry_cli::utils::block_on::h4d9f0ec2e187febc
      at <unknown source file>:<unknown line>
  23: forge::main::h469aee4d72cc16d6
      at <unknown source file>:<unknown line>
  24: std::sys_common::backtrace::__rust_begin_short_backtrace::h18db68be3590f519
      at <unknown source file>:<unknown line>
  25: std::rt::lang_start::h3357cb7fc1c26948
      at <unknown source file>:<unknown line>
  26: _main<unknown>
      at <unknown source file>:<unknown line>

Run with COLORBT_SHOW_HIDDEN=1 environment variable to disable frame filtering.
```

RPate97:
> Looks great, but I switched to the default foundry fork via `foundryup`, then ran the first test in `deploy.spec.ts`, which led to the error below. The expected result is that we throw an error message saying, "pls install our Foundry fork"
> 
> The error:
> 
> ```
> The application panicked (crashed).
> Message:  missing CALL account accesses
> Location: /Users/runner/work/foundry/foundry/crates/cheatcodes/src/inspector.rs:1002
> 
> This is a bug. Consider reporting it at https://github.com/foundry-rs/foundry
> 
>   ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
>                                 ⋮ 8 frames hidden ⋮                               
>    9: core::option::expect_failed::h055bb418ffad3409
>       at <unknown source file>:<unknown line>
>   10: revm::inspector::_::<impl revm::inspector::Inspector<DB> for &mut T>::call_end::h2425ef58cd8ebae6
>       at <unknown source file>:<unknown line>
>   11: <revm::evm_impl::EVMImpl<GSPEC,DB> as revm_interpreter::host::Host>::call::h02c194a8ddae1580
>       at <unknown source file>:<unknown line>
>   12: revm_interpreter::instructions::host::call_inner::h067e5fb825047732
>       at <unknown source file>:<unknown line>
>   13: revm::inspector::instruction::inspector_instruction::{{closure}}::h024bb4888e7a193b
>       at <unknown source file>:<unknown line>
>   14: revm::evm_impl::EVMImpl<GSPEC,DB>::run_interpreter::h0750fad5de3dce64
>       at <unknown source file>:<unknown line>
>   15: <revm::evm_impl::EVMImpl<GSPEC,DB> as revm_interpreter::host::Host>::call::h02c194a8ddae1580
>       at <unknown source file>:<unknown line>
>   16: revm::evm_impl::EVMImpl<GSPEC,DB>::transact_preverified_inner::h394d2d5cb636b951
>       at <unknown source file>:<unknown line>
>   17: <revm::evm_impl::EVMImpl<GSPEC,DB> as revm::evm_impl::Transact<<DB as revm_primitives::db::Database>::Error>>::transact::h9abb5af983c736ca
>       at <unknown source file>:<unknown line>
>   18: foundry_evm::executors::Executor::call_raw::h9e316397606b377f
>       at <unknown source file>:<unknown line>
>   19: forge::cmd::script::runner::ScriptRunner::call::h1b1e1904b4771d40
>       at <unknown source file>:<unknown line>
>   20: forge::cmd::script::executor::<impl forge::cmd::script::ScriptArgs>::execute::{{closure}}::h147130fcddcb3302
>       at <unknown source file>:<unknown line>
>   21: forge::cmd::script::cmd::<impl forge::cmd::script::ScriptArgs>::run_script::{{closure}}::h8a3fd927481694ac
>       at <unknown source file>:<unknown line>
>   22: foundry_cli::utils::block_on::h4d9f0ec2e187febc
>       at <unknown source file>:<unknown line>
>   23: forge::main::h469aee4d72cc16d6
>       at <unknown source file>:<unknown line>
>   24: std::sys_common::backtrace::__rust_begin_short_backtrace::h18db68be3590f519
>       at <unknown source file>:<unknown line>
>   25: std::rt::lang_start::h3357cb7fc1c26948
>       at <unknown source file>:<unknown line>
>   26: _main<unknown>
>       at <unknown source file>:<unknown line>
> 
> Run with COLORBT_SHOW_HIDDEN=1 environment variable to disable frame filtering.
> ```

Good catch, seems like this was related to the other issue I noticed where foundry does not like the use of the state diff cheat code if you’re not calling it from the top level contract. I’ve refactored the script to call the cheat code from within `Sphinx`. We’ll just want to keep in mind that we should not use this cheatcode in the `SphinxUtils` contract.

packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -50,6 +51,44 @@ contract SphinxUtils is SphinxConstants, StdUtils {
         return _data[_start:_end];
     }
 
+    function checkAccesses(Vm.AccountAccess[] memory accountAccesses, bytes32 creationCodeHash, bytes32 runtimeCodeHash) public view returns (bool) {
+        // If there aren't two calls, (one to the deployment proxy, and another to deploy the contract)
```
RPate97, line 55: We have to pass in the creation and runtime code because `Sphinx.sol` (which deploys the test contract) is compiler by the users machine whereas `SphinxUtils.sol` is compiled by us and injected with raw bytecode. So if we tried to get the creation and runtime code directly from `SphinxForkTest` within `SphinxUtils.sol` it would not match the code for the corresponding contract in `Sphinx.sol`.

sam-goldman:
Getting this warning when compiling the plugins package. (This will show up for users)

> Warning (2072): Unused local variable.
   --> ../contracts/contracts/foundry/Sphinx.sol:102:9:
    |
102 |         address addr = address(new SphinxForkCheck{ salt: 0 }());
    |         ^^^^^^^^^^^^

sam-goldman (approved):
Looks good, just a few minor things that should be addressed before merging

PR #1419, opened 2024-02-06
fix(pg): resolve 'max depth exceeded' error for large deployments

sam-goldman:
Removes the "max depth exceeded" error described in the [previous PR](https://github.com/sphinx-labs/sphinx/pull/1415), which was happening when ABI decoding a large `DeploymentInfo` object. Now, we serialize the `DeploymentInfo` struct to JSON in Solidity instead of ABI encoding it.

This PR uses the `vm.serialize*` cheatcodes extensively, which are pretty unintuitive IMO. Here's a brief overview of the cheatcodes, which should hopefully save you some time.

Each `vm.serialize` cheatcode is in the format `vm.serialize(<object_key>, <value_key>, <value>)`.
* `<object_key>`: An arbitrary string that Foundry uses internally to store the JSON in memory. This key is *not* part of the actual JSON.
* `<value_key>`: An arbitary string, which is an actual key in the JSON.
* `<value>`: The value of the `<value_key>`.

Here's an example to illustrate:
```
string memory objKey = "myObjKey";
string memory json = vm.serializeUint(objKey, "myValueKey", "myVal"); // Creates: "myObjKey: { myValueKey: "myVal" }
console.log(json) // Logs "{ myValueKey: "myVal" }". Notice that "myObjKey" is not in the JSON.
```

Here are the [Foundry docs](https://book.getfoundry.sh/cheatcodes/serialize-json) for the `serializeJson` cheatcodes.

### DIFF
--- a/packages/contracts/contracts/foundry/Sphinx.sol
+++ b/packages/contracts/contracts/foundry/Sphinx.sol
@@ -15,7 +15,7 @@ import {
     HumanReadableAction,
     Network,
     SphinxConfig,
-    DeploymentInfo,
+    FoundryDeploymentInfo,
     NetworkInfo,
     Wallet,
     GnosisSafeTransaction,
@@ -98,12 +98,12 @@ abstract contract Sphinx {
     function sphinxCollectProposal(string memory _deploymentInfoPath) external {
         sphinxUtils.validateProposal(sphinxConfig);
 
-        DeploymentInfo memory deploymentInfo = sphinxCollect(
+        string memory serializedDeploymentInfo = sphinxCollect(
             ExecutionMode.Platform,
             constants.managedServiceAddress()
         );
 
-        vm.writeFile(_deploymentInfoPath, vm.toString(abi.encode(deploymentInfo)));
+        vm.writeFile(_deploymentInfoPath, serializedDeploymentInfo);
     }
 
     function sphinxCollectDeployment(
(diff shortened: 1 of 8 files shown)


packages/contracts/contracts/foundry/Sphinx.sol
```diff
@@ -15,7 +15,7 @@ import {
     HumanReadableAction,
     Network,
     SphinxConfig,
-    DeploymentInfo,
+    FoundryDeploymentInfo,
```
sam-goldman, line 18: Renamed the Solidity `DeploymentInfo` to `FoundryDeploymentInfo` to distinguish it from the TypeScript `DeploymentInfo` object. See the [docs for `FoundryDeploymentInfo`](https://github.com/sphinx-labs/sphinx/pull/1419/files#diff-3b9851f6e1331971546aeb627e15507982bec64d50995404788145fb06cdf709R80-R91) for an explanation

sam-goldman:
The plugins tests are now faster than the demo tests :o

sam-goldman:
I checked that the demo tests pass locally, so don't worry about the CI error

packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -929,6 +934,114 @@ contract SphinxUtils is SphinxConstants, StdUtils {
...
+            deploymentInfoKey,
+            "encodedAccountAccesses",
+            _deployment.encodedAccountAccesses
+        );
+        // Next, we'll serialize `uint` values as ABI encoded bytes. We don't serialize them as
+        // numbers to prevent the possibility that they lose precision due JavaScript's relatively
+        // low integer size limit. We'll ABI decode these values in TypeScript. It'd be simpler to
+        // serialize the numbers as strings without ABI encoding them, but that strategy is blocked
+        // by a Foundry bug: https://github.com/foundry-rs/foundry/issues/6533
+        vm.serializeBytes(deploymentInfoKey, "nonce", abi.encode(_deployment.nonce));
```
RPate97, line 966: It’s probably worth thinking about if there’s a way we can prevent ourselves from introducing a new `uint` in the future which is not handled this way. I can’t think of anything off the top of my head, but I’m sure there’s something we could do. I don’t think it’s a huge deal, but it is definitely a foot-gun.

RPate97 (approved):
Looks good

packages/contracts/contracts/foundry/SphinxUtils.sol
sam-goldman: Agreed it's a footgun. I think it'd be pretty straightforward to check that there aren't any numbers in the `DeploymentInfo` object in TypeScript, but IMO it's not a big deal, since I think a mistake like this is unlikely to make it past both of us. Feel free to make a ticket if you disagree

packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -929,6 +934,114 @@ contract SphinxUtils is SphinxConstants, StdUtils {
         return count;
     }
 
+    function serializeFoundryDeploymentInfo(
+        FoundryDeploymentInfo memory _deployment
+    ) public returns (string memory) {
```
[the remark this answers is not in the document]
sam-goldman: Added some docs to clarify this

PR #1420, opened 2024-02-06
feat(pg): Check if create2 factory option not enabled

RPate97:
## Purpose
Adds checks for the required foundry.toml options.

### DIFF
--- a/packages/plugins/src/foundry/options.ts
+++ b/packages/plugins/src/foundry/options.ts
@@ -23,6 +23,29 @@ export const resolvePaths = (outPath: string, buildInfoPath: string) => {
   }
 }
 
+export const checkRequiredTomlOptions = (toml: FoundryToml) => {
+  if (toml.alwaysUseCreate2Factory !== true) {
+    throw new Error(
+      'Missing required option in foundry.toml file:\nalways_use_create_2_factory = true\nPlease update your foundry.toml file and try again.'
+    )
+  }
+
+  if (toml.buildInfo !== true) {
+    throw new Error(
+      'Missing required option in foundry.toml file:\nbuild_info = true\nPlease update your foundry.toml file and try again.'
+    )
+  }
+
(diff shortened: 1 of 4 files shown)


packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -35,6 +35,10 @@ import { IGnosisSafe } from "./interfaces/IGnosisSafe.sol";
 import { IMultiSend } from "./interfaces/IMultiSend.sol";
 import { IEnum } from "./interfaces/IEnum.sol";
 
+contract SphinxForkCheck {
+    constructor() {}
+}
+
```
sam-goldman, line 41: Seems like this shouldn't be here since it already exists in Sphinx.sol

packages/plugins/src/foundry/options.ts
```diff
@@ -23,6 +23,29 @@ export const resolvePaths = (outPath: string, buildInfoPath: string) => {
...
+export const checkRequiredTomlOptions = (toml: FoundryToml) => {
+  if (toml.alwaysUseCreate2Factory !== true) {
+    throw new Error(
+      'Missing required option in foundry.toml file:\nalways_use_create_2_factory = true\nPlease update your foundry.toml file and try again.'
+    )
+  }
+
+  if (toml.buildInfo !== true) {
+    throw new Error(
+      'Missing required option in foundry.toml file:\nbuildInfo = true\nPlease update your foundry.toml file and try again.'
```
sam-goldman, line 35: Should be `build_info`, not `buildInfo` (although, interestingly, specifying `buildInfo` in the foundry.toml works too)

sam-goldman (approved):
Looks good, just a couple minor things to fix before merging

packages/plugins/src/foundry/options.ts
sam-goldman: I'm referring to the error string here btw

PR #1422, opened 2024-02-08
fix(pg): throw error if script contains linked library

sam-goldman:


packages/plugins/src/cli/deploy.ts
```diff
@@ -75,6 +77,13 @@ export const deploy = async (
     targetContract,
   } = args
 
+  if (!isFile(scriptPath)) {
```
sam-goldman, line 80: Context: Previously, it was possible for the `scriptPath` to be a contract's fully qualified name, which lets the user avoid specifying `--target-contract` when there are multiple contracts in a single script. However, this undocumented pattern makes it more difficult to reason about the `findFullyQualifiedNames` function, which relies on an input `sourceName`. (The `findFullyQualifiedNames` function is necessary in this PR, and will be necessary when we support linked libraries.)

IMO, the cleanest way to handle this is to throw an error if the user attempts to set the `scriptPath` to be a fully qualified name, since this is undocumented anyways.

packages/plugins/src/cli/deploy.ts
```diff
@@ -222,6 +231,14 @@ export const deploy = async (
   )
   const configArtifacts = await getConfigArtifacts(initCodeWithArgsArray)
 
+  await assertNoLinkedLibraries(
```
sam-goldman, line 234: This comes after `await getConfigArtifacts` so that the build info cache is up to date. (The `assertNoLinkedLibraries` function uses the build info cache; more context later)

RPate97 (approved):
Looks good

PR #1423, opened 2024-02-08
feat(pg): Track compiler config major version

RPate97:
## Purpose
Adds a version number to the compiler config which is supplied to the website when the config is pinned. 

## Context
I’m adding logic to the website to ensure that we never accidentally attempt execute a compiler config which is not compatible with the current version of the executor. 

I also decided to make this release backwards compatible. We can do this by running two versions of the executor, one of the current version of the plugin, and one for the updated version. I decided to do this because Mean finance is currently in the process of doing some deployments using the old format and I would like those to go smoothly. I think this is the best way to do that. Once Mean Finance is done with their deployments, we will drop support for the old compiler config version. I also think this is good practice for when we inevitably have to do this while supporting more users.

### DIFF
--- a/packages/core/src/config/types.ts
+++ b/packages/core/src/config/types.ts
@@ -191,12 +191,15 @@ interface AbstractActionInput extends SphinxTransaction {
   index: string
 }
 
+export const COMPILER_CONFIG_VERSION = '0.1.0'
+
 /**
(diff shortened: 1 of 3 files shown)


sam-goldman (approved):
Looks good

PR #1431, opened 2024-02-08
fix(pg): prevent Foundry from serializing strings as numbers

sam-goldman:
Foundry converts strings like `projectName` to integers if the string contains all numbers. This is a [bug in Foundry](https://github.com/foundry-rs/foundry/issues/6533).

We now ABI encode all string fields in the `DeploymentInfo` object when we serialize it in Solidity. Then, in the TypeScript `decodeDeploymentInfo` function, we ABI decode the strings. We ABI encode the strings to ensure that Foundry doesn't serialize them as numbers, which could cause them to lose precision due to JavaScript's relatively low integer size limit.

Notes:
* I manually checked that every field in the `DeploymentInfo` will convert to its correct type
* Created a type predicate, `isDeploymentInfo`, to enforce that the fields are the correct type

### DIFF
--- a/packages/contracts/contracts/foundry/SphinxUtils.sol
+++ b/packages/contracts/contracts/foundry/SphinxUtils.sol
@@ -976,8 +976,8 @@ contract SphinxUtils is SphinxConstants, StdUtils {
 
     /**
      * @notice Serializes the `FoundryDeploymentInfo` struct. The serialized string is the
-     *         same structure as the `FoundryDeploymentInfo` struct except all `uint` fields
-     *         are ABI encoded (see inline docs for details).
+     *         same structure as the `FoundryDeploymentInfo` struct except all `uint` and `string`
+     *         fields are ABI encoded (see inline docs for details).
      */
     function serializeFoundryDeploymentInfo(
         FoundryDeploymentInfo memory _deployment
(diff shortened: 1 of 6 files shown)


packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -1022,6 +1017,15 @@ contract SphinxUtils is SphinxConstants, StdUtils {
...
         vm.serializeBytes(deploymentInfoKey, "gasEstimates", abi.encode(_deployment.gasEstimates));
+        // Serialize the Sphinx library version as an ABI encoded string. We ABI encode it to ensure
+        // that Foundry doesn't serialize it as a number, which will happen if the
+        // `sphinxLibraryVersion` consists only of numbers. If Foundry serializes it as a number,
+        // it'll be prone to the same precision loss due to JavaScript's low integer size limit.
+        vm.serializeBytes(
+            deploymentInfoKey,
+            "sphinxLibraryVersion",
+            abi.encode(_deployment.sphinxLibraryVersion)
+        );
```
RPate97, line 1028: Since we use semver versions (i.e `v1.0.0`), isn't this unnecessary? 

I don't think doing this is a big deal to be clear, I'm just curious about the extent of the bug we're working around.

RPate97 (approved):
Looks good

packages/contracts/contracts/foundry/SphinxUtils.sol
sam-goldman: It is unnecessary, but I intentionally did it this way out of an abundance of caution (e.g. maybe in the future we'll change it to a string that can contain all numbers)

PR #1432, opened 2024-02-09
fix(pg): Recommend using github action in CI

RPate97:
## Purpose
- Updates the docs to recommend using our [GitHub action](https://github.com/marketplace/actions/sphinx-foundry-toolchain) to install the Foundry fork in CI (note that this does mean other CI platforms than GitHub will not work, but I think that is an acceptable limitation for now).
- Update the `install` command to include the `ci` option that is explicitly intended to be used in CI and completely skips the installation of our fork (because the GitHub action handles this).

### DIFF
--- a/docs/ci-proposals.md
+++ b/docs/ci-proposals.md
@@ -2,9 +2,7 @@
 
 It's a best practice to propose deployments from a CI process instead of using the command line. This ensures that your deployments are reproducible and don't depend on a single developer's machine, which can be a source of bugs.
 
-This guide will show you how to integrate proposals into your CI process. We'll create two workflows: one that dry runs the proposal when a pull request is opened or updated, and another that proposes the deployment when the pull request is merged.
-
-We'll use GitHub Actions as the CI platform in this guide. You can still follow this guide if you're using a different CI platform, but the exact configuration may be slightly different.
+This guide will show you how to integrate proposals into your GitHub Actions CI process. Currently, Sphinx only officially supports GitHub Actions. If you need support for an alternative CI platform, please let us know. In this guide, we'll create two workflows: one that dry runs the proposal when a pull request is opened or updated, and another that proposes the deployment when the pull request is merged.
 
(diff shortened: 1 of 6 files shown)


packages/plugins/src/cli/install/index.ts
```diff
@@ -9,28 +9,27 @@ import { sphinxUp } from './sphinxup'
...
-  confirm: boolean
+  ci: boolean
 ) => {
-  if (!confirm) {
+  if (!ci) {
     await userConfirmation(
       'Would you like to install the Sphinx Foundry fork? This is required to use Sphinx and will replace your existing Foundry installation.\nConfirm? (y\\n):'
     )
-  }
+    const foundryupStatus = await spawnAsync('/bin/sh', [
```
sam-goldman, line 18: I've seen `/bin/bash` frequently, but I haven't seen `/bin/sh` before. Is this correct?

docs/ci-proposals.md
```diff
@@ -4,8 +4,6 @@ It's a best practice to propose deployments from a CI process instead of using t
 
 This guide will show you how to integrate proposals into your CI process. We'll create two workflows: one that dry runs the proposal when a pull request is opened or updated, and another that proposes the deployment when the pull request is merged.
 
-We'll use GitHub Actions as the CI platform in this guide. You can still follow this guide if you're using a different CI platform, but the exact configuration may be slightly different.
```
sam-goldman, line 7: Should we mention up-front that this guide only works for GitHub actions?

sam-goldman (approved):
Looks good, just a few comments/questions

docs/ci-proposals.md
RPate97: We can if you want, I'm not that worried about it.

packages/plugins/src/cli/install/index.ts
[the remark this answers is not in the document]
RPate97: Yeah it would be better to use `/bin/bash`.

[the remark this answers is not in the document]
RPate97: Well, below is specifically a bash script vs. this is just a curl command, which is then piped into bash. The command below probably wouldn't work with `sh` (since it's just a link to whatever the system shell is), but this command probably would since it just depends on cURL being installed. That said, it's probably better to use `/bin/bash` for consistency.

docs/ci-proposals.md
sam-goldman: Let's not bother then

PR #1436, opened 2024-02-10
fix(pg): set max block gas limit in Forge script during collection

sam-goldman:
Fixes the following edge case:
1. User specifies a `gas_limit` field in their `foundry.toml` ([docs](https://book.getfoundry.sh/reference/config/testing?highlight=gas_limit#gas_limit)). (xkeep3r sets it to 30M)
2. The `gas_limit` is used in the forge script that collects their transactions
3. An "out of gas" error occurs in the collection script

This didn't happen before in keep3r's integration because we used to collect the transactions in a separate Forge script from estimating the Merkle leaf gas. The 'out of gas' error occurs after estimating the Merkle leaf gas.

I manually checked that this fixes the issue in keep3r's integration

### DIFF
--- a/.changeset/twelve-baboons-swim.md
+++ b/.changeset/twelve-baboons-swim.md
@@ -1,0 +1,6 @@
+---
+'@sphinx-labs/plugins': patch
+'@sphinx-labs/core': patch
+---
+
+Set max block gas limit in Forge script during collection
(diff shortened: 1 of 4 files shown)


sam-goldman:
@RPate97 can you ping me when you approve this? I'm gonna make a release b/c I think this bug is decently likely to occur

RPate97 (approved):
Looks good

PR #1440, opened 2024-02-10
fix(pg): check mismatch between plugins package and contracts library version early

sam-goldman:
Fixes a bug that occurred when there was a mismatch between the Sphinx contract library version (installed via git submodules) and the npm plugin package.

Here are the steps that led to the bug:
1. The plugin version was slightly newer than the contracts version. Specifically, the plugin package included [this PR that prevented Foundry from serializing strings as numbers](https://github.com/sphinx-labs/sphinx/pull/1431), whereas the contracts package did not include this PR.
2. The `decodeDeploymentInfo` function in the plugin expected an ABI encoded `projectName` field ([source](https://github.com/sphinx-labs/sphinx/blob/main/packages/plugins/src/foundry/decode.ts#L99)), but the contract package was not ABI encoding this field because it didn't include the PR that serializes strings as numbers.
3. The `abiDecodeString` call in the `decodeDeploymentInfo` function failed because the `deploymentInfo.newConfig.projectName` field wasn't ABI encoded. This error occurred before the `checkLibraryVersion` function was called, since `checkLibraryVersion` runs after the `decodeDeploymentInfo` function.

I think this type of error is likely to occur again if we don't validate the library version very early in the process.

This PR does two things:
* It validates the library version very early in the process
* It creates a new Solidity function, `Sphinx.sol:sphinxValidate`, which will let us add new validation steps in the future in a backwards compatible manner.

The `sphinxValidate` function currently does two things:
1. It fetches the Sphinx contracts library version, and
2. It replaces `Sphinx.sol:sphinxCheckFork`, which empirically checked that our Foundry fork is functioning properly.

I merged these checks into a single Solidity call as a performance optimization and because it's a cleaner pattern IMO.

Going forward, we must keep the `sphinxValidate` function backwards compatible in the manner described in its Natspec docs.

There's one downside of this PR: if a user has a plugins package version that includes this PR and a contracts library version that doesn't, or vice versa, they'll see an opaque error. They can resolve it by updating the outdated package, but it won't be clear that they need to do that. It's generally a bad idea to introduce changes like this, but I think it's an acceptable tradeoff because this situation won't happen again, and the number of users that this impacts will be low (or perhaps zero).


RPate97 (approved):
Looks good

PR #1441, opened 2024-02-11
fix(pg): make simulation more reliable by using less recent block number

sam-goldman:
Previously, our simulation was always using the latest block, which was potentially causing it to be flaky. This PR attempts to make the simulation more reliable by using a less recent block number. Specifically, we use the same block number in the simulation and the collection phase.

An alternative approach is to use a block number that's a certain number of blocks behind the latest block number. I think it's an acceptable alternative, but less than ideal for a couple reasons:
1. Different networks have vastly different block times, so it may be tricky to choose a single number of blocks to rewind. E.g. Rootstock's block time is 33s whereas Arbitrum's seems to be multiple blocks per second.
2. Rewinding a certain number of blocks is prone to the following edge case, which is admittedly contrived:
  a. User executes a transaction on the live network
  b. User calls the Deploy or Propose command. The user's script relies on the transaction in the previous step.
  c. The collection phase succeeds because it uses the latest block number, but the simulation fails because it uses an older block number.

**Note**: This PR adds an additional field to the `ParsedConfig` (the `blockNumber`) which is potentially a breaking change. I checked that this shouldn't break the `verifySphinxConfig`, `executeDeployment`, and `buildDeploymentWithCompilerConfigs` functions. I could've implemented this without adding a field to the `ParsedConfig`, but it would've introduced complexity, which didn't seem worth it given that we're simply adding a new field to the `ParsedConfig`.

This PR should hopefully improve the reliability of the simulation, but I think we should implement retries to make it really reliable. ([Relevant ticket](https://linear.app/chugsplash/issue/CHU-447/implement-better-timeout-and-retry-logic-in-execution-logic))

### DIFF
--- a/.changeset/clever-vans-yell.md
+++ b/.changeset/clever-vans-yell.md
@@ -1,5 +1,0 @@
----
-'@sphinx-labs/plugins': patch
----
-
-Allow user to send native gas tokens from their Gnosis Safe
--- a/.changeset/hot-frogs-film.md
+++ b/.changeset/hot-frogs-film.md
@@ -1,0 +1,7 @@
+---
+'@sphinx-labs/contracts': patch
+'@sphinx-labs/plugins': patch
+'@sphinx-labs/core': patch
+---
+
+Make simulation more reliable by using less recent block number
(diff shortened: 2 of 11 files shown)


RPate97 (approved):
Looks good

PR #1443, opened 2024-02-11
fix(pg): normalize script path

sam-goldman:
This doesn't fix any particular bug, but it decreases the chance that [this type of bug](https://github.com/sphinx-labs/sphinx/pull/1425) happens in the future. This PR is simply a preventive measure.


RPate97:
Does this need to be rebased onto develop?

PR #1445, opened 2024-02-11
maint(core): Refactor simulation and exeuction logic to be shared

RPate97:
## Purpose
Refactors the execution logic to be shared between the proposal simulation, deploy command simulation/execution, and the website execution with the goal of reducing the chance that the proposal simulation will succeed and then the deployment fails during execution on the website. 

At a high level, what I've done in this PR is take the function used to execute deployments on the website, and I've moved it into the monorepo. I then refactored it so it could be reused outside of the website and updated the proposal simulation as well as the deployment simulation/execution logic to use it.

This does introduce some redundancy in the simulation/deployment logic because we're doing some extra work (i.e., recompiling all the contracts), but I view this as an inefficiency in the design of the execution flow used on the website that should be dealt with directly. I don't think it's a compelling reason not to reuse the same execution logic everywhere. I believe strongly that sticking to the same execution interface will help us write far more robust and reliable code.

Note that I've chosen to kick [CHU-446 back to triage](https://linear.app/chugsplash/issue/CHU-446/refactor-proposal-and-execution-process-to-only-compile-locally) for the moment. After spending a good amount of time thinking about it, I don't think it's going to improve the reliability of our system very significantly (assuming we reuse the same execution logic everywhere as is implemented in this PR). I believe it is largely a speed optimization. Since it's also a breaking change, I think we should wait on it until we have other breaking changes we'd like to ship with it.

## Review Recommendations
There's a lot going on in this PR that needs context. I recommend starting with the following before reviewing the rest of the PR:
- Taking a look through the [`compileAndExecuteDeployment`](https://github.com/sphinx-labs/sphinx/blob/40fc3291ab5f00c0db636b36d2cefc34fa7ee86b/packages/core/src/actions/execute.ts#L1059C14-L1059C41) function. The bulk of the changes in this PR are related to implementing this function and making it usable in the proposal simulation, deployment simulation + execution, and tests. I recommend reading the inline docs for the function.
- Then look read through the docs for the [`DeploymentContext`](https://github.com/sphinx-labs/sphinx/blob/pate/chu-444-make-simulation-use-same-logic-as-production-deployments/packages/core/src/actions/execute.ts#L155) object
- Finally, take a look at where the `compileAndExecuteDeployment` function is invoked using the `DeploymentContext` object in the [deploy command](https://github.com/sphinx-labs/sphinx/blob/pate/chu-444-make-simulation-use-same-logic-as-production-deployments/packages/plugins/src/cli/deploy.ts#L375) and [simulation](https://github.com/sphinx-labs/sphinx/blob/pate/chu-444-make-simulation-use-same-logic-as-production-deployments/packages/plugins/src/hardhat/simulate.ts#L319).

## Notes
- I also updated the compiler config upload process to send the config directly to S3 from the CLI instead of sending it through our backend. This isn't directly related to this PR, but I ran into a number of problems related to the compiler config size while testing this. I believe the state diff changes may have meaningfully increased the size of a typical compiler configuration, making it necessary to go ahead and make this change.
- I also updated the deployment process to not automatically cancel previous Drippie drip versions b/c canceling a previous version is a breaking change for the executor.


packages/core/src/actions/execute.ts
```diff
@@ -14,122 +13,277 @@ import {
...
-import { convertEthersTransactionReceipt } from '../artifacts'
+import { shouldBufferExecuteActionsGasLimit } from '../networks'
 import {
   SphinxTransactionReceipt,
   ensureSphinxAndGnosisSafeDeployed,
 } from '../languages'
-import { ParsedConfig } from '../config'
-import { shouldBufferExecuteActionsGasLimit } from '../networks'
-
-export const executeDeployment = async (
```
RPate97, line 49: I completely removed this function since the website will now also use the `compileAndExecuteDeployment` function.

packages/core/src/actions/execute.ts
```diff
@@ -321,24 +475,90 @@ export const approveDeploymentViaSigner: ApproveDeployment = async (
...
-    await sphinxModule.approve(
-      merkleRoot,
-      approvalLeafWithProof,
-      ownerSignature,
-      await getGasPriceOverrides(provider, signer, executionMode)
+export const executeTransactionViaSigner: ExecuteTransaction = async (
+  deploymentContext: DeploymentContext,
+  transaction: MinimumTransaction,
+  executionMode: ExecutionMode
+): Promise<ethers.TransactionReceipt> => {
```
RPate97, line 508: Used for sending transactions via an ethers signer in the deploy task and proposal simulation.

packages/core/src/actions/execute.ts
```diff
@@ -321,24 +475,90 @@ export const approveDeploymentViaSigner: ApproveDeployment = async (
...
+  if (txReceipt === null) {
+    throw new Error(
+      'No transaction receipt returned by ethers. This is a bug, please report it to the developers.'
+    )
+  }
+
+  return txReceipt
+}
+
+export const injectRoles: InjectRoles = async (
```
RPate97, line 537: Used to set the relayer role and gnosis safe owners in the simulation and when deploying on a local node.

packages/core/src/actions/execute.ts
```diff
@@ -496,19 +696,16 @@ export const executeActionsViaSigner: ExecuteActions = async (
...
+  return deploymentContext.executeTransaction(
+    deploymentContext,
+    {
+      to: moduleAddress,
+      data: executionData,
+      gasLimit: overrides.gasLimit?.toString(),
+      chainId: (await provider.getNetwork()).chainId.toString(),
+    },
+    executionMode
+  )
```
RPate97, line 708: It's not strictly necessary to use the `deploymentContext.executeTransaction` function here since `executeActionsViaSigner` won't get called on the website. However, I like to make things easy to reason about so in this PR I tried to follow the general rule of "if the transaction is valid on a live network, then use executeTransaction". I find that rule to be easier to reason about than trying to think about whether each specific transaction will need to be executed on the website or not. 

For this reason, I recommend we generally bias towards always using the deploymentContext.executeTransaction function even if it's not always strictly necessary. That being said, this PR doesn't fully follow that anyway (ensureSphinxAndGnosisSafeDeployed doesn't use it for example). 

I mainly think it will save you time and mental effort to just use this function all the time since you don't really work on the website backend at all.

packages/core/src/languages/solidity/predeploys.ts
```diff
@@ -281,19 +257,71 @@ export const deploySphinxSystem = async (
...
+
+  if (includeManagedServiceRoles) {
+    await assignManagedServiceRoles(
+      provider,
+      signer,
+      relayers,
+      executionMode,
+      spinner
+    )
+  }
```
RPate97, line 324: Previously, the `ensureSphinxAndGnosisSafeDeployed` function would fail on live networks since all the logic for configuring the managed service requires a specific signer. So, I went ahead and refactored this so that it only runs when we deploy the contracts ourselves.

packages/plugins/contracts/test/RevertDuringSimulation.sol
```diff
@@ -0,0 +1,29 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+interface ISphinxModule {
+    function activeMerkleRoot() external view returns (bytes32);
+}
+
+contract RevertDuringSimulation {
```
RPate97, line 8: I had to move this contract out of the `RevertDuringSimulation.s.sol` script file [due to this bug](https://linear.app/chugsplash/issue/CHU-407/deployment-fails-if-you-attempt-to-deploy-a-contract-which-is-in-the). 

This is actually a great example of a bug that would not currently be caught in our simulation logic since it only occurs b/c of the recompilation process. After this refactor, this bug started getting caught when we tried to execute this script with the `deploy` command in our tests so we have to move the contract to a different file to work around it.

packages/plugins/package.json
```diff
@@ -19,7 +19,7 @@
     "copy-hardhat-config": "cp src/hardhat.config.js dist/hardhat.config.js",
     "clean": "rimraf out/ dist/ ./tsconfig.tsbuildinfo ./forge-cache",
     "test": "yarn kill-nodes && yarn test:solc && yarn test:ts",
-    "test:ts": "npx mocha --require ts-node/register --timeout 500000 'test/mocha/**/*.ts'",
+    "test:ts": "forge build --force && npx mocha --require ts-node/register --timeout 500000 'test/mocha/**/*.ts'",
```
RPate97, line 22: The `build-info` file is now required for our tests since we recompile the contracts during the deploy/proposal simulation steps. To speed up the tests, I had them read in the file directly using JSON.parse, which is significantly faster than using the normal `makeGetConfigArtifacts` function. This requires exactly one build info file, though, so I added `--force` here to ensure that's always the case.

Unfortunately, the changes in this PR do result in our tests being slightly slower overall due to the added burden of recompiling all the contracts. With these changes, our plugin test suite now takes about 8.5 minutes vs 6.5 previously. It's not ideal, but it's not a terrible regression all things considered.

packages/plugins/test/mocha/mock.ts
```diff
@@ -10,17 +11,24 @@ import {
...
 import { propose } from '../../src/cli/propose'
 import { deploy } from '../../src/cli/deploy'
 import { makeSphinxContext } from '../../src/cli/context'
+import { readContractArtifact } from '../../dist'
+
+const file = readdirSync(`./out/artifacts/build-info/`).at(0)
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const buildInfo = JSON.parse(
+  readFileSync(`./out/artifacts/build-info/${file}`, 'utf8')
+)
```
RPate97, line 23: See comment above, we read this in directly now since it's faster than using the `makeGetConfigArtifacts` function.

packages/contracts/src/networks.ts
```diff
@@ -302,7 +302,7 @@ export const SPHINX_NETWORKS: Array<SupportedNetwork> = [
     dripSize: '0.15',
     requiredEnvVariables: ['CHIADO_RPC_URL'],
     networkType: 'Testnet',
-    dripVersion: 0,
+    dripVersion: 1,
```
RPate97, line 305: This is intentional. I ran into some problems when testing gnosis due to the maximum drip limit. So I bumped this and redeployed the contracts to resolve the issue.

packages/plugins/src/hardhat/simulate.ts
```diff
@@ -237,11 +249,93 @@ export const simulateDeploymentSubtask = async (
...
   }
 
-  const deploymentData = makeDeploymentData(parsedConfigArray)
+  const deploymentData = makeDeploymentData(compilerConfigArray)
   const merkleTree = makeSphinxMerkleTree(deploymentData)
 
-  const { receipts, batches, finalStatus, failureAction } =
-    await runEntireDeploymentProcess(parsedConfig, merkleTree, provider, signer)
+  await fundAccountMaxBalance(signer.address, provider)
+  await setManagedServiceRelayer(signer.address, provider)
```
sam-goldman, line 256: Seems like `setManagedServiceRelayer` is redundant since we already call it in `injectRoles`?

packages/plugins/src/cli/deploy.ts
```diff
@@ -277,13 +326,65 @@ export const deploy = async (
...
-    parsedConfig,
-    merkleTree,
+  const treeSigner = {
+    signer: await signer.getAddress(),
+    signature: await signMerkleRoot(merkleTree.root, signer),
+  }
+  const deployment: Deployment = {
+    id: 'only required on website',
+    multichainDeploymentId: 'only required on website',
+    projectId: 'only required on website',
```
sam-goldman, line 336: thoughts on making the types of these fields `string | undefined` or `string | null`? seems cleaner than saying 'only required on website' a bunch of times

sam-goldman (commented):
first pass, only reviewed half of it so far. gonna go to sleep and pick it back up tmrw. feel free to address the comments i've made so far

packages/plugins/src/hardhat/simulate.ts
```diff
@@ -237,11 +249,93 @@ export const simulateDeploymentSubtask = async (
...
+
+  // Create a list of auto-generated wallets. We'll later add these wallets as Gnosis Safe owners.
+  const sphinxWallets = getSphinxWalletsSortedByAddress(
+    BigInt(compilerConfig.newConfig.threshold),
+    provider
+  )
+  const treeSigners = await Promise.all(
+    sphinxWallets.map(async (wallet) => {
+      return {
+        signer: await wallet.getAddress(),
```
sam-goldman, line 266: nit: can be `wallet.address`, which is synchronous

packages/core/src/actions/execute.ts
```diff
@@ -14,122 +13,274 @@ import {
...
+  multichainDeploymentId: string
+  projectId: string
+  chainId: string
+  status:
+    | 'approved'
+    | 'cancelled'
+    | 'executed'
+    | 'verified'
+    | 'verification_unsupported'
+    | 'failed'
```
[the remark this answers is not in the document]
RPate97: It's not relevant. This field doesn't correspond to the status on-chain. It corresponds to the status of the deployment in our database where `empty` is not a valid option.

packages/core/src/actions/execute.ts
```diff
@@ -14,122 +13,274 @@ import {
...
+  configArtifacts: ConfigArtifacts,
+  networkName: string
+) => Promise<void>
+
+export type HandleExecutionFailure = (
+  deploymentContext: DeploymentContext,
+  networkName: string,
+  targetNetworkConfig: CompilerConfig,
+  configArtifacts: ConfigArtifacts,
+  failureAction: HumanReadableAction | undefined,
```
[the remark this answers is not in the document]
RPate97: No, it should never be undefined if the simulation fails.

packages/core/src/actions/execute.ts
```diff
@@ -14,122 +13,274 @@ import {
...
+   * Handles if the deployment has already been executed.
+   *
+   * We use the `DeploymentContext` for this because the website implements retry logic, as a result
+   * there are some cases where the deployment may be partially or fully executed already when we call
+   * the `compileAndExecuteDeployment` function. For the propose/deploy commands, this should never happen
+   * so we just error.
+   *
+   * It's worth noting that if we wanted to implement retry logic in the deploy cli command to make it more
+   * reliable, we would want to do that by implementing this function and then repeatedly calling the
+   * `compileAndExecuteDeployment` function. This is how the website currently works.
```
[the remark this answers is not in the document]
RPate97: We'll probably still want to have this in the context object even after implementing retries in the simulation/deploy command because the website handles retries by writing info to the database. So unless we're going to use a local db like sqlite, it probably wouldn't make very much sense to share that logic.

packages/plugins/src/cli/deploy.ts
RPate97: That would add a lot of complexity to the website because we'd have to include handling for cases where these fields are not defined since the type structure no longer enforces it. Given that, I think it's better to require them to be defined and just give them values that convey that they are not necessary for certain contexts.

packages/core/src/actions/execute.ts
```diff
@@ -669,109 +932,353 @@ export const runEntireDeploymentProcess = async (
...
+ * The allow for this function to be reused in all three situations this function accepts a `deploymentContext` object
+ * which implements all the logic that needs to be different in each of the three situations.
+ *
+ * We aim to *only* use the `deploymentContext` for logic which absolutely cannot be shared. For example, logic for
+ * writing information about a deployment to the website backend is implemented in the `deploymentContext`.
+ *
+ * See documentation on the `DeploymentContext` type at the top of this file for more information of the specific fields
+ * and justification for why the logic implemented in them cannot be shared.
+ */
+export const compileAndExecuteDeployment = async (
```
sam-goldman, line 1106: I think we should add docs somewhere (not sure where) saying something like:

> The `ExecutionMode.Platform` is used by the website in this function, which means it can't contain any local network operations for this mode, such as assigning roles to an arbitrary address. However, we can perform local network operations via dependency injection, as long as the injected dependency isn't used by the website. This is the purpose of the `injectRoles` function.

sam-goldman (commented):
Looks solid, just minor stuff

packages/plugins/src/cli/deploy.ts
```diff
@@ -277,13 +326,65 @@ export const deploy = async (
...
+    chainId: parsedConfig.chainId,
+    status: 'approved',
```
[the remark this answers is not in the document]
RPate97: Well, from the perspective of the website, `approved` means the user has submitted the signatures required to execute the deployment. `empty` doesn't make any sense in that context either. I sympathize with your viewpoint on this, but IMO, it's not worth the effort to refactor the website just to update this to a value that makes more sense in both situations.

packages/core/src/actions/execute.ts
```diff
@@ -721,23 +973,27 @@ export const runEntireDeploymentProcess = async (
...
-        merkleRootState.status
-      )}`
+    return handleStatus(
+      merkleRootState.status,
+      [],
+      ethersReceipts,
+      humanReadableActions[chainId.toString()][
+        Number(merkleRootState.leavesExecuted) - 2
+      ],
+      spinner
```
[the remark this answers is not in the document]
RPate97: Well we shouldn't be throwing errors within the `compilerAndExecuteDeployment` function. We need to call `deploymentContext.throwError` so that the error can be handled correctly depending on the context. Additionally, the status not being `APPROVED` in this situation doesn't mean we should throw an error at all. If we're repeatedly running the `compileAndExecuteDeployment` function (such as is the case when retrying), then the status could very well be either `COMPETED` or `FAILED` both of which require different handling. 

In the case of `FAILED`, we handle that after this function returns which is why we're just returning `handleStatus` here. In the case of `COMPLETED`, we want to just continue so we can do verification and artifact generation. The only case we don't handle is if the status is `CANCELED` which should not be the case unless the user manually intervenes while the deployment is happening. We should probably add handling for that, but it's outside the scope of this PR IMO.

packages/core/src/actions/execute.ts
```diff
@@ -750,28 +1006,241 @@ export const runEntireDeploymentProcess = async (
...
+      await deploymentContext.handleAlreadyExecutedDeployment(
+        deploymentContext,
+        targetNetworkConfig,
+        configArtifacts,
+        networkName
+      )
+
+      return
+    }
+  }
```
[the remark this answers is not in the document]
RPate97: Yeah, we want the website to be able to recover from any possible failure, including one where we failed to update the status of the deployment in the DB properly. So, we use the real on-chain state as a source of truth to determine whether the deployment was completed. 

For the same reason, when we later handle if the deployment failed. We do that by checking the on-chain state instead of relying fully on the data in the DB.

packages/core/src/actions/execute.ts
```diff
@@ -750,28 +1006,241 @@ export const runEntireDeploymentProcess = async (
...
+  let configArtifacts: ConfigArtifacts
+
+  // Handle if the config cannot be fetched
+  try {
+    ;({ merkleTree, compilerConfigs, configArtifacts } =
+      await buildDeploymentWithCompilerConfigs(rawCompilerConfigs!))
+  } catch (e: any) {
+    await deploymentContext.handleError(e, deployment, networkName, [])
+
+    return
```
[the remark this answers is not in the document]
RPate97: We only rethrow the error if `handleError` throws an error.  On the website, we just log an error and return. In long-lived processes, it's generally much better to just smoothly handle these sorts of things instead of actually throwing an error.

packages/core/src/actions/execute.ts
```diff
@@ -750,28 +1006,241 @@ export const runEntireDeploymentProcess = async (
...
+
+  // get active deployment ID for this project
+  const moduleAddress = deployment.moduleAddress
+  const module = new ethers.Contract(
+    moduleAddress,
+    SphinxModuleABI,
+    deploymentContext.provider
+  )
+
+  if ((await deploymentContext.provider.getCode(moduleAddress)) !== '0x') {
```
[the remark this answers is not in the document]
RPate97: We do attempt to verify in `handleAlreadyExecutedDeployment`. You could probably make an argument that we should refactor this so that we don't have to call `verify` within this function, but I think that would require a somewhat involved refactor. I would probably wait until we do the ticket to share the verification logic to improve this.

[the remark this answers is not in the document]
RPate97: It's just supposed to handle if the status is `COMPLETED`. You're correct in your previous suggestion to change the if statement to check the status directly. This is just left over from the first module implementation, which did not have an explicit `FAILURE` status.

packages/plugins/script/Sample.s.sol
```diff
@@ -11,37 +11,52 @@ contract Sample is Sphinx {
...
+            Network.evmos_testnet,
+            Network.kava_testnet,
+            Network.scroll_sepolia,
+            Network.zora_sepolia,
+            Network.rari_sepolia
+        ];
         sphinxConfig.mainnets = [Network.ethereum, Network.arbitrum];
         sphinxConfig.orgId = "clo6byksj0001cbld6lelntej";
-        sphinxConfig.saltNonce = 0;
+        sphinxConfig.saltNonce = 2112414;
```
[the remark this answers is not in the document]
RPate97: I mean changes to this file don't really matter since it's only used for testing. At least, I only use it for testing. I can revert the changes though, idc

RPate97:
@sam-goldman 

> This is going to slow down local development b/c I run yarn test:ts --grep <specific_test> a lot. Would really prefer not to re-compile the entire plugins contracts every time I do that. Any alternatives?

Ok, I've moved the force recompile to the `yarn test` command. You'll just want to keep in mind that if you're running the `test:ts` command directly tests may fail during simulation if there are multiple build info files in the plugins package. Personally, I'm not willing to load in all the build info files using the standard `makeGetConfigArtifacts` b/c that nearly doubles the runtime of the tests.

packages/plugins/src/cli/deploy.ts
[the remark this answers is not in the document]
sam-goldman: I think this is a footgun since 'approved' always means 'approved on-chain' in the monorepo. Imo calling it 'signed' would be safer. It's not worth stalling this PR over it, but just want to point it out

sam-goldman (approved):
Looks good

PR #1446, opened 2024-02-13
fix(pg): allow user to send native tokens from their deployer

sam-goldman:
I thought I removed this limitation when adding support for the `CREATE` opcode, but it must've slipped through somehow.


RPate97 (approved):
Looks good

PR #1447, opened 2024-02-13
feat(core): Do not recompile during execution

RPate97:
## Purpose
- Removes the recompilation step during execution; instead, we supply all of the necessary data in the compiler config.
- Changes the format of the `CompilerConfig` so it includes all necessary data for the deployment and only includes the compiler inputs and config artifacts once for the entire deployment. We now only supply a single compiler config object to the website. Previously, we used an array of compiler configs, so these values would have been duplicated if we kept that format.
- Renames `CompilerConfig` -> `DeploymentConfig` because it includes all the info for an entire deployment which may be executed across multiple networks
- Renames `ParsedConfig` -> `NetworkConfig` because it includes all info that may be specific to each network
- Adds logic to trim down the build info files, removing any unnecessary fields to minimize the size of the file we're uploading to the DB. 

## Notes
We could reduce the size of the file we're uploading to our backend further by avoiding duplication of the build info file (since we include one for every contract artifact in the `ConfigArtifacts` type). I chose not to do that as part of this PR, though, since this is already quite large. We don't technically need to do it since uploading files directly to s3 allows us to get away with having very large files. However, since it would be a breaking change, and it's not that hard, I'll probably go ahead and do that before releasing this. I'm just going to do it separately from this PR.


packages/core/hardhat.config.ts
```diff
@@ -72,7 +71,7 @@ task('deploy-system')
     ) => {
       // Throw an error if we're on the Hardhat network. This ensures that the `url` field is
       // defined for this network.
-      if (!isHttpNetworkConfig(hre.network.config)) {
+      if (!('url' in hre.network.config)) {
```
RPate97, line 74: Just removed the `isHttpNetworkConfig` because it required another type called `NetworkConfig` which conflicted with the new name for the parsed config which is `NetworkConfig`.

packages/core/src/actions/types.ts
```diff
@@ -127,6 +127,7 @@ export type ProposalRequest = {
   gasEstimates: Array<{ chainId: number; estimatedGas: string }>
   diff: SphinxPreview
   compilerConfigId: string | undefined
+  deploymentConfigId: string | undefined
```
RPate97, line 130: This is intentional and is for backward compatibility on the website.

packages/core/src/languages/solidity/types.ts
```diff
@@ -318,13 +317,9 @@ export interface SolcInput {
 
 export interface CompilerOutputContract {
   abi: Array<any>
-  storageLayout?: SolidityStorageLayout
   evm: {
     bytecode: CompilerOutputBytecode
     deployedBytecode: CompilerOutputBytecode
-    methodIdentifiers: {
-      [methodSignature: string]: string
-    }
```
RPate97, line 327: I trimmed out everything that did not appear to be necessary from the `BuildInfo` type and then just made sure all the tests still passed. I still need to do live network testing with this, so I'm not totally sure these changes are all good, but it seems fine. Let me know if anything seems off though.

packages/plugins/src/cli/propose/index.ts
```diff
@@ -392,27 +397,25 @@ export const propose = async (
...
   } else {
-    const compilerConfigId = await sphinxContext.storeCanonicalConfig(
+    const deploymentConfigId = await sphinxContext.storeDeploymentConfig(
       apiKey,
       newConfig.orgId,
       canonicalConfigData
     )
-    proposalRequest.compilerConfigId = compilerConfigId
+    proposalRequest.deploymentConfigId = deploymentConfigId
+    proposalRequest.compilerConfigId = deploymentConfigId
```
RPate97, line 418: Again, this is intentional for backward compatibility.

packages/plugins/src/foundry/utils/trim.ts
```diff
@@ -0,0 +1,143 @@
+import { BuildInfo } from '@sphinx-labs/core'
+
+export const trimObjectToType = <T>(obj: any, typeTemplate: T): T => {
```
RPate97, line 3: This function is used to trim down the build info file to exactly the fields we specify in our type. To do this, we use a template object which contains all the fields we want. We cannot use the typescript type itself since that doesn't exist at runtime. So, if we want to update the `BuildInfo` type, we also need to update the template here.

packages/plugins/test/mocha/mock.ts
```diff
@@ -69,36 +65,32 @@ export const makeMockSphinxContext = (
...
-      const files = readdirSync(`./out/artifacts/build-info/`)
-      if (!buildInfos) {
-        buildInfos = await Promise.all(
-          files.map((file) =>
-            JSON.parse(
-              readFileSync(`./out/artifacts/build-info/${file}`, 'utf8')
-            )
-          )
-        )
-      }
```
RPate97, line 81: We don't have to do this anymore since we don't recompile during execution, how wonderful.

packages/plugins/src/foundry/utils/trim.ts
```diff
@@ -0,0 +1,143 @@
...
+  // If the template is either an array then we just return since we don't do any filtering on arrays
+  if (Array.isArray(typeTemplate)) {
+    return obj
+  } else if (typeof typeTemplate !== 'object') {
```
sam-goldman, line 7: if `typeTemplate === null`, this function will error b/c `null` is of `object` type. (It'll error in the for-loop below)

packages/plugins/src/foundry/utils/trim.ts
```diff
@@ -0,0 +1,143 @@
+import { BuildInfo } from '@sphinx-labs/core'
+
+export const trimObjectToType = <T>(obj: any, typeTemplate: T): T => {
+  // If the template is either an array then we just return since we don't do any filtering on arrays
+  if (Array.isArray(typeTemplate)) {
+    return obj
```
sam-goldman, line 6: If i'm understanding this correctly, it seems this function doesn't trim any elements that exist inside of an array. Is that right? If so, we may want to either fix that or just add docs to this function mentioning this limitation

packages/core/src/config/types.ts
```diff
@@ -195,7 +196,10 @@ interface AbstractActionInput extends SphinxTransaction {
  * Config object with added compilation details. Must add compilation details to the config before
  * the config can be published or off-chain tooling won't be able to re-generate the deployment.
  */
-export interface CompilerConfig extends ParsedConfig {
+export interface DeploymentConfig {
+  networkConfigs: Array<NetworkConfig>
+  merkleTree: SphinxMerkleTree
+  configArtifacts: ConfigArtifacts
```
sam-goldman, line 202: Just want to mention there's a lot of redundancy here b/c the `configArtifacts` maps each fully qualified name to a separate copy of the build info. Could make this pretty large when there are a lot of unique contracts in the deployment. (I know that we trimmed the size of the build info significantly). Still, could the size of this be problematic when uploading to aws?

sam-goldman (commented):
Looks solid - just some minor comments / questions

Also, I think this could've potentially been split into two PRs: one that changes the names of the variables and another that removes the remote compilation process. It was tricky to sift through hundreds of lines of name changes to find the meaningful changes in this PR

packages/core/src/config/types.ts
RPate97: > We could reduce the size of the file we're uploading to our backend further by avoiding duplication of the build info file (since we include one for every contract artifact in the ConfigArtifacts type). I chose not to do that as part of this PR, though, since this is already quite large. We don't technically need to do it since uploading files directly to s3 allows us to get away with having very large files. However, since it would be a breaking change, and it's not that hard, I'll probably go ahead and do that before releasing this. I'm just going to do it separately from this PR.

packages/core/src/actions/types.ts
```diff
@@ -107,7 +107,7 @@ export type ContractInfo = {
 }
 
 /**
- * @param canonicalConfig Deprecated field.
+ * @param canonicalConfigId Deprecated field.
```
sam-goldman, line 110: nit: `canonicalConfigId` -> `compilerConfigId`

packages/plugins/src/foundry/utils/trim.ts
```diff
@@ -0,0 +1,158 @@
...
+  } else if (typeof typeTemplate !== 'object') {
+    // If the template is a primative type, then we just return so the value is included in the result
+    return obj
+  } else if (!typeTemplate) {
+    throw new Error('Cannot specify a typeTemplate that is null or undefined')
+  }
```
sam-goldman, line 25: This'll throw an error for null-ish `typeTemplate` values, like `0` and `''`

sam-goldman (approved):
Looks great, just a couple last minor comments

packages/plugins/src/foundry/utils/trim.ts
RPate97: That's actually not true because primative types that are falsy will be caught and returned in the if statement above, but we can specifically check for `null` here it is more specific.

PR #1451, opened 2024-02-14
fix(pg): Optimize BuildInfo storage in DeploymentConfig

RPate97:
## Purpose
Optimize the size of the DeploymentConfig by only including each build info object once.

### DIFF
--- a/packages/core/src/etherscan.ts
+++ b/packages/core/src/etherscan.ts
@@ -16,10 +16,15 @@ import { ChainConfig } from '@nomicfoundation/hardhat-verify/types'
 import { Etherscan } from '@nomicfoundation/hardhat-verify/etherscan'
 
 import { customChains } from './constants'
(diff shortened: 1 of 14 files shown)


packages/core/src/etherscan.ts
```diff
@@ -103,12 +113,15 @@ export const verifySphinxConfig = async (
...
 export const verifyDeploymentWithRetries = async (
-  networkConfig: NetworkConfig,
-  configArtifacts: ConfigArtifacts,
+  deploymentConfig: DeploymentConfig,
   provider: ethers.Provider,
   apiKey: string
 ): Promise<void> => {
   const maxAttempts = 10
+  const networkConfig = fetchNetworkConfigFromDeploymentConfig(
+    (await provider.getNetwork()).chainId,
```
sam-goldman, line 122: nit: We can pass the `chainId` into this function to remove this RPC call

sam-goldman (approved):
Looks good!

PR #1474, opened 2024-02-19
fix(ct): Resolve block.number bug on Arbitrum

RPate97:
## Purpose
Fixes [CHU-558](https://linear.app/chugsplash/issue/CHU-558/mainnet-simulation-fails-on-arbitrum)


sam-goldman (approved):
I'll approve this in the interest of time, but IMO it's unnecessary to pass the block number into Solidity just to pass it back into TypeScript. I think it'd be cleaner to just pass it as an input parameter to `decodeDeploymentInfo`

sam-goldman:
> I'll approve this in the interest of time, but IMO it's unnecessary to pass the block number into Solidity just to pass it back into TypeScript. I think it'd be cleaner to just pass it as an input parameter to `decodeDeploymentInfo`

If it matters, the Foundry `DeploymentInfo` and the TypeScript `DeploymentInfo` types will diverge when we implement linked libraries, so it's not necessary to maintain parity between these types

RPate97:
> I'll approve this in the interest of time, but IMO it's unnecessary to pass the block number into Solidity just to pass it back into TypeScript. I think it'd be cleaner to just pass it as an input parameter to `decodeDeploymentInfo`

addressed

PR #1479, opened 2024-02-20
fix(core): increase overridden gas limit on local and forked networks

sam-goldman:
Previously, it was possible to run into a "SphinxModule: insufficient gas" error in the simulation on Scroll. The error was occurring because we always hardcode the transaction's `gasLimit` to equal 75% of the block gas limit during the simulation in `getGasPriceOverrides` on [this line](https://github.com/sphinx-labs/sphinx/blob/main/packages/core/src/utils.ts#L184). Since Scroll's max batch size is 80% of the block gas limit, it was possible to underfund transactions by hardcoding the transaction's gas limit to be 75%. It was particularly easy to hit this bug with a max batch size of 97% because it significantly increases the chance that a gas limit of 75% will underfund the transaction.

I manually checked that this PR lets us propose Mean Finance's script on Scroll using the `develop` branch, even when `getMaxGasLimit` returns a value that's 97% of the block gas limit.


packages/core/src/utils.ts
```diff
@@ -168,20 +168,21 @@ export const getGasPriceOverrides = async (
...
+    // transactions. We only make this optimization in situations where we can safely assume that
+    // the caller has an unlimited amount of ETH. We don't override this on live networks because
+    // the signer is the user's wallet, which may have a limited amount of ETH.
+    //
+    // We set this value higher than the max batch gas limit to ensure that we don't accidentally
+    // underfund the transaction. If we set this value equal to the block gas limit, a situation
+    // could occur where the block gas limit decreases slightly after we set this value, which would
+    // cause an error due to the fact that the transaction's gas limit exceeds the block gas limit.
+    // This occurred when simulating a deployment on Polygon.
+    overridden.gasLimit = (maxGasLimit + block.gasLimit) / BigInt(2)
```
sam-goldman, line 185: I think this is preferable to hardcoding `overridden.gasLimit`to something like 95% because this strategy ensures that we won't run into this bug again, even if we set a max batch size that's _really_ close to the block gas limit.

packages/plugins/src/foundry/decode.ts
```diff
@@ -144,25 +143,12 @@ export const makeNetworkConfig = (
...
     const { root, nested } = accountAccesses[i]
     const gas = gasEstimates[i].toString()
 
-    if (BigInt(gas) > maxAllowedGasPerLeaf) {
-      const networkName = fetchNameForNetwork(BigInt(chainId))
-      throw new Error(
-        `Estimated gas for a transaction is too close to the block gas limit on ${networkName}.`
-      )
-    }
-
```
sam-goldman, line 165: This change is irrelevant to the PR, but I'm removing it anyways because this check is pointless and confusing. If a leaf is too large, the `findMaxBatchSize` function in the simulation will throw an error. Keeping this here will only confuse maintainers in the future. I'm removing it now because I noticed that you mentioned you were modifying this value when working on Rootstock

RPate97 (approved):
Looks good, can't believe I forgot about this.

packages/plugins/src/foundry/decode.ts
RPate97: I'm not like super opinionated on this, but the UX of this error definitely seems better than the error we throw in the simulation. So, I would lean towards keeping this. If we do, we should use the getMaxGasLimit function to fetch the limit instead of doing the calculation here.

sam-goldman: Wdym "the UX of this error"? Do you mean that it's better for us to throw this error before the simulation, or do you mean that this error message is better? If it's the latter, i'd rather just improve the error message in the simulation

sam-goldman: I'm good with keeping it. I improved the error message by using the `decodedAction`, which required me to put it further down in this function

RPate97 (approved):
Looks good

PR #1480, opened 2024-02-21
fix(core): Rootstock

RPate97:
## Purpose
- Properly calculates the gas cost in our heuristic using the call data cost of 68 per non-zero byte on Rootstock since they do not implement EIP-2028
- Standardizes the way we handle using the 80% block gas limit on Rootstock and Scroll and uses a boolean in the SPHINX_NETWORKS array to implement that
- Moves the buffer we add on the gas limit for action execution on Rootstock out of the `executeActionsViaManagedService` function and into the relayer service
- Adds support for Rootstock testnet.  I'm ok with supporting this and was able to find a functional node provider for it. I would prefer people to run into issues with the gas limit on a testnet rather than mainnet
- I also added a minor unrelated thing to this PR: A check that the user has a compatible NodeJS version installed. I've ran into this issue a few times (because I have Node 14 as my default globally). The current error the user encounters if they do not use at least `16.16.0` is functionally impossible to debug.

I've created [another ticket](https://linear.app/chugsplash/issue/CHU-580/support-max-size-limit-contracts-on-rootstock) to handle increasing the max batch size and/or reducing the buffers we add to our estimates to maximize the size of contract that is deployable on Rootstock using our system. I think we should come back to that if people run into issues using our system on Rootstock, or if Rootstock finishes with implementing EIP-2028.

### DIFF
--- a/packages/contracts/src/networks.ts
+++ b/packages/contracts/src/networks.ts
@@ -18,6 +18,8 @@ export type SupportedNetwork = {
   queryFilterBlockLimit: number
   legacyTx: boolean
   actionGasLimitBuffer: boolean
+  useHigherMaxGasLimit: boolean
+  eip2028: boolean
   rollupStack?: {
     provider: RollupProvider
     type: RollupType
@@ -30,6 +32,8 @@ export type SupportedLocalNetwork = {
   networkType: NetworkType
   legacyTx: false
   actionGasLimitBuffer: false
+  useHigherMaxGasLimit: false
+  eip2028: true
 }
 
 export const SPHINX_LOCAL_NETWORKS: Array<SupportedLocalNetwork> = [
@@ -39,6 +43,8 @@ export const SPHINX_LOCAL_NETWORKS: Array<SupportedLocalNetwork> = [
     networkType: 'Local',
     legacyTx: false,
     actionGasLimitBuffer: false,
+    useHigherMaxGasLimit: false,
+    eip2028: true,
   },
 ]
 
@@ -69,6 +75,8 @@ export const SPHINX_NETWORKS: Array<SupportedNetwork> = [
     queryFilterBlockLimit: 2000,
     legacyTx: false,
     actionGasLimitBuffer: false,
+    useHigherMaxGasLimit: false,
+    eip2028: true,
   },
   {
     name: 'sepolia',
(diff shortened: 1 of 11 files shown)


packages/contracts/src/networks.ts
```diff
@@ -760,6 +830,8 @@ export const SPHINX_NETWORKS: Array<SupportedNetwork> = [
     legacyTx: true,
     decimals: 18,
     actionGasLimitBuffer: false,
+    useHigherMaxGasLimit: true,
```
sam-goldman, line 833: Why is this `true` on this network?

packages/core/src/actions/execute.ts
```diff
@@ -682,32 +695,25 @@ export const executeActionsViaManagedService: ExecuteActions = async (
...
-    if (limit > maxGasLimit) {
-      limit = maxGasLimit
-    }
-
-    overrides.gasLimit = limit
+    minimumActionsGasLimit = estimateGasViaManagedService(
+      moduleAddress,
+      batch,
+      BigInt(deploymentContext.deployment.chainId)
+    )
```
sam-goldman, line 704: How is `minimumActionsGasLimit` used in the website?

packages/core/src/actions/execute.ts
```diff
@@ -839,13 +869,17 @@ export const estimateGasViaManagedService: EstimateGas = (
  * This buffer also makes our estimate closer to the value that would be returned by the
```
sam-goldman, line 869: Looks like number 2 in the JSDoc should be updated

sam-goldman (commented):
Looks good, just some questions and minor comments

sam-goldman:
If the user uses the Deploy CLI command on any network that doesn’t support EIP-2028 (e.g. Rootstock), would the transactions fail due to EthersJS underestimating them? If so, we can just address this issue later. Want to make a ticket or should I?

sam-goldman:
Have you checked that a live network deployment via the website works on Rootstock?

sam-goldman:
Just curious, does EthersJS automatically calculate the gas correctly when executing via the website on non-EIP 2028 networks? If it doesn’t, how do we fix that?

packages/contracts/src/networks.ts
RPate97: Good catch, it shouldn't be.

RPate97:
> Just curious, does EthersJS automatically calculate the gas correctly when executing via the website on non-EIP 2028 networks? If it doesn’t, how do we fix that?

It does not. See this comment:
https://github.com/sphinx-labs/sphinx/pull/1480/files#r1499987207

RPate97:
> Have you checked that a live network deployment via the website works on Rootstock?

yes

sam-goldman (approved):

PR #1482, opened 2024-02-21
fix(pg): Improve error message when config is not defined

RPate97:
## Purpose
Adds a better error message if none of the fields in the `sphinxConfig` object are defined.


packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -527,6 +527,18 @@ contract SphinxUtils is SphinxConstants, StdUtils {
      *         proposals, broadcasting, etc).
      */
     function validate(SphinxConfig memory _config) external pure {
+        if (
+            _config.owners.length == 0 &&
+            _config.threshold == 0 &&
+            bytes(_config.projectName).length == 0 &&
+            _config.mainnets.length == 0 &&
+            _config.testnets.length == 0
+        ) {
```
sam-goldman, line 536: Looks like this is missing a couple fields: `saltNonce` and `orgId`

packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -527,6 +527,18 @@ contract SphinxUtils is SphinxConstants, StdUtils {
...
+            revert(
+                "Sphinx: Detected an empty 'sphinxConfig' struct. Did you forget to add fields to it in your script's setUp function or constructor? If you've already added fields to it in your setUp function, have you called 'super.setUp()' in any contracts that inherit from your script?"
+            );
```
sam-goldman, line 539: I think we may want to use `\n` to avoid words getting wrapped weirdly in the user's terminal

packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -781,15 +793,6 @@ contract SphinxUtils is SphinxConstants, StdUtils {
...
     ) public pure returns (bytes memory safeInitializerData) {
-        require(
-            _owners.length > 0,
-            "Sphinx: You must have at least one owner in your 'sphinxConfig.owners' array."
-        );
-        require(
-            _threshold > 0,
-            "Sphinx: You must set your 'sphinxConfig.threshold' to a value greater than 0."
-        );
-
```
sam-goldman, line 792: I think we should keep these checks. Otherwise, this function will return initializer data that's bound to fail on-chain / yield an incorrect gnosis safe address. Kinda a footgun imo since this function is used in a few different contexts currently, and there may be more in the future

packages/contracts/test/Sphinx.t.sol
```diff
@@ -44,4 +44,16 @@ contract Sphinx_Test is Test, Sphinx, SphinxTestUtils {
...
+        sphinxConfig.owners = new address[](0);
+        sphinxConfig.threshold = 0;
+        sphinxConfig.testnets = new Network[](0);
+        sphinxConfig.mainnets = new Network[](0);
+        sphinxConfig.projectName = "";
+        vm.expectRevert(
+            "Sphinx: Detected an empty 'sphinxConfig' struct. Did you forget to add fields to it in your script's setUp function or constructor? If you've already added fields to it in your setUp function, have you called 'super.setUp()' in any contracts that inherit from your script?"
+        );
+        sphinxConfigABIEncoded();
+    }
```
sam-goldman, line 58: nit: The convention w/ our Forge test function names is generally in the format `test_<FUNCTION_NAME>_revert_<REASON>`

sam-goldman (approved):
Looks good, just a few minor comments

PR #1483, opened 2024-02-22
fix(ci): Stop demo tests from failing

RPate97:
## Purpose
Uses a limited batch size when compiling in the demo tests to prevent the process from timing out due to running out of memory. I found that this didn't slow down the demo test suite too much, so I kept it enabled for all PRs.


sam-goldman (approved):
A working test suite - what a concept!

PR #1485, opened 2024-02-23
feat(pg): Add timeout and retry logic to simulation

RPate97:
## Purpose
Adds timeout and retry logic to the simulation mainly for the purpose of making proposals more reliable.

Note that I chose not to implement retries in the deploy CLI command as part of this PR because that is actually a pretty separate task, would have caused this to take much longer, and generally wasn't the reason we decided to prioritize this. 

I've created a [new ticket](https://linear.app/chugsplash/issue/CHU-583/add-logic-to-relayer-service-to-handle-retrying-with-a-higher-gas) specifically to cover implementing retries in the deploy CLI task.

### DIFF
--- a/packages/core/src/actions/execute.ts
+++ b/packages/core/src/actions/execute.ts
@@ -13,7 +13,7 @@ import {
   decodeExecuteLeafData,
 } from '@sphinx-labs/contracts'
 import ora from 'ora'
-import { TransactionReceipt, ethers } from 'ethers'
+import { Contract, TransactionReceipt, ethers } from 'ethers'
 
 import { DeploymentConfig, ConfigArtifacts, NetworkConfig } from '../config'
 import {
@@ -1135,6 +1135,64 @@ export const sortSigners = (arr: Array<TreeSigner>): void => {
       return 0
     }
   })
+}
+
+/**
+ * This is a utility function that is helpful for recovering from a deployment that has timed out. It fetches
+ * all the transaction receipts related to a deployment using the expected events and returns them. If this
+ * function is used when some of the transaction receipts are already tracked, they can be passed in and this
+ * function will return an array of all of the unique transactions.
+ *
+ * This function modifies the passed in `receipts` array.
+ *
+ * This function returns standard ethers.TransactionReceipt objects instead of our SphinxTransactionReceipt
+ * type because that is more flexible and generally useful.
+ */
+export const fetchExecutionTransactionReceipts = async (
+  receipts: ethers.TransactionReceipt[],
+  moduleAddress: string,
+  merkleRoot: string,
+  provider: SphinxJsonRpcProvider | HardhatEthersProvider
+) => {
+  const module = new ethers.Contract(moduleAddress, SphinxModuleABI, provider)
+
+  const SphinxMerkleRootApprovedFilter =
+    module.filters.SphinxMerkleRootApproved(merkleRoot)
+  const SphinxMerkleRootCanceledFilter =
+    module.filters.SphinxMerkleRootCanceled(merkleRoot)
+  const SphinxMerkleRootFailedFilter =
+    module.filters.SphinxMerkleRootFailed(merkleRoot)
+  const SphinxActionSucceededFilter =
+    module.filters.SphinxActionSucceeded(merkleRoot)
+  const SphinxActionFailedFilter = module.filters.SphinxActionFailed(merkleRoot)
+
+  const filters = [
+    SphinxMerkleRootApprovedFilter,
+    SphinxMerkleRootCanceledFilter,
+    SphinxMerkleRootFailedFilter,
+    SphinxActionSucceededFilter,
+    SphinxActionFailedFilter,
+  ]
+
+  const txHashes = receipts.map((r) => r.hash)
+
+  const latestBlock = await provider.getBlockNumber()
+  for (const filter of filters) {
+    const startingBlock = latestBlock - 1999 > 0 ? latestBlock - 1999 : 0
+    const events = await module.queryFilter(filter, startingBlock, latestBlock)
+
+    for (const event of events) {
+      const receipt = await provider.getTransactionReceipt(
+        event.transactionHash
+      )
+      if (receipt && !txHashes.includes(receipt.hash)) {
+        txHashes.push(receipt.hash)
+        receipts.push(receipt)
+      }
+    }
+  }
+
+  return receipts
 }
 
 /**
--- a/packages/plugins/src/hardhat/simulate.ts
+++ b/packages/plugins/src/hardhat/simulate.ts
@@ -3,7 +3,6 @@ import { join } from 'path'
 import {
   spawnAsync,
   getSphinxWalletPrivateKey,
-  toSphinxLeafWithProofArray,
   SphinxTransactionReceipt,
   ExecutionMode,
   MerkleRootStatus,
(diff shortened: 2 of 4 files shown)


packages/plugins/src/hardhat/simulate.ts
```diff
@@ -275,10 +276,16 @@ export const simulateDeploymentSubtask = async (
...
+      executionCompleted = true
+      receipts = (
+        await fetchExecutionTransactionReceipts(
+          [],
+          deploymentContext.deployment.moduleAddress,
+          deploymentContext.deployment.deploymentConfig.merkleTree.root,
+          deploymentContext.provider
+        )
+      ).map(convertEthersTransactionReceipt)
     },
```
RPate97, line 289: Doing this here is probably a little overkill. Still, it keeps this function returning all the receipts in all situations, as is expected by the various places where this function is implemented. I think just adding this logic here is worth it for that purpose. This function already existed on the website, so I just moved it over into the mono repo.

packages/plugins/src/hardhat/simulate.ts
```diff
@@ -210,7 +210,6 @@ export const simulateDeploymentSubtask = async (
   hre: any
 ): Promise<{
   receipts: Array<SphinxTransactionReceipt>
-  batches: Array<Array<SphinxLeafWithProof>>
```
RPate97, line 213: We don't use the `batches` for anything, so I removed it to simplify this function.

packages/plugins/src/hardhat/simulate.ts
```diff
@@ -192,10 +192,14 @@ export const simulate = async (
...
-
-  return { receipts, batches }
+  try {
+    const receipts = JSON.parse(stdout).receipts
+    return { receipts }
+  } catch (e) {
+    console.log(stdout)
+    console.error(stderr)
+    process.exit(1)
+  }
```
RPate97, line 202: I've encountered a few situations where an error occurs that causes stdout not to be valid JSON. It will then fail to be parsed, and the process will crash. This is annoying because the real error that caused the problem will not be logged at all. So this just catches that case and logs `stdout` and `stderr` so we can access whatever the error is.

packages/core/src/actions/execute.ts
```diff
@@ -1137,6 +1137,62 @@ export const sortSigners = (arr: Array<TreeSigner>): void => {
...
+    SphinxMerkleRootCanceledFilter,
+    SphinxMerkleRootFailedFilter,
+    SphinxActionSucceededFilter,
+    SphinxActionFailedFilter,
+  ]
+
+  const txHashes = receipts.map((r) => r.hash)
+
+  for (const filter of filters) {
+    const latestBlock = await provider.getBlockNumber()
```
sam-goldman, line 1178: Seems like we should put `await provider.getBlockNumber` outside of the for-loop so that we only call it once instead of 5 times

packages/core/src/actions/execute.ts
```diff
@@ -1137,6 +1137,62 @@ export const sortSigners = (arr: Array<TreeSigner>): void => {
...
+    const startingBlock = latestBlock - 1999 > 0 ? latestBlock - 1999 : 0
+    const events = await module.queryFilter(filter, startingBlock, latestBlock)
+
+    for (const event of events) {
+      const receipt = await provider.getTransactionReceipt(
+        event.transactionHash
+      )
+      if (receipt && !txHashes.includes(receipt.hash)) {
+        txHashes.push(receipt.hash)
+        receipts.push(receipt as any)
```
sam-goldman, line 1188: seems like the `as any` is unnecessary

packages/core/src/actions/execute.ts
```diff
@@ -1137,6 +1137,62 @@ export const sortSigners = (arr: Array<TreeSigner>): void => {
...
 
+/**
+ * This is a utility function that is helpful for recovering from a deployment that has timed out. It fetches
+ * all the transaction receipts related to a deployment using the expected events and returns them. If this
+ * function is used when some of the transaction receipts are already tracked, they can be passed in and this
+ * function will return an array of all of the unique transactions.
+ *
+ * This function returns standard ethers.TransactionReceipt objects instead of our SphinxTransactionReceipt
+ * type because that is more flexible and generally useful.
+ */
```
sam-goldman, line 1148: nit: May be worth mentioning that this function modifies the given `receipts` array

packages/core/src/actions/execute.ts
```diff
@@ -1137,6 +1137,62 @@ export const sortSigners = (arr: Array<TreeSigner>): void => {
...
+    SphinxActionSucceededFilter,
+    SphinxActionFailedFilter,
+  ]
+
+  const txHashes = receipts.map((r) => r.hash)
+
+  for (const filter of filters) {
+    const latestBlock = await provider.getBlockNumber()
+    const startingBlock = latestBlock - 1999 > 0 ? latestBlock - 1999 : 0
+    const events = await module.queryFilter(filter, startingBlock, latestBlock)
```
sam-goldman, line 1180: Not sure if this is a concern or out of scope of this PR, but chains like Arbitrum can produce 2000 blocks in a matter of minutes. E.g. [this block](https://arbiscan.io/block/184139961) on Arbitrum occurs 2000 blocks before [this block](https://arbiscan.io/block/184141961), and they're 9 mins apart. Could this potentially prevent us from retrieving the receipts earlier in the deployment here?

packages/plugins/src/hardhat/simulate.ts
sam-goldman: Sounds good. nit: it'd be nice if this explanation is documented in the code

packages/plugins/src/hardhat/simulate.ts
```diff
@@ -303,25 +314,61 @@ export const simulateDeploymentSubtask = async (
...
-    throw new Error(
-      'Simulation failed for an unexpected reason. This is a bug. Please report it to the developers.'
-    )
-  }
+  let attempts = 0
+  while (executionCompleted === false) {
+    try {
+      const result = await callWithTimeout(
+        compileAndExecuteDeployment(simulationContext),
+        90000,
```
sam-goldman, line 323: is 90 seconds a safe value? i'm trying to think of the worst case scenario that could cause this call to prematurely timeout. maybe the worst case is that the user has a very large deployment on a single chain w/ a free tier (i.e. slow) provider. i'm not opinionated, but curious to hear your thoughts

sam-goldman (commented):
Looks good, just a couple minor comments / questions

packages/plugins/src/hardhat/simulate.ts
RPate97: I think you're looking at this incorrectly. We have to operate under the assumption that this call will inevitably time out in cases where it should not. For example, a rootstock deployment via the website will generally always timeout since the network is so slow. We would have to use a high timeout period to avoid that (like 10 minutes or more). Furthermore, even if we use a very high timeout period a user will inevitably come around with a deployment that is so large that the execution time exceeds the timeout period we set. We cannot prevent this from happening.

Since this function will inevitably timeout, we must design it so that it doesn't matter if it times out unnecessarily. The next attempt to execute the `compileAndExecuteDeployment` function will flawlessly recover no matter what. I've put a lot of work into ensuring this is the case. This strategy has proven to be very reliable on the website already. We've had lots of bugs but never related to this. 

It's worth noting that even if we could ensure that the function would never ever timeout, we would still need to design the `compileAndExecuteDeployment` to recover in all situations because our backend server will inevitably go down and some point, and we need to be able to recover from that. 

So the right way to look at the timeout is to think about what a reasonable time period is for timing out in cases where we legitimately want a timeout to occur (such as an rpc provider stalling, a transaction failing to get executed, etc). Then we should set a timeout period that is appropriate for how long we are willing to wait for that type of situation. 

Honestly, since this simulation is running against a local node that is quite fast (in comparison to a live network) it is very likely that we could reduce this time out further. However, I don't really think there's much value in that. I do not expect the timeout to actually occur very often during the simulation.

packages/core/src/actions/execute.ts
RPate97: [I created a ticket for it.](https://linear.app/chugsplash/issue/CHU-630/use-higher-block-query-block-limit-when-fetching-deployment)

PR #1493, opened 2024-02-24
Reverted

sam-goldman:


packages/contracts/contracts/core/SphinxGnosisSafeProxyFactory.sol
```diff
@@ -0,0 +1,162 @@
...
+        moduleProxyFactory = _moduleProxyFactory;
+        multiSend = _multiSend;
+    }
+
+    /**
+     * @param _owners The addresses that will own the deployed Gnosis Safe. These should be sorted
+     *                in ascending order off-chain so that the address of the Gnosis Safe can easily
+     *                be calculated by off-chain tooling.
+     */
+    function deployGnosisSafeWithModule(
```
sam-goldman, line 36: This function includes a lot of logic that was previously off-chain. Specifically, it contains all of the logic for constructing the Gnosis Safe's initializer data, which involves a MultiSend call. Although including this logic in the contract may not be strictly necessary, I included it for two reasons:
* I think this logic is important enough to include in the audit
* Putting this logic on-chain makes it easier for users to inspect the transaction that deploys their Gnosis Safe

packages/contracts/contracts/core/SphinxGnosisSafeProxyFactory.sol
```diff
@@ -0,0 +1,162 @@
...
+        // Get the address of the Sphinx module that will be deployed.
+        moduleProxy = moduleProxyFactory.computeSphinxModuleProxyAddress(
+            safeProxy,
+            safeProxy,
+            _saltNonce
+        );
+
+        // Encode the call to the Gnosis Safe's `setup` function, which is an input to the
+        // Gnosis Safe proxy's `CREATE2` salt.
+        bytes memory safeInitializerData = abi.encodeWithSelector(
```
sam-goldman, line 58: We use `abi.encodeWithSelector` throughout this function instead of `abi.encodeCall` because the latter doesn't exist in Solidity v0.8.4. The latter is preferable generally because it provides type safety.

packages/contracts/contracts/core/SphinxGnosisSafeProxyFactory.sol
```diff
@@ -0,0 +1,162 @@
...
+
+        // Check that the Sphinx Module proxy is deployed and that it's enabled in the Gnosis Safe
+        // proxy. We don't need to check that the Gnosis Safe proxy is deployed because we deployed
+        // it via `CREATE2` above, which must have succeeded if we were able to make it to this
+        // point.
+        require(moduleProxy.code.length > 0, "SphinxGnosisSafeProxyFactory: module not deployed");
+        require(
+            GnosisSafe(payable(safeProxy)).isModuleEnabled(moduleProxy),
+            "SphinxGnosisSafeProxyFactory: module not enabled"
+        );
```
sam-goldman, line 117: These are the lines that prevent the griefing attack. An alternative approach is to check for the existence of contracts such as the Gnosis Safe singleton and the MultiSend contract, but I found that approach to be brittle because there's a lot of complexity in the Gnosis Safe proxy's initialization process, which means there's more surface area to trigger the griefing attack. I'm much more confident in this approach.

packages/contracts/package.json
```diff
@@ -56,6 +55,7 @@
   },
   "dependencies": {
     "@gnosis.pm/safe-contracts": "https://github.com/safe-global/safe-contracts.git#767ef36bba88bdbc0c9fe3708a4290cabef4c376",
+    "@gnosis.pm/safe-contracts-1.4.1": "https://github.com/safe-global/safe-contracts.git#192c7dc67290940fcbc75165522bb86a37187069",
```
sam-goldman, line 58: Gnosis Safe v1.4.1 needs to be a production dependency because we import its `SafeProxy` in the `SphinxGnosisSafeProxyFactory` so that we can deploy the proxy

packages/contracts/contracts/core/SphinxGnosisSafeProxyFactory.sol
```diff
@@ -0,0 +1,166 @@
...
+        // it via `CREATE2` above, which must have succeeded if we were able to make it to this
+        // point.
+        require(
+            moduleProxy.code.length > 0,
+            "SphinxGnosisSafeProxyFactory: module proxy not deployed"
+        );
+        require(
+            GnosisSafe(payable(safeProxy)).isModuleEnabled(moduleProxy),
+            "SphinxGnosisSafeProxyFactory: module proxy not enabled"
+        );
```
sam-goldman, line 120: These are the lines that prevent the griefing attack. An alternative approach is to check for the existence of contracts such as the Gnosis Safe singleton and the MultiSend contract, but I found that approach to be brittle because there's a lot of complexity in the Gnosis Safe proxy's initialization process, which means there's more surface area to trigger the griefing attack. I'm much more confident in this approach.

packages/contracts/package.json
RPate97: I don't think either of these need to be production dependencies because we will not be compiling them in production, nor will we be importing any artifacts from them in production. Instead, we must once compile the `SphinxGnosisSafeProxyFactory` contract and store the compiler artifact in the `contracts` package. We'll then use that artifact whenever we need to deploy or interact with this contract.

PR #1494, opened 2024-02-25
fix(core): use `getBytesLength` to support bytecode that has library placeholders

sam-goldman:
Replaces all instances of `ethers.dataLength` with our own `getBytesLength` function. Copying the JSDoc of our `getBytesLength` function for context:

```
/**
 * Returns the length in bytes of the input hex string.
 *
 * The difference between this function and `ethers.dataLength` is that this function does not throw
 * an error if the input hex string contains library placeholders. This function will return the
 * correct length for hex strings with library placeholders because the placeholders are the same
 * length as library addresses. For more info on library placeholders, see:
 * https://docs.soliditylang.org/en/v0.8.23/using-the-compiler.html#library-linking
 *
 * @example
 * // returns 3
 * getBytesLength("0x123456")
 */
```

### DIFF
--- a/.changeset/funny-pans-do.md
+++ b/.changeset/funny-pans-do.md
@@ -1,0 +1,5 @@
+---
+'@sphinx-labs/core': patch
+---
+
+Use `getBytesLength` to support bytecode that has library placeholders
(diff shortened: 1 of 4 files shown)


RPate97 (approved):
Looks good

PR #1496, opened 2024-02-27
feat(pg): Collection deployment options

RPate97:
## Purpose
- Sets `FOUNDRY_SENDER` and `ETH_FROM` during collection phase.
- Sets the `--always-use-create-2-factory` cli flag during collection phase.


sam-goldman (approved):
Looks good, just a few minor comments

PR #1501, opened 2024-02-27
feat(pg): Allow configuration via configureSphinx() function

RPate97:
## Purpose
Allows configuration by implementing the `configureSphinx` function instead of setting the `sphinxConfig` state variable.


packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -35,6 +35,11 @@ import { IGnosisSafe } from "./interfaces/IGnosisSafe.sol";
 import { IMultiSend } from "./interfaces/IMultiSend.sol";
 import { IEnum } from "./interfaces/IEnum.sol";
 
+interface ISphinxScript {
+    function sphinxConfig() external view returns (SphinxConfig memory);
+    function configureSphinx() external pure returns (SphinxConfig memory);
+}
```
RPate97, line 41: Unfortunately, for this to be backward compatible, we must use a name other than `sphinxConfig` for the configuration function. I chose `configureSphinx`, but I'm open to changing it to something else if you have a better name.

PR #1502, opened 2024-02-28
fix(pg): Remove sphinx fork installation and instructions

RPate97:
## Purpose
Removes the Foundry fork and setup instructions


packages/plugins/src/foundry/utils/index.ts
```diff
@@ -1369,9 +1369,8 @@ export const assertValidVersions = async (
 
   if (forkInstalled === 'false') {
     throw new Error(
-      `Detected invalid Foundry version. Please use Sphinx's fork of Foundry by\n` +
-        `running the command:\n` +
-        `foundryup --repo sphinx-labs/foundry --branch sphinx-patch-v0.1.0`
+      `Detected invalid Foundry version. Please update to the latest version of Foundry by running:\n` +
+        `foundryup`
```
RPate97, line 1373: We still do this check since our changes were implemented in Foundry relatively recently so it's possible that the user may have not updated their installation in a while.

packages/plugins/test/mocha/foundry/utils.spec.ts
```diff
@@ -733,7 +733,8 @@ describe('Utils', async () => {
       const broadcast = await runForgeScript(
         'contracts/test/script/Libraries.s.sol',
         foundryToml.broadcastFolder,
-        getAnvilRpcUrl(chainId)
+        getAnvilRpcUrl(chainId),
+        'MyContractWithLibraries_Script'
       )
```
RPate97, line 738: After switching back to the official foundry release, I noticed this test was failing with the following error:
`Could not find target contract: contracts/test/script/Libraries.s.sol`

Since the changes in this PR are relatively minor, and because none of our other tests broke, this is likely a regression in Foundry. So I just fixed this by updating the `runForgeScript` function to accept a target contract function.

packages/plugins/src/cli/setup.ts
```diff
@@ -137,31 +137,23 @@ export const makeCLI = (
...
+            'Skips installing the Sphinx library and only installs the Sphinx Foundry fork.',
+          boolean: true,
+          default: false,
+        }),
       async (argv) => {
-        const { skipLibrary, ci } = argv
+        const { skipLibrary } = argv
 
-        if (skipLibrary && ci) {
+        if (skipLibrary) {
```
sam-goldman, line 149: Seems like the "skip library" flag isn't necessary? Also, if the user enters `--skip-library`, this if-statement will always throw an error

sam-goldman (commented):
Looks good, but it looks like you didn't update a few things

sam-goldman:
Looks like you need to remove the `--ci` flag in `ci-proposals.md`

sam-goldman:
Looks like you need to remove the Foundry fork troubleshooting section in `troubleshooting-guide.md`

packages/plugins/src/cli/setup.ts
```diff
@@ -137,31 +137,23 @@ export const makeCLI = (
       'install',
       'Installs the required version of the Sphinx Solidity library contracts and Sphinx Foundry fork',
       (y) =>
```
sam-goldman, line 139: This description is outdated

packages/plugins/src/foundry/utils/index.ts
sam-goldman: Can you update the docs in the `assertValidVersions` function to reflect this?

packages/plugins/src/cli/setup.ts
```diff
@@ -137,31 +137,23 @@ export const makeCLI = (
...
-          })
-          .option('ci', {
-            describe:
-              'Continuous integration mode (skips installing the Sphinx Foundry fork)',
-            boolean: true,
-            default: false,
-          }),
+        y.usage('Usage: sphinx install').option('ci', {
+          describe:
+            'Continuous integration mode (skips installing the Sphinx Foundry fork)',
```
sam-goldman, line 142: I think we should remove the `ci` flag description since it's deprecated. If you disagree, I think the description should at least be updated so that it doesn't mention the Foundry fork

packages/plugins/src/cli/setup.ts
sam-goldman: The description is still outdated

docs/ci-proposals.md
sam-goldman, line 1: A `--ci` flag is still used in this file

sam-goldman (approved):
Needs a few more updates, but otherwise lgtm

PR #1503, opened 2024-02-28
feat(pg): Use call depth when filtering AccountAccesses

RPate97:
## Purpose
- Use the `depth` field in the `AccountAccess` struct to determine if a transaction must be included in a deployment.
- Switches to using a forked version of `forge-std` that has been updated to include the `depth` field. We'll switch back to standard forge-std once the next release happens for it.


.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "packages/contracts/lib/forge-std"]
 	path = packages/contracts/lib/forge-std
-	url = https://github.com/foundry-rs/forge-std
+	url = https://github.com/sphinx-labs/forge-std
```
RPate97, line 3: I've tested that the switch to this fork works smoothly and [created a ticket to switch back once the next release for `forge-std` happens.](https://linear.app/chugsplash/issue/CHU-640/switch-back-to-official-forge-std-version)

packages/contracts/lib/forge-std
RPate97, line 1: There are more changes here than you might expect to be required. Technically, we only needed to add the `depth` field to the `AccountAccess` struct for our purpose. However, since this library is autogenerated, I think it's prudent to use their generation script to do the update rather than editing the file manually. That means that our fork includes all of the changes that would be included in the next release to forge-std if it were to happen now.

RPate97, line 1: There are more changes here than you might expect to be required. Technically, we only needed to add the `depth` field to the `AccountAccess` struct for our purpose. However, since this library is autogenerated, I think it's prudent to use their generation script to do the update rather than editing the file manually. That means that our fork includes all of the changes that would be included in the next release to forge-std if it were to happen now.

sam-goldman (approved):
Looks good!

sam-goldman:
Looks like CI is failing though

PR #1505, opened 2024-02-28
feat(pg): Allow configuration via configureSphinx() function

RPate97:
## Purpose
Allow configuration via a new `configureSphinx()` function

### DIFF
--- a/packages/contracts/contracts/foundry/SphinxUtils.sol
+++ b/packages/contracts/contracts/foundry/SphinxUtils.sol
@@ -35,6 +35,11 @@ import { IGnosisSafe } from "./interfaces/IGnosisSafe.sol";
 import { IMultiSend } from "./interfaces/IMultiSend.sol";
 import { IEnum } from "./interfaces/IEnum.sol";
 
+interface ISphinxScript {
+    function sphinxFetchConfig() external view returns (SphinxConfig memory);
+    function configureSphinx() external;
+}
+
 contract SphinxUtils is SphinxConstants, StdUtils {
     Vm private constant vm = Vm(address(uint160(uint256(keccak256("hevm cheat code")))));
 
@@ -521,12 +526,7 @@ contract SphinxUtils is SphinxConstants, StdUtils {
         return keccak256(abi.encodePacked(_key, _mappingSlotKey));
     }
 
-    /**
-     * @notice Performs validation on the user's deployment. This mainly checks that the user's
-     *         configuration is valid. This validation occurs regardless of the `SphinxMode` (e.g.
-     *         proposals, broadcasting, etc).
-     */
-    function validate(SphinxConfig memory _config) public pure {
+    function isConfigObjectEmpty(SphinxConfig memory _config) internal pure returns (bool) {
         if (
             _config.owners.length == 0 &&
             _config.threshold == 0 &&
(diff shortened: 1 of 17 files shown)


packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -536,8 +536,34 @@ contract SphinxUtils is SphinxConstants, StdUtils {
...
+     * @notice Performs validation on the user's deployment. This mainly checks that the user's
+     *         configuration is valid. This validation occurs regardless of the `SphinxMode` (e.g.
+     *         proposals, broadcasting, etc).
+     */
+    function validate(SphinxConfig memory _config) public {
+        // We still explicitly check if the config is empty b/c you could define the sphinxConfig
+        // function, but not actually configure any options in it.
+        if (isConfigObjectEmpty(_config)) {
             revert(
-                "Sphinx: Detected an empty 'sphinxConfig' struct. Did you forget to add fields to it in your script's\nsetUp function or constructor? If you've already added fields to it in your setUp function, have you\ncalled 'super.setUp()' in any contracts that inherit from your script?"
```
RPate97, line 540: `vm.expectRevert()` only works on the immediate next function call. If you do a call within that call before reverting, then `vm.expectRevert()` won't catch the revert, even if it is correct. So, to keep the `validate` function easy to test, I left it still accepting a `SphinxConfig` object and then created a new function, `fetchAndValidateConfig`, which fetches, validates, and returns the config object. So we'll usually want to use that function. It's not required, though; it's just generally more convenient, IMO.

packages/plugins/src/sample-project/sample-contracts.ts
```diff
@@ -55,10 +55,7 @@ contract HelloSphinxScript is Sphinx {
     }
 
     function run() public sphinx {
-        // Set the \`CREATE2\` salt to be the hash of the owner(s). Prevents
-        // address collisions.
-        bytes32 salt = keccak256(abi.encode(sphinxConfig.owners));
-        helloSphinx = new HelloSphinx{ salt: salt }("Hi", 2);
+        helloSphinx = new HelloSphinx("Hi", 2);
         helloSphinx.add(8);
     }
```
RPate97, line 60: Just simplified this a bit.

packages/contracts/test/ScriptConfiguration.t.sol
```diff
@@ -0,0 +1,40 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+// Technically, we could use a forge-std remapping in our non-production contracts (scripts, tests, etc). We chose not too because allowing
+// that would increase the chance that we accidentally ship something that may not compile in the users project because we accidentially
+// relied on a remapping in the wrong file.
```
sam-goldman, line 6: > We chose not too because allowing that would increase the chance that we accidentally ship something that may not compile in the users project because we accidentially relied on a remapping in the wrong file.

I don't understand this sentence. Would you mind clarifying?

sam-goldman (approved):
Looks good, just some minor comments/questions

packages/contracts/test/ScriptConfiguration.t.sol
RPate97: This is copied from the same comment in `Sphinx.t.sol`, so it's unrelated to this PR. I just kept it b/c it's helpful to have this documented in multiple places. 

I do think it's worth adding a sentence to this comment in both places for a bit more clarity. It should be something like this:

We do not use remappings for forge-std because if we did, the user would need to define them in their remappings, or our library contracts would not work. Technically, we could use a forge-std remapping in our non-production contracts (scripts, tests, etc.), such as this file, without forcing the user to define a remapping. We chose not to because doing that would increase the chance of accidentally shipping something that may not compile in the user's project because we accidentally relied on a remapping in the wrong file.

PR #1506, opened 2024-02-28
fix(core): keep previous deployment artifacts

sam-goldman:
The changes to the artifact generation logic are simple; the vast majority of this PR is just adding dummy/fake variables for tests, and also updating the existing artifact testing logic to reflect the updated interface to the `makeDeploymentArtifacts` function

### DIFF
--- a/packages/core/src/artifacts.ts
+++ b/packages/core/src/artifacts.ts
@@ -22,6 +22,7 @@ import {
   DeploymentConfig,
   ConfigArtifacts,
   NetworkConfig,
+  BuildInfos,
 } from './config/types'
 import {
   fetchNetworkConfigFromDeploymentConfig,
@@ -32,7 +33,7 @@ import {
 } from './utils'
 import { ExecutionMode } from './constants'
 
-type NetworkArtifacts = {
+export type NetworkArtifacts = {
   executionArtifacts: {
     [txArtifactFileName: string]: ExecutionArtifact
   }
@@ -268,23 +269,24 @@ export const convertEthersTransactionReceipt = (
 }
 
 /**
- * Makes contract deployment artifacts on a single network for a single deployment.
+ * Makes contract deployment artifacts on a single network for a single deployment. Mutates the
+ * input `artifacts` object.
  *
- * @param previousArtifacts An object containing all previous contract deployment artifacts on the
- * network for the project.
+ * @param artifacts An object containing all previous contract deployment artifacts on the network
+ * for the project.
  */
-const makeContractDeploymentArtifacts = async (
+export const makeContractDeploymentArtifacts = async (
   merkleRoot: string,
   networkConfig: NetworkConfig,
-  deploymentConfig: DeploymentConfig,
+  buildInfos: BuildInfos,
   receipts: Array<SphinxTransactionReceipt>,
   configArtifacts: ConfigArtifacts,
-  previousArtifacts: {
+  artifacts: {
     [fileName: string]: ContractDeploymentArtifact | undefined
   },
   provider: SphinxJsonRpcProvider
-): Promise<{ [fileName: string]: ContractDeploymentArtifact }> => {
-  const isSuffixed = Object.keys(previousArtifacts).every((fileName) =>
+): Promise<void> => {
+  const isSuffixed = Object.keys(artifacts).every((fileName) =>
     fileName.endsWith('.json')
   )
   if (!isSuffixed) {
@@ -297,14 +299,13 @@ const makeContractDeploymentArtifacts = async (
   const moduleInterface = new ethers.Interface(SphinxModuleABI)
 
   const { gitCommit, chainId } = networkConfig
-  const artifacts: { [fileName: string]: ContractDeploymentArtifact } = {}
   const numDeployments: { [fileName: string]: number | undefined } = {}
   for (const action of networkConfig.actionInputs) {
     for (const contract of action.contracts) {
       const { fullyQualifiedName, initCodeWithArgs, address } = contract
       const { artifact: compilerArtifact, buildInfoId } =
         configArtifacts[fullyQualifiedName]
-      const buildInfo = deploymentConfig.buildInfos[buildInfoId]
+      const buildInfo = buildInfos[buildInfoId]
 
       if (!compilerArtifact || !buildInfo) {
         throw new Error(`Could not find artifact for: ${fullyQualifiedName}`)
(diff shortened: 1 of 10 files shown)


packages/core/src/artifacts.ts
```diff
@@ -268,23 +269,24 @@ export const convertEthersTransactionReceipt = (
 }
 
 /**
- * Makes contract deployment artifacts on a single network for a single deployment.
+ * Makes contract deployment artifacts on a single network for a single deployment. Mutates the
+ * input `artifacts` object.
```
sam-goldman, line 273: The easiest way to implement this patch was to mutate the previous artifacts (i.e. `artifacts`)

packages/core/src/artifacts.ts
```diff
@@ -268,23 +269,24 @@ export const convertEthersTransactionReceipt = (
...
   receipts: Array<SphinxTransactionReceipt>,
   configArtifacts: ConfigArtifacts,
-  previousArtifacts: {
+  artifacts: {
     [fileName: string]: ContractDeploymentArtifact | undefined
   },
   provider: SphinxJsonRpcProvider
-): Promise<{ [fileName: string]: ContractDeploymentArtifact }> => {
-  const isSuffixed = Object.keys(previousArtifacts).every((fileName) =>
+): Promise<void> => {
```
sam-goldman, line 288: We don't return anything from this function to hopefully make it clearer that it mutates the input `artifacts`

packages/core/src/artifacts.ts
```diff
@@ -528,12 +527,11 @@ export const isContractDeploymentArtifact = (
 }
 
 /**
- * Make deployment artifacts for the most recent deployment in a project.
+ * Make deployment artifacts for the most recent deployment in a project. Mutates the input
+ * `artifacts` object.
```
sam-goldman, line 531: My comments for the `makeContractDeploymentArtifacts` function applies to this function too

packages/core/src/artifacts.ts
```diff
@@ -543,23 +541,25 @@ export const makeDeploymentArtifacts = async (
...
-): Promise<DeploymentArtifacts> => {
-  const allNetworkArtifacts: DeploymentArtifacts['networks'] = {}
-  const compilerInputArtifacts: DeploymentArtifacts['compilerInputs'] = {}
+  configArtifacts: ConfigArtifacts,
+  artifacts: DeploymentArtifacts
+): Promise<void> => {
+  const {
+    networks: allNetworkArtifacts,
+    compilerInputs: compilerInputArtifacts,
+  } = artifacts
```
sam-goldman, line 553: We mutate the `allNetworkArtifacts` and `compilerInputArtifacts` variables

packages/plugins/test/mocha/artifacts.spec.ts
```diff
@@ -125,15 +300,13 @@ const makeThenRunThenCheckDeployment = async (
     merkleRootNonce: number
     executionMode: ExecutionMode
     accountAccesses: Array<ParsedAccountAccess>
-    expectedNumExecutionArtifacts: number
```
sam-goldman, line 128: The rest of the diff in this file just updates the helper functions for the existing artifact tests

packages/plugins/test/mocha/cli/deploy.spec.ts
sam-goldman, line 1: The changes in this file just update the aspects of the deployment tests to reflect the latest artifact helper function logic

packages/plugins/test/mocha/common.ts
sam-goldman, line 1: The updates to this file also just update the existing artifact helper functions. Particularly, it updates `expectedNumExecutionArtifacts`, which is different now that we're keeping previous deployment artifacts

packages/plugins/test/mocha/mock.ts
```diff
@@ -119,58 +97,6 @@ export const makeMockSphinxContext = (
   }
 }
 
-const makeMockNetworkConfig = (): NetworkConfig => {
```
sam-goldman, line 122: I think this is technically a dummy variable, so I moved it to `dummy.ts`

packages/plugins/test/mocha/cli/deploy.spec.ts
sam-goldman, line 1: The changes in this file just update the aspects of the deployment tests to reflect the latest artifact helper function logic

packages/plugins/test/mocha/common.ts
sam-goldman, line 1: The updates to this file also just update the existing artifact helper functions. Particularly, it updates `expectedNumExecutionArtifacts`, which is different now that we're keeping previous deployment artifacts

packages/plugins/src/cli/deploy.ts
```diff
@@ -423,17 +423,32 @@ export const deploy = async (
...
+  // Create a deployment artifact object that contains the previous contract deployment artifacts,
+  // which will potentially be modified when we make the deployment artifacts. Specifically, the
+  // `history` field of the contract deployment artifacts could be modified. We don't include the
+  // existing execution artifacts or the compiler inputs in this object because we don't need to
+  // modify them.
+  const deploymentArtifacts = {
+    networks: {
+      [chainId.toString()]: {
+        contractDeploymentArtifacts: previousContractArtifacts,
+        executionArtifacts: {},
```
RPate97, line 435: Nit: Not including the `executionArtifacts` here could potentially be a footgun if we later add logic that does modify them and forget that we didn't pass them in here. I would probably include them here unless it requires a huge amount of work to do.

packages/plugins/test/mocha/artifacts.spec.ts
```diff
@@ -93,30 +140,156 @@ describe('Artifacts', () => {
...
+        `${dummyContractName}.json`
+      ]
+    const newContractArtifact =
+      artifacts.networks[newChainId].contractDeploymentArtifacts[
+        `${newContractName}.json`
+      ]
+    expect(isContractDeploymentArtifact(previousContractArtifact)).equals(true)
+    expect(previousContractArtifact.chainId).equals(dummyChainId)
+    expect(isContractDeploymentArtifact(newContractArtifact)).equals(true)
+    expect(newContractArtifact.chainId).equals(newChainId)
```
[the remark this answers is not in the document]
sam-goldman: Agree that it would be better, but I'm gonna keep this unchanged because this test case checks for different things than the `checkArtifacts` function, and it'd be nontrivial to update the `checkArtifacts` function. Specifically, the `checkArtifacts` function assumes that the networks and contracts remain the same for all deployments, whereas this test case changes the network and the contract in each deployment.

packages/plugins/src/cli/deploy.ts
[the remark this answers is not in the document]
sam-goldman: Yeah agreed. Updated this and manually checked that it works

PR #1509, opened 2024-02-29
fix(core): Continue with retries on rate limiting

RPate97:
## Purpose
- Removes the logic to stop retries from happening if an error is due to rate limiting
- Adds logic to reset the simulation fork if a rate limit happens. See the inline comment for more information on this.

### DIFF
--- a/packages/plugins/src/hardhat/simulate.ts
+++ b/packages/plugins/src/hardhat/simulate.ts
@@ -208,6 +208,36 @@ export const simulate = async (
 }
 
 /**
+ * Handles setting up any
+ *
+ * @param provider
+ * @returns
+ */
+export const setupPresimulationState = async (
+  provider: any,
+  executionMode: ExecutionMode
+) => {
+  let signer: ethers.Wallet
+  if (executionMode === ExecutionMode.LiveNetworkCLI) {
+    const privateKey = process.env.PRIVATE_KEY
+    if (!privateKey) {
+      throw new Error(`Could not find 'PRIVATE_KEY' environment variable.`)
+    }
+    signer = new ethers.Wallet(privateKey, provider)
+  } else if (
+    executionMode === ExecutionMode.LocalNetworkCLI ||
+    executionMode === ExecutionMode.Platform
+  ) {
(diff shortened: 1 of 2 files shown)


packages/plugins/src/hardhat/simulate.ts
```diff
@@ -350,17 +350,54 @@ export const simulateDeploymentSubtask = async (
...
       }
 
+      await provider.send('hardhat_reset', [
+        {
+          forking: {
+            jsonRpcUrl: hre.config.networks.hardhat.forking?.url,
+            blockNumber: hre.config.networks.hardhat.forking?.blockNumber,
+          },
+        },
+      ])
```
RPate97, line 390: This is actually amazingly reliable. It completely solved my method of replicating the rate limit issue, which was to deploy on Avalanche, Celo, and Linea Goerli while using the infura provider. We'll have to see if we find other configurations that continue to cause problems, though.

packages/plugins/src/hardhat/simulate.ts
```diff
@@ -350,17 +350,54 @@ export const simulateDeploymentSubtask = async (
...
+
+      // Since we're resetting back to the initial state, we also have to fund the simulation sender wallet
+      if (
+        executionMode === ExecutionMode.LocalNetworkCLI ||
+        executionMode === ExecutionMode.Platform
+      ) {
+        signer = new ethers.Wallet(getSphinxWalletPrivateKey(0), provider)
+        await fundAccountMaxBalance(signer.address, provider)
+      }
```
sam-goldman, line 399: Makes sense, but I'm probably going to forget to update this code if we add more pre-simulation steps in the future. Would be less of a maintenance hazard if this code and the [corresponding code](https://github.com/sphinx-labs/sphinx/blob/265060df7c542cb4d02973e9fa77c4778e57f16e/packages/plugins/src/hardhat/simulate.ts#L247-L248) in `simulateDeploymentSubtask` are refactored into a single function

sam-goldman (approved):
Looks good, just a couple minor comments

PR #1510, opened 2024-03-01
docs: remove networks in sphinx config in favor of rpc_endpoints

sam-goldman:
Updates the docs to include arbitrary network names in proposals. Also includes configuration instructions for the `propose --networks [mainnets|testnets]` option. The next couple PRs will implement the new network UX.

### DIFF
--- a/docs/cli-existing-project.md
+++ b/docs/cli-existing-project.md
@@ -1,6 +1,6 @@
 # Getting Started with an Existing Foundry Project
 
-In this guide, you'll integrate Sphinx with your existing Foundry project. Then, you'll deploy your project on a few testnets.
+In this guide, you'll integrate Sphinx with your existing Foundry project. Then, you'll deploy your project on test networks.
 
 Deployments are a three-step process with the DevOps Platform:
 
@@ -25,11 +25,10 @@ In this guide, you'll propose the deployment on the command line and then approv
   e. [Handle new sender address](#e-handle-new-sender-address)\
   f. [Add configuration options](#e-add-configuration-options)
 7. [Add environment variables](#7-add-environment-variables)
-8. [Update RPC endpoints](#8-update-rpc-endpoints)
-9. [Update `foundry.toml` settings](#9-update-foundrytoml-settings)
-10. [Run tests](#10-run-tests)
-11. [Propose on testnets](#11-propose-on-testnets)
-12. [Next steps](#12-next-steps)
+8. [Update `foundry.toml` settings](#8-update-foundrytoml-settings)
+9. [Run tests](#9-run-tests)
+10. [Propose on testnets](#10-propose-on-testnets)
+11. [Next steps](#11-next-steps)
 
 ## 1. Prerequisites
 
@@ -158,11 +157,6 @@ Copy and paste the following `configureSphinx()` function template into your scr
 function configureSphinx() public override {
   sphinxConfig.owners = [<your address>];
   sphinxConfig.orgId = <Sphinx org ID>;
-  sphinxConfig.testnets = [
-    Network.sepolia,
-    Network.optimism_sepolia,
-    Network.arbitrum_sepolia
-  ];
   sphinxConfig.projectName = "My_First_Project";
   sphinxConfig.threshold = 1;
 }
@@ -171,7 +165,6 @@ function configureSphinx() public override {
 You'll need to update the following fields in this template:
 * Enter your address in the `owners` array.
 * Enter your Sphinx Organization ID in the `orgId` field. It's a public field, so you don't need to keep it secret. You can find it in the Sphinx UI.
-* If you'd like to deploy on networks other than Sepolia, Optimism Sepolia, and Arbitrum Sepolia, update the `testnets` array. You can find a list of valid fields in the [Sphinx Configuration Options reference](https://github.com/sphinx-labs/sphinx/blob/main/docs/configuration-options.md#network-testnets).
 
 ## 7. Add environment variables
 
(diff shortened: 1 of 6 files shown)


docs/cli-propose.md
```diff
@@ -21,37 +23,73 @@ The following steps occur when this command is run:
...
 
+## Advanced Usage
+
+### Proposing on all Test Networks or Production Networks
+
+To clearly differentiate between test networks and production networks in your `foundry.toml`, you must label the networks in its `rpc_endpoints` section. This involves inserting specific comment lines before and after the networks you want to categorize.
+
+Here is how you can label your networks:
+
+* **Test Networks**: Enclose your test network entries with `# sphinx-testnets-start` and `# sphinx-testnets-end`.
```
sam-goldman, line 56: Copied this syntax from slither

sam-goldman:
This PR is an opportune time to change the name "mainnets" to something else, like "prod" or "production". Any preference @RPate97?

docs/cli-propose.md
```diff
@@ -21,37 +23,73 @@ The following steps occur when this command is run:
...
 - `--confirm`: **Optional**. Confirm the proposal without previewing it. Useful for automating proposals.
 - `--dry-run`: **Optional**. Perform a trial run without sending data to Sphinx's backend. Useful for testing and validation.
 - `--silent`: **Optional**. Suppress output to display only error messages. Must be combined with `--confirm` to confirm the proposal without previewing it.
 - `--target-contract <TARGET_CONTRACT>` (Alias: `--tc <TARGET_CONTRACT>`): **Optional**. Specify a contract in multi-contract scripts.
 
+## Advanced Usage
+
+### Proposing on all Test Networks or Production Networks
+
+To clearly differentiate between test networks and production networks in your `foundry.toml`, you must label the networks in its `rpc_endpoints` section. This involves inserting specific comment lines before and after the networks you want to categorize.
```
RPate97, line 52: You jump straight into explaining why the comment is necessary before actually covering the feature itself. I found this confusing even having read the info above on the `testnets` and `mainnets` keywords. 

We should add a short sentence here just giving brief overview of the feature. Something like this would be good:
> To more easily propose on a large number of networks simultaneously, you can use the `mainnets` or `testnets` keywords. These keywords are aliases for all of the networks you've labeled as `testnets` or `mainnets` in your `foundry.toml` file.

RPate97 (approved):
Looks pretty good, I like the idea of doing the docs first. It's almost like making a spec. 

My main opinion on this is that if we're going to have the `testnets`/`mainnets` keywords, we should make an effort to make sure people are aware of the feature. If we don't then there's not much point in supporting it at all. That said, we don't want to distract the user during the initial integration. 

Though I will say, generally, I think making sure people are aware of all of the functionality we provide outside of just what's mentioned in the getting started guides is an ongoing issue. So, we should spend more time improving that outside of this specific issue.

sam-goldman:
> This PR is an opportune time to change the name "mainnets" to something else, like "prod" or "production". Any preference @RPate97?

@RPate97

RPate97:
> > This PR is an opportune time to change the name "mainnets" to something else, like "prod" or "production". Any preference @RPate97?
> 
> @RPate97

Did you have something specific in mind? I don't see a strong reason to change it.

docs/cli-propose.md
sam-goldman: Yeah that's much better

docs/cli-existing-project.md
```diff
@@ -237,6 +217,6 @@ Here are the steps that occur when you run this command:
 
 When the proposal is finished, go to the [Sphinx UI](https://sphinx.dev) to approve the deployment. After you approve it, you can monitor the deployment's status in the UI while it's executed.
 
-## 12. Next steps
+## 11. Next steps
 
 Before you use Sphinx in production, we recommend reading the [Writing Deployment Scripts with Sphinx guide](https://github.com/sphinx-labs/sphinx/blob/main/docs/writing-scripts.md), which covers essential information for using Sphinx.
```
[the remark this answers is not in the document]
sam-goldman: I'm concerned that adding more links to the "Next Steps" section will reduce the likelihood that users will read the "Writing Scripts with Sphinx" guide, which is the most important guide IMO

I agree generally that we need to do a better job of making our guides visible to the user, but I think the proper solution is to migrate to something like Docusaurus so we can at least have a sidebar. As long as we make the proposal reference guide visible to the user somehow (like in a sidebar), I think it's completely reasonable that a user will eventually get tired of manually typing all of their network names, then think, "maybe Sphinx has a shorthand for this", then go to the proposal reference guide, where this shorthand is thoroughly documented

I'm gonna keep this section unchanged, but lmk if you disagree

docs/configuration-options.md
```diff
@@ -12,110 +12,65 @@ function configureSphinx() public override {
...
-| Moonbeam | `Network.moonbeam` |
-| Moonriver | `Network.moonriver` |
-| Fuse | `Network.fuse` |
-| Evmos | `Network.evmos` |
-| Kava | `Network.kava` |
-| OKT Chain | `Network.oktc` |
-| Rootstock | `Network.rootstock` |
-
-### `Network[] testnets`
+A list of production networks to propose on via the `--networks mainnets` flag. Provides a convenient way to propose on many networks without needing to specify them on the command line. Note that the strings in this array must match the network names in the `rpc_endpoints` section of your `foundry.toml` file.
```
RPate97, line 68: It’s probably worth noting somewhere that when using the propose command, you are limited to the networks we support and then maybe just link to the table in the main readme.

RPate97 (approved):
Looks good

PR #1512, opened 2024-03-02
fix(pg): Handle higher storage cost on Moonbeam

RPate97:
## Purpose
- Handle [Moonbeams' irregular method of calculating gas](https://docs.moonbeam.network/builders/get-started/eth-compare/tx-fees/) per byte of storage used. 
- Enables the `useHigherMaxGasLimit` option on Moonbeam and related networks.

## Notes
Since Moonbeam has a rather low block gas limit (15 million), and the increased storage costs make it more expensive to deploy contracts, I realized we probably don't support max size limit contracts on Moonbeam or its related networks. I tested this by trying to deploy the DCAHub contract, which I couldn't deploy even with the `useHigherMaxGasLimit` option enabled. It is quite close, though, so we should be able to make it work pretty easily. 

I went ahead and enabled the `useHigherMaxGasLimit` option in this PR. I've created a [separate ticket](https://linear.app/chugsplash/issue/CHU-646/use-ethers-caching-logic-on-website) to ensure we have support for maximum-size-limit contracts on the network.

### DIFF
--- a/packages/contracts/contracts/foundry/SphinxPluginTypes.sol
+++ b/packages/contracts/contracts/foundry/SphinxPluginTypes.sol
@@ -75,6 +75,11 @@ struct ParsedAccountAccess {
     Vm.AccountAccess[] nested;
 }
 
+struct DeployedContractSize {
+    address account;
+    uint size;
+}
+
 /**
  * @notice Contains all of the information that's collected in a deployment on a single chain.
  *         The only difference between this struct and the TypeScript `DeploymentInfo` object is
@@ -107,6 +112,7 @@ struct FoundryDeploymentInfo {
     string sphinxLibraryVersion;
     bytes[] encodedAccountAccesses;
     uint256[] gasEstimates;
+    bytes encodedDeployedContractSizes;
 }
 
 enum ExecutionMode {
@@ -250,5 +256,11 @@ contract SphinxPluginTypes {
         external
         view
         returns (SphinxLeafWithProof[][] memory batches)
+    {}
+
+    function deployedContractSizesType()
+        external
+        view
+        returns (DeployedContractSize[] memory deployedContractSizes)
     {}
 }
--- a/packages/contracts/src/networks.ts
+++ b/packages/contracts/src/networks.ts
@@ -1,3 +1,121 @@
+import {
+  AccountAccess,
+  AccountAccessKind,
+  DeployedContractSize,
+  ParsedAccountAccess,
+} from './types'
+
+/**
+ * Calculates the storage used by a single AccountAccess. This function is intentionally very simple.
+ * We naively assume that every write takes up a full 32 bytes of storage despite the fact that there
+ * are many cases where the storage usage is less or even negative. We do this because we prefer to
+ * always overestimate the cost by a reasonable amount. Since the block gas limit on Moonbeam and
+ * related networks is comfortably high (15 million), this does not impede the users ability to deploy
+ * large contracts.
+ */
+const calculateStorageSizeForAccountAccess = (
+  access: AccountAccess,
+  deployedContractSizes: DeployedContractSize[]
+): { contractStorageSize: number; writeStorageSize: number } => {
+  const storageWriteSize =
+    access.storageAccesses.filter((storageAccess) => storageAccess.isWrite)
+      .length * 32
+
+  if (access.kind === AccountAccessKind.Create) {
+    const deployedContractSize = deployedContractSizes.find(
+      (deployedContact) => deployedContact.account === access.account
+    )
+
+    if (!deployedContractSize) {
+      throw new Error(
+        'Failed to find deployed contract size. This is a bug, please report it to the developers.'
+      )
+    }
+
+    return {
+      writeStorageSize: storageWriteSize,
+      contractStorageSize: Number(deployedContractSize.size),
+    }
+  } else {
+    return {
+      writeStorageSize: storageWriteSize,
+      contractStorageSize: 0,
+    }
+  }
+}
+
+/**
+ * Calculates the cost of a transaction on Moonbeam using their higher gas cost per byte.
+ *
+ * @param baseGas The estimated gas cost according to Foundry.
+ * @param deployedContractSizes The sizes of any contracts deployed during this transaction.
+ * @param access The ParsedAccountAccess for the transaction.
+ * @returns
+ */
+export const calculateActionLeafGasForMoonbeam = (
+  foundryGas: string,
+  deployedContractSizes: DeployedContractSize[],
+  access: ParsedAccountAccess
+): string => {
+  // Fetch the storage used by the root account access
+  const { contractStorageSize, writeStorageSize } =
+    calculateStorageSizeForAccountAccess(access.root, deployedContractSizes)
+
+  // Fetch the storage used by all the nested accesses
+  const nestedStorageSizes = access.nested.map((nestedAccountAccess) =>
+    calculateStorageSizeForAccountAccess(
+      nestedAccountAccess,
+      deployedContractSizes
+    )
+  )
+  const nestedContractStorageSize = nestedStorageSizes
+    .map((storageSize) => storageSize.contractStorageSize)
+    .reduce((prev, curr) => prev + curr, 0)
+  const nestedWriteStorageSize = nestedStorageSizes
+    .map((storageSize) => storageSize.writeStorageSize)
+    .reduce((prev, curr) => prev + curr, 0)
+
+  // Calculate the total storage for the full transaction
+  const totalContractSize = contractStorageSize + nestedContractStorageSize
+  const totalWriteStorageSize = writeStorageSize + nestedWriteStorageSize
+
+  // Gas per byte ratio = Block Gas Limit / (Block Storage Limit (kb) * 1024 Bytes)
+  const ratio = 15_000_000 / (40 * 1024)
+
+  // Total gas cost for storage on moonbeam
+  // The ratio isn't an exact integer, so we round the result up
+  const moonbeamStorageCost = Math.ceil(
+    (totalContractSize + totalWriteStorageSize) * ratio
+  )
+
+  /**
+   * Calculate the cost of the transaction according to Foundry without gas costs related to
+   * storage of the contracts.
+   *
+   * If we did not subtract the 200 gas / byte here, then we would wildly overestimate the cost of
+   * deploying contracts which could prevent the user from being able to deploy even reasonably
+   * sized contracts on Moonbeam due to their lower block gas limit.
+   *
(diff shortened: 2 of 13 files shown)


packages/contracts/src/networks.ts
```diff
@@ -1,3 +1,90 @@
...
+
+/**
+ * Calculates the cost of a transaction on Moonbeam using their higher gas cost per byte.
+ *
+ * @param baseGas The estimated gas cost according to Foundry.
+ * @param deployedContractSizes The sizes of any contracts deployed during this transaction.
+ * @param access The ParsedAccountAccess for the transaction.
+ * @returns
+ */
+const calculateActionLeafGasForMoonbeam = (
```
RPate97, line 49: Optimizing this would maybe help a bit with supporting max size limit contracts on Moonbeam. However, I don't really think that's worth doing. I would just bring our buffers down a little instead.

packages/contracts/src/networks.ts
```diff
@@ -1,3 +1,90 @@
...
+        deployedContractSizes
+      )
+    )
+    .reduce((prev, curr) => prev + curr)
+
+  // Calculate the total storage for the full transaction
+  const totalStorageSize = rootStorageSize + nestedStorageSize
+
+  // Gas per byte ratio = Block Gas Limit / (Block Storage Limit (kb) * 1024 Bytes)
+  const ratio = 15_000_000 / (40 * 1024)
```
sam-goldman, line 74: Just pointing out that this ratio can change unpredictably if moonbeam changes any of these parameters

packages/contracts/src/networks.ts
```diff
@@ -1,3 +1,90 @@
...
+
+  // Total gas cost for storage on moonbeam
+  // The ratio isn't an exact integer, so we round the result up
+  const moonbeamStorageCost = Math.ceil(totalStorageSize * ratio)
+
+  // Calculate the cost of the transaction according to Foundry without any storage usage
+  const baseGasLessStorageCost = Number(foundryGas) - 200 * totalStorageSize
```
sam-goldman, line 81: I get the idea, but where did 200 come from? For context, here's a [website](https://www.evm.codes/?fork=shanghai) that breaks down the cost of EVM opcodes, including SSTORE

sam-goldman, line 81: Is it possible for baseGasLessStorageCost to be less than 0 if we radically overestimate the amount of storage written?

sam-goldman, line 81: Are you sure that Moonbeam's storage costs aren't _in addition_ to the EVM's? I didn't see a clear answer when I read their "Calculating Transactions" page. However, the language in the document seems to imply that the storage costs are added onto the existing EVM gas cost:
> Then, you can take the storage growth in bytes for a given transaction and multiply it by the gas-to-storage growth ratio to determine how many units of gas to add to the transaction.

packages/contracts/src/networks.ts
```diff
@@ -24,6 +111,11 @@ export type SupportedNetwork = {
     provider: RollupProvider
     type: RollupType
   }
+  handleNetworkSpecificMerkleLeafGas?: (
+    foundryGas: string,
+    deployedContractSizes: DeployedContractSize[],
+    access: ParsedAccountAccess
+  ) => string
```
sam-goldman, line 118: There seems to be a mismatch between the type definition for `handleNetworkSpecificMerkleLeafGas` and the parameters that it's actually invoked with. This definition doesn't have the `chainId` as the first parameter, but this function is called with the chain ID as the first parameter throughout the codebase

sam-goldman: This also seems to apply to the definition for `calculateActionLeafGasForMoonbeam`

packages/contracts/contracts/foundry/SphinxPluginTypes.sol
```diff
@@ -75,6 +75,11 @@ struct ParsedAccountAccess {
     Vm.AccountAccess[] nested;
 }
 
+struct DeployedContractSize {
+    address account;
+    uint size;
+}
```
sam-goldman, line 81: I think this PR could've been simplified a decent amount by just using the `deployedCode` field of the `Vm.AccountAccess` type, which already exists in the `ParsedAccountAccess`

sam-goldman (commented):
Left some comments and questions

Generally, it’d be much easier for me to review non-trivial logic like this if there’s a test case that I can step through. Otherwise, I need to mentally walk through the behavior of the logic, which is error-prone and difficult

sam-goldman:
Just going to list some sources of gas cost variation that this PR doesn’t account for. Nothing in this is necessarily actionable - I just want to record this somewhere. I didn’t include the fact that this PR only loosely estimates the storage costs since you already mentioned that.

* This PR doesn’t implement the first three bullet points in Moonbeam's ["Key Differences with Ethereum” section](https://docs.moonbeam.network/builders/get-started/eth-compare/tx-fees/). 
* In Moonbeam’s [documentation page](https://docs.moonbeam.network/builders/get-started/eth-compare/tx-fees/), a call to the `sayHello` view function uses more gas than Ethereum. I think this is caused by one of the three bullet points in their docs that I mentioned above
* Moonbeam has a block storage limit, which is again mentioned in their documentation page. I think we could inadvertently create a batch that exceeds this storage limit

sam-goldman:
I think this approach is fine in the short term, but I think it’d be much safer and reliable to figure out a way to use eth_estimateGas to calculate the Merkle leaf gas values. Foundry is somehow able to use the real eth_estimateGas method in their simulation ([source](https://github.com/foundry-rs/foundry/pull/6460/files)), so it’s probably possible for us to do this too. I think this'd be pretty difficult to do, but it seems to be the most reliable method, especially if there are other networks like Moonbeam that implement their own gas fee mechanisms. Obviously that’s out of scope for this PR - I’m happy to make a ticket for that, or you can if you want. Just lmk

sam-goldman:
Did you manually check that this works on Moonbeam?

RPate97:
> Did you manually check that this works on Moonbeam?

yes

packages/contracts/contracts/foundry/SphinxPluginTypes.sol
RPate97: Unfortunately, I found that `deployedCode` isn't reliably available I believe due to a bug in Foundry. If you want, I can send you the replication steps. To my knowledge, this doesn't cause a problem for us, but I was going to bring it up in the next sync. [I created a ticket for it.](https://linear.app/chugsplash/issue/CHU-657/investigate-if-missing-deployedcode-in-accountaccesses-could-cause)

I'll add a comment documenting this.

packages/contracts/src/networks.ts
```diff
@@ -1,3 +1,90 @@
...
+ * related networks is comfortably high (15 million), this does not impede the users ability to deploy
+ * large contracts.
+ */
+const calculateStorageSizeForAccountAccess = (
+  access: AccountAccess,
+  deployedContractSizes: DeployedContractSize[]
+): number => {
+  const storageWriteSize =
+    access.storageAccesses.filter((storageAccess) => storageAccess.isWrite)
+      .length + 32
```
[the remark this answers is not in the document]
RPate97: This should be `* 32` because it's 32 bytes per write. Each write pertains to a single storage slot.

packages/contracts/src/networks.ts
```diff
@@ -1,3 +1,90 @@
...
+    if (!deployedContractSize) {
+      throw new Error(
+        'Failed to find deployed contract size. This is a bug, please report it to the developers.'
+      )
+    }
+
+    return storageWriteSize + Number(deployedContractSize.size)
+  } else {
+    return storageWriteSize
+  }
```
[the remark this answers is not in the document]
RPate97: It should return a non-negative number for any `AccountAccessKind` including `Create`.

packages/contracts/src/networks.ts
RPate97: I agree, that is less than ideal. The reason why I didn't use logic that can fetch the actual values is that I believe we would only be able to fetch the block gas limit and not the other parameters. I think it would be slightly more dangerous to use a potentially fluctuating value for the block gas limit if we do not have a way to fetch the real values for the other parameters, than is to just use a static ratio like this.

I can provide more details on that if you're interested.

packages/contracts/src/networks.ts
RPate97: Good question, this is difficult to get exactly right because we're trying to subtract the exact cost of storing the data. SSTORE is a pretty complicated opcode. The gas cost is highly situational and depends on if there's data in the slot, if the slot is warm etc. The gas formula is designed to consider the processing power to complete the operation in addition to the intrinsic storage cost.

The 200 used here comes from the code deposit cost of 200 gas/byte. On the website you linked, you can see this for the create opcode if you go to the calculator and enter it 1 for the bottom row `Estimated deployed code size in bytes`. You'll see it adds 200 dynamic gas to the cost. You can also read about this [here](https://github.com/wolflo/evm-opcodes/blob/main/gas.md#a9-f-code-deposit-cost). Unfortunately, I couldn't find the exact EIP where this is defined. This cost is also applied to the create2 opcode.

I had seen documentation elsewhere that implied that the gas cost per byte for other storage methods was the same. However, after spending a while looking I wasn't able to find concrete documentation on that. After reading through the documentation on SSTORE, it doesn't seem like SSTORE has a similar type of direct cost per byte to what create and create2 have.

So, I think the safest thing to do would be to only subtract the 200 bytes for the storage size of contracts and not subtract anything for other storage mechanisms. That way we won't wildly overestimate the cost when deploying contracts, and we won't run the risk of subtracting too much because of other writes.

RPate97: Just documenting my thoughts on this here. Empirically, this does not appear to be the case because we are able to very reliably execute deployments when subtracting the 200 gas / byte cost for contract deployments.

RPate97: Yeah possibly, I think that's another argument for only subtracting the 200 gas for contract storage.

packages/contracts/src/networks.ts
RPate97: There were actually two functions with the same name which is obviously not ideal.

packages/contracts/contracts/foundry/SphinxPluginTypes.sol
sam-goldman: Got it. We could probably just manually set the `deployedCode` field in the account accesses via `address(access.account).code`, assuming that this returns the correct bytecode

sam-goldman: Not important enough to block this PR, but just pointing that out

sam-goldman (approved):
Looks good

PR #1514, opened 2024-03-03
feat(pg): add support for arbitrary entry point functions in the user's script

sam-goldman:
One note: If the user specifies parameters for `--sig` that result in valid calldata, but don't correspond to a function in their script, the error message isn't good. "Valid calldata" means that the function signature is correctly formatted, e.g. `deploy()`. This isn't trivial to resolve, so I made a [ticket](https://linear.app/chugsplash/issue/CHU-655/improve-error-message-when-user-calls-a-non-existent-function-on-their) to improve this error message.

If the user specifies parameters for `--sig` that result in _invalid_ calldata, then we throw a good error message (in `parseScriptFunctionCalldata`). E.g. if they specify an invalid function selector like `--sig dep(123)loy`, we'll throw a good error


docs/cli-deploy.md
```diff
@@ -68,8 +68,10 @@ pnpm sphinx deploy <SCRIPT_PATH> --network <NETWORK_NAME> [options]
 
 ### Options
 - `--network <NETWORK_NAME>`: **Required**. The name of the network to deploy on, which must match a network in the `rpc_endpoints` section of your `foundry.toml`.
+- `--sig <SIGNATURE [PARAMETERS...] | CALLDATA>` (Alias: `-s`): **Optional**. The signature of the function to call in the script, or raw calldata. Matches the interface of Forge Script's `--sig` parameter.
+  - **Default**: `run()`
 - `--confirm`: **Optional**. Confirm the deployment without previewing it. Useful for automated deployments.
-- `--target-contract <TARGET_CONTRACT>` (Alias: `--tc <TARGET_CONTRACT>`): **Optional**. Specify a contract within the script file. Necessary for scripts with multiple contracts.
```
sam-goldman, line 72: Removed `<TARGET_CONTRACT>` to make the format of aliases consistent

packages/plugins/src/foundry/utils/index.ts
```diff
@@ -1489,6 +1494,67 @@ const isStorageAccess = (
...
 }
 
+/**
+ * Parses the parameters to `--sig` specified by the user. We attempt to maintain the exact same
+ * interface as Foundry's `--sig` parameter to make it easy for users to migrate their Forge Scripts
+ * to Sphinx.
+ *
+ * @returns { string } The 0x-prefixed raw calldata.
+ */
+export const parseScriptFunctionCalldata = async (
```
sam-goldman, line 1504: This function is the only non-trivial aspect of this PR

docs/cli-existing-project.md
```diff
@@ -120,11 +120,11 @@ contract MyDeploymentScript is
...
 ```
 
-### c. Update your `run()` function
+### c. Add the `sphinx` modifier
 
-The entry point of your deployment script must be a `run()` function; it cannot be named anything else. Please change its name if necessary.
+Navigate to the entry point function in your deployment script. This is typically a `run()` function.
 
-Then, add a `sphinx` modifier to your `run` function:
+Then, add a `sphinx` modifier to this function. For example:
```
RPate97, line 127: Since we're mentioning the possibility that the entry point is not the run function. We should update the `Propose on testnets` section to cover what you would need to do if you're using an entry point other than run. 

If we don't, people will assume they can use other entry points and immediately run into an error when attempting to propose. This is made worse because we won't immediately have good handling for the error they are likely to run into.

docs/cli-deploy.md
```diff
@@ -68,8 +68,10 @@ pnpm sphinx deploy <SCRIPT_PATH> --network <NETWORK_NAME> [options]
 
 ### Options
 - `--network <NETWORK_NAME>`: **Required**. The name of the network to deploy on, which must match a network in the `rpc_endpoints` section of your `foundry.toml`.
+- `--sig <SIGNATURE [PARAMETERS...] | CALLDATA>` (Alias: `-s`): **Optional**. The signature of the function to call in the script, or raw calldata. Matches the interface of Forge Script's `--sig` parameter.
```
RPate97, line 71: The alias listed here is `-s`, but the alias for `--target-contract` is `--tc` which has two dashes. I would expect these to be the same format. Is one of them wrong?

docs/cli-propose.md
```diff
@@ -40,10 +40,12 @@ pnpm sphinx propose <SCRIPT_PATH> --networks <NETWORK_NAMES...|testnets|mainnets
   - Arbitrary network names (e.g., `ethereum optimism arbitrum`): Propose on one or more networks, which must match the network names in the `rpc_endpoints` section of your `foundry.toml`.
   - `testnets`: Propose on the test networks in your `sphinxConfig.testnets` array. Provides a convenient way to propose on many networks without specifying them on the command line. Requires additional configuration; see the [Configuration Options section for `sphinxConfig.testnets`](https://github.com/sphinx-labs/sphinx/blob/main/docs/configuration-options.md#string-testnets-optional).
   - `mainnets`: Propose on the production networks in your `sphinxConfig.mainnets` array. Provides a convenient way to propose on many networks without specifying them on the command line. Requires additional configuration; see the [Configuration Options section for `sphinxConfig.mainnets`](https://github.com/sphinx-labs/sphinx/blob/main/docs/configuration-options.md#string-mainnets-optional).
+- `--sig <SIGNATURE [PARAMETERS...] | CALLDATA>` (Alias: `-s`): **Optional**. The signature of the function to call in the script, or raw calldata. Matches the interface of Forge Script's `--sig` parameter.
```
RPate97, line 43: > The alias listed here is `-s`, but the alias for `--target-contract` is `--tc` which has two dashes. I would expect these to be the same format. Is one of them wrong?

RPate97 (changes requested):
Looks pretty good, just had a couple comments on the docs.

I think we need to add something to the integration guide to cover how the user calls the propose script if they have a custom run function.

docs/cli-deploy.md
sam-goldman: Nope, they're actually both correct. This matches Foundry's aliases, which have a different number of dashes for whatever reason. Even though it's bizarre, I think we should maintain the same UX as Foundry to make existing script commands portable to Sphinx

sam-goldman: To be clear, `--s` and `-tc` both don't work for Forge Scripts

docs/cli-existing-project.md
sam-goldman: Yeah good call. Added this

RPate97 (approved):
Looks good

PR #1516, opened 2024-03-04
feat(core): Smoothly handle blockexplorer outages on the website

RPate97:
## Purpose
- Changes the `etherscan` field in the SPHINX_NETWORKS array to support multiple block explorers
- Adds support for Blockscout on all networks where it's running
- Changes the name of some of the environment variables used for block explorer keys to clearly differentiate between etherscan and blockscout
- Adds logic to the deploy task to handle choosing the block explorer if there are multiple available. We chose etherscan if available before falling back to blockscout

### DIFF
--- a/packages/contracts/src/networks.ts
+++ b/packages/contracts/src/networks.ts
@@ -116,17 +116,28 @@ export const calculateActionLeafGasForMoonbeam = (
   return (baseGasLessStorageCost + moonbeamStorageCost).toString()
 }
 
+export type ExplorerName = 'Blockscout' | 'Etherscan'
+
+export type BlockExplorers = {
+  etherscan?: {
+    apiURL: string
+    browserURL: string
+    envKey: string
+  }
+  blockscout?: {
+    apiURL: string
+    browserURL: string
+    envKey: string
+    selfHosted: boolean
+  }
+}
+
 export type SupportedNetwork = {
   name: string
   displayName: string
   chainId: bigint
   rpcUrl: () => string
-  etherscan: {
-    apiURL: string
-    browserURL: string
-    envKey: string
-    blockExplorer: BlockExplorer
-  }
+  blockexplorers: BlockExplorers
   currency: string
   dripSize: string
   requiredEnvVariables: Array<string>
@@ -171,7 +182,6 @@ export const SPHINX_LOCAL_NETWORKS: Array<SupportedLocalNetwork> = [
   },
 ]
 
-type BlockExplorer = 'Etherscan' | 'Blockscout' | 'Unsupported'
 export type NetworkType = 'Testnet' | 'Mainnet' | 'Local'
 type RollupProvider = 'Conduit' | 'Caldera'
 type RollupType = 'OP Stack' | 'Arbitrum'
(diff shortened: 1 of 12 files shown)


packages/core/src/etherscan.ts
```diff
@@ -26,39 +26,22 @@ import {
...
-    ...[...customChains].reverse(), // the last entry has higher precedence
-  ].find((config) => config.chainId === chainId)
-
-  if (chainConfig === undefined) {
-    throw new Error(`Could not find chain config for: ${chainId}`)
-  }
-
-  return chainConfig
-}
-
```
RPate97, line 54: I removed this in favor of using a utility function that pulls the data from the `SPHINX_NETWORKS` array which is consistent with the way we handle all other network specific data.

packages/core/src/networks.ts
```diff
@@ -122,11 +124,85 @@ export const fetchDripVersionForNetwork = (chainId: bigint) => {
...
+  chainId: bigint,
+  mockSphinxNetworks?: any
+) => {
+  return (
+    isEtherscanSupportedForNetwork(chainId, mockSphinxNetworks) ||
+    isBlockscoutSupportedForNetwork(chainId, mockSphinxNetworks)
+  )
+}
+
+export const fetchEtherscanConfigForNetwork = (
```
RPate97, line 167: This is a surprisingly complex function. You can look at the test file below for a good overview of the different cases it covers. 

There are basically two cases this function is useful for:
1. We have a specific block explorer, and we want the config for exactly that block explorer
2. We don't know what block explorers are available on the network and just want to get the first explorer available, defaulting to etherscan if more than one is available.

We could have used two separate functions for this, but then we'd have to change the verifySphinxConfig function to handle that and then we'd end up with even less code shared between the website and deploy command. So I decided to just do this and make sure the function is well tested.

packages/core/src/actions/execute.ts
```diff
@@ -92,12 +92,7 @@ export type Deployment = {
   treeSigners: Array<TreeSigner>
 }
 
-export type HandleError = (
-  e: any,
-  deployment: Deployment,
-  deploymentTransactionReceipts: ethers.TransactionReceipt[],
-  wallet?: ethers.Wallet
-) => Promise<void>
+export type HandleError = (e: any, deployment: Deployment) => Promise<void>
```
RPate97, line 95: I simplified many of the functions in the `deploymentContext` since verification now happens outside of the deployment process on the website. 

I also removed some other fields that are no longer required by the website for various reasons.

packages/contracts/src/networks.ts
```diff
@@ -60,11 +70,18 @@ export const SPHINX_NETWORKS: Array<SupportedNetwork> = [
...
+        apiURL: 'https://api.etherscan.io/api',
+        browserURL: 'https://etherscan.io',
+        envKey: 'ETH_ETHERSCAN_API_KEY',
+      },
+      blockscout: {
+        apiURL: 'https://eth.blockscout.com/api',
+        browserURL: 'https://eth.blockscout.com/',
+        envKey: 'ETH_BLOCKSCOUT_API_KEY',
+        selfHosted: false,
+      },
```
RPate97, line 84: I'm still testing verification on the newly supported blockscout networks, so these fields may change. I don't think that needs to block the review though.

packages/core/src/etherscan.ts
```diff
@@ -194,9 +178,16 @@ export const attemptVerification = async (
   minimumCompilerInput: SolcInput,
   provider: ethers.Provider,
   chainId: string,
-  etherscanApiKey: string
+  etherscanApiKey: string,
+  explorer?: ExplorerName
 ): Promise<{ success: boolean; message?: string }> => {
-  const { urls } = getChainConfig(Number(chainId))
+  const urls = fetchEtherscanConfigForNetwork(BigInt(chainId), explorer)
```
sam-goldman, line 184: Does this line throw an error if the network isn't supported by the DevOps platform? If so, will the Deploy command fail during verification for unsupported networks? We don't need to resolve that in this PR, but could you make a ticket for it?

sam-goldman (approved):
Looks good. Could you update the main README to say "block explorers" instead of "Etherscan"?

packages/core/src/etherscan.ts
RPate97: This line throws an error if verification (via either Blockscout or Etherscan) isn't supported on this network period, which I believe is the current behavior. 

It would definitely be better to error at the beginning of the deployment if the user requests verification on a network where we don't support verification. [I made a ticket for that.](https://linear.app/chugsplash/issue/CHU-662/error-before-the-deployment-starts-if-the-user-request-verification-on)

PR #1522, opened 2024-03-05
fix(pg): check that contracts are deployed in simulation

sam-goldman:
Adds logic that would have caught the error in this ticket: [Contract addresses not properly calculated for new keyword](https://linear.app/chugsplash/issue/CHU-659/contract-addresses-not-properly-calculated-for-new-keyword). Specifically, we now check that all of the contracts in the NetworkConfig have been deployed at the end of the simulation. I also updated an integration test in `deploy.spec.ts` to deploy the same project twice, which would have caused this error if we were still using `FOUNDRY_SENDER`.

Notes:
* This PR doesn't fix the `FOUNDRY_SENDER` error in Foundry. We created [this ticket](https://linear.app/chugsplash/issue/CHU-681/fix-foundry-sender-bug-in-foundrys-codebase) to fix it in Foundry.
* We'll add `FOUNDRY_SENDER` back to our logic when we support linked libraries. ([Here's the relevant ticket for supporting linked libraries](https://linear.app/chugsplash/issue/CHU-297/support-linked-libraries))

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -1,6 +1,7 @@
 // Converts BigInt values to strings when calling JSON.stringify. An error would be thrown
 // otherwise. For more context, see these sources:
 // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt#use_within_json
+
 // https://github.com/GoogleChromeLabs/jsbi/issues/30#issuecomment-1006088574
 BigInt.prototype['toJSON'] = function () {
   return this.toString()
@@ -19,3 +20,4 @@ export * from './preview'
 export * from './provider'
 export * from './artifacts'
 export * from './types'
+export * from './errors'
--- a/packages/core/src/utils.ts
+++ b/packages/core/src/utils.ts
@@ -1544,7 +1544,7 @@ export const isCreateActionInput = (
     typeof create.value === 'string' &&
     typeof create.txData === 'string' &&
     typeof create.gas === 'string' &&
-    create.operation === Operation.Call &&
+    create.operation === Operation.DelegateCall &&
     typeof create.requireSuccess === 'boolean'
   )
 }
@@ -1633,6 +1633,18 @@ export const readDeploymentArtifactsForNetwork = (
 }
 
 export const isArrayMixed = <T>(arr: T[]): boolean => new Set(arr).size > 1
+
+export const getContractAddressesFromNetworkConfig = (
+  networkConfig: NetworkConfig
+): Array<string> => {
+  const unlabeledAddresses = networkConfig.unlabeledContracts.map(
+    (ct) => ct.address
+  )
+  const labeledAddresses = networkConfig.actionInputs
+    .flatMap((actions) => actions.contracts)
+    .map((ct) => ct.address)
+  return unlabeledAddresses.concat(labeledAddresses)
+}
 
 /**
  * Checks if a string contains an opening then closing parenthesis.
--- a/packages/plugins/test/mocha/cli/deploy.spec.ts
+++ b/packages/plugins/test/mocha/cli/deploy.spec.ts
@@ -6,7 +6,6 @@ import chaiAsPromised from 'chai-as-promised'
 import {
   DeploymentConfig,
   ConfigArtifacts,
-  Create2ActionInput,
   DeploymentArtifacts,
   ExecutionMode,
   NetworkConfig,
@@ -19,12 +18,17 @@ import {
   makeDeploymentArtifacts,
   setBalance,
   isLiveNetwork,
+  InitialChainState,
+  getContractAddressesFromNetworkConfig,
+  ActionInputType,
+  ActionInput,
 } from '@sphinx-labs/core'
-import { ethers } from 'ethers'
+import { ethers, keccak256 } from 'ethers'
 import {
   CREATE3_PROXY_INITCODE,
   DETERMINISTIC_DEPLOYMENT_PROXY_ADDRESS,
   SphinxMerkleTree,
+  parseFoundryContractArtifact,
 } from '@sphinx-labs/contracts'
 
 import * as MyContract2Artifact from '../../../out/artifacts/MyContracts.sol/MyContract2.json'
@@ -39,8 +43,12 @@ import {
   startAnvilNodes,
   getSphinxModuleAddressFromScript,
   getEmptyDeploymentArtifacts,
+  getGnosisSafeProxyAddress,
+  makeActionInputsWithoutGas,
+  encodeFunctionCalldata,
 } from '../common'
 import { makeMockSphinxContextForIntegrationTests } from '../mock'
+import { getFakeConfigArtifactsFromContractArtifacts } from '../fake'
 
 const coder = new ethers.AbiCoder()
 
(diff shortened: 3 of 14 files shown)


packages/contracts/contracts/foundry/Sphinx.sol
```diff
@@ -314,21 +314,21 @@ abstract contract Sphinx {
...
         if (callerMode == VmSafe.CallerMode.RecurrentPrank) vm.stopPrank();
 
         sphinxUtils.fetchAndValidateConfig(address(this));
 
-        // Prank the Gnosis Safe then execute the user's script. We prank the Gnosis
-        // Safe to replicate the production environment.
-        vm.startPrank(safeAddress());
+        // Broadcast from the Gnosis Safe then execute the user's script. We broadcast from the
+        // Gnosis Safe to replicate the production environment.
+        vm.startBroadcast(safeAddress());
```
sam-goldman, line 329: Broadcasting instead of pranking solves the [contract address miscalculation ticket](https://linear.app/chugsplash/issue/CHU-659/contract-addresses-not-properly-calculated-for-new-keyword)

packages/core/src/utils.ts
```diff
@@ -1544,7 +1544,7 @@ export const isCreateActionInput = (
     typeof create.value === 'string' &&
     typeof create.txData === 'string' &&
     typeof create.gas === 'string' &&
-    create.operation === Operation.Call &&
+    create.operation === Operation.DelegateCall &&
```
sam-goldman, line 1547: This is irrelevant to the PR, but I noticed that the `Operation` was incorrect. This doesn't impact anything because we don't currently use this function

packages/plugins/test/mocha/cli/deploy.spec.ts
```diff
@@ -6,7 +6,6 @@ import chaiAsPromised from 'chai-as-promised'
 import {
   DeploymentConfig,
   ConfigArtifacts,
-  Create2ActionInput,
```
sam-goldman, line 9: The diff in this file is a mess, so you don't need to scrutinize it. I overhauled the structure of the integration tests in this file to reduce the amount of boilerplate code, since this made it easier to write the regression test, which executes two deployments.

packages/plugins/test/mocha/cli/deploy.spec.ts
```diff
@@ -718,51 +721,14 @@ const expectValidDeployment = async (
   expect(await provider.getCode(expectedContractAddress)).to.not.equal('0x')
   expect(await contract.number()).to.equal(BigInt(2))
 
-  expect(preview).to.deep.equal({
```
sam-goldman, line 721: It was pointless to check the preview here because it's not used for anything when it's returned by the `deploy` function. It was a maintenance burden, so I removed it from these tests. It'd make more sense to check the `actionInput` array, since that's a direct input to the preview, but it wasn't necessary for this PR, so I didn't bother

packages/plugins/test/mocha/cli/deploy.spec.ts
```diff
@@ -690,26 +664,55 @@ const getContract = (address: string, artifact: any): ethers.Contract => {
...
 
-  expect(
-    (networkConfig.actionInputs[0] as Create2ActionInput).create2Address
-  ).to.equal(expectedContractAddress)
+  expect(networkConfig.executionMode).equals(expectedExecutionMode)
+
+  const contractAddresses = getContractAddressesFromNetworkConfig(networkConfig)
+  for (const address of contractAddresses) {
+    expect(await provider.getCode(address)).does.not.equal('0x')
+  }
```
sam-goldman, line 714: This catches the error that we were running into

packages/plugins/test/mocha/dummy.ts
```diff
@@ -285,3 +306,37 @@ export const getDummyDeploymentArtifacts = (): DeploymentArtifacts => {
     },
   }
 }
+
+export const getDummyDeploymentContext = (): DeploymentContext => {
```
sam-goldman, line 310: Ended up not using this, but I kept it in b/c I figured it'll probably be useful at some point soon

packages/plugins/test/mocha/cli/deploy.spec.ts
```diff
@@ -136,9 +132,19 @@ describe('Deploy CLI command', () => {
   })
 
   describe('With preview', () => {
-    const projectName = 'Simple_Project_1'
+    it('Executes deployment on local network twice', async () => {
```
sam-goldman, line 135: Updated this integration test to execute two deployments so that it serves the purpose of a regression test. The error would have occurred on the second deployment

packages/core/src/index.ts
```diff
@@ -19,3 +22,8 @@ export * from './preview'
 export * from './provider'
 export * from './artifacts'
 export * from './types'
+export * from './errors'
+
+export const SphinxCore = {
+  compileAndExecuteDeployment,
+}
```
sam-goldman, line 29: I used this export pattern for `compileAndExecuteDeployment` so that I can stub it in `simulate.spec.ts` when testing `simulateDeploymentSubtask`. We can't pass a mocked function as an input parameter to `simulateDeploymentSubtask` (like [this](https://github.com/sphinx-labs/sphinx/blob/main/packages/core/src/networks.ts#L150-L153)) because this function is called as a Hardhat subtask.

If we don't export `compileAndExecuteDeployment` as a member of an object, this error occurs when attempting to stub it: `TypeError: Descriptor for property compileAndExecuteDeployment is non-configurable and non-writable`

RPate97:
@sam-goldman look at your telegram please

packages/plugins/test/mocha/cli/deploy.spec.ts
RPate97: What do you mean it's not used for anything in the deploy function? Isn't the preview displayed in the deploy command? I thought the purpose of returning and checking this here was to confirm that the preview is correct.

RPate97 (changes requested):
Looks pretty good. I think we should use [p-limit](https://www.npmjs.com/package/p-limit) instead of implementing our own promise batching logic.

packages/core/src/utils.ts
```diff
@@ -1634,6 +1634,32 @@ export const readDeploymentArtifactsForNetwork = (
...
+  fn: (item: T) => Promise<R>
+): Promise<R[]> => {
+  const results: R[] = []
+  for (let i = 0; i < items.length; i += batchSize) {
+    const batchItems = items.slice(i, i + batchSize)
+    const batchResults = await Promise.all(batchItems.map((item) => fn(item)))
+    results.push(...batchResults)
+  }
+  return results
+}
```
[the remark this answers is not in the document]
sam-goldman: Changed this to use `p-limit`

packages/plugins/test/mocha/cli/deploy.spec.ts
sam-goldman: Yeah thanks for fact checking me. I updated this test suite to check for the presence of the correct action inputs instead of checking for the preview since this is more comprehensive (the action inputs contain several more fields than the preview), maintainable (the action inputs are less likely to change than the structure of the preview since changing the action inputs is a breaking change), and logically equivalent (since the preview is created directly from the action inputs and the other fields that we already check in the network config)

RPate97 (approved):
Looks good

PR #1527, opened 2024-03-06
fix(ct): Detect missing depth field

RPate97:
## Purpose
Adds logic to the foundry version validation process to detect if the `depth` field is not included.


sam-goldman (approved):
Looks good, just a comment

PR #1528, opened 2024-03-06
fix(ct): filter calls to SphinxUtils out of state diff

RPate97:
## Purpose
Fixes an issue where calls to the SphinxUtils contract could be included in the deployment if they happened during the user script. I.e if the user called `safeAddress()`, then the call to SphinxUtils to fetch the address would have been included.

### DIFF
--- a/packages/contracts/contracts/foundry/SphinxUtils.sol
+++ b/packages/contracts/contracts/foundry/SphinxUtils.sol
@@ -849,7 +849,7 @@ contract SphinxUtils is SphinxConstants, StdUtils {
     function getNumRootAccountAccesses(
         Vm.AccountAccess[] memory _accesses,
         address _safeAddress
-    ) private pure returns (uint256) {
+    ) private view returns (uint256) {
         uint256 count = 0;
         for (uint256 i = 0; i < _accesses.length; i++) {
             Vm.AccountAccess memory access = _accesses[i];
@@ -874,14 +874,17 @@ contract SphinxUtils is SphinxConstants, StdUtils {
      * - The call depth is equal to 2. The expected depth is 2 because the depth value starts
      * at 1 and because we initiate the collection process by doing a delegatecall to the entry
      * point function so the depth is 2 by the time any transactions get sent in the users script.
+     * - The target contract is not `SphinxUtils`. This can occur if the user calls a function
+     * that calls into this contract during their script. I.e calling safeAddress().
      */
     function isRootAccountAccess(
         Vm.AccountAccess memory _access,
         address _safeAddress
-    ) private pure returns (bool) {
+    ) private view returns (bool) {
         return
             _access.accessor == _safeAddress &&
             _access.depth == 2 &&
+            _access.account != address(this) &&
             (_access.kind == VmSafe.AccountAccessKind.Call ||
                 _access.kind == VmSafe.AccountAccessKind.Create);
     }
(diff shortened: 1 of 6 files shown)


packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -874,14 +874,17 @@ contract SphinxUtils is SphinxConstants, StdUtils {
...
      */
     function isRootAccountAccess(
         Vm.AccountAccess memory _access,
         address _safeAddress
-    ) private pure returns (bool) {
+    ) private view returns (bool) {
         return
             _access.accessor == _safeAddress &&
             _access.depth == 2 &&
+            _access.account != address(this) &&
```
RPate97, line 887: This is the fix.

packages/plugins/test/mocha/cli/propose.spec.ts
```diff
@@ -715,6 +716,93 @@ describe('Propose CLI command', () => {
 
     expect(errorThrown).to.be.true
   })
+
+  describe('Issues', () => {
+    it('CHU-676: Deploys with call to safeAddress() in script', async () => {
```
RPate97, line 721: Couple thoughts here: 
1. We should start following a pattern where we label which tests are related to which issues, ideally, we would also organize regression tests together into suites. I've tried to start that pattern here. The Foundry devs do this, and it's really nice for keeping things organized. It also makes it easy to find additional context on specific tests, which is super useful if you ever make a change that breaks them. 

2. I implemented the regression test for this issue as a full-on integration test. This is less than ideal for regression tests like this b/c we will eventually have a ton of them. We should generally try to avoid doing this, especially if the test is specific to the Solidity library contracts. Ideally, the test for this would have been a simple Foundry test that calls the sphinxCollect function and checks the results. I put what I would consider a reasonable amount of effort into trying to do that and found it impractical in this case.

packages/plugins/contracts/test/script/issues/Owned.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+contract Owned {
+    address public owner;
+
+    constructor(address _owner) {
+        owner = _owner;
+    }
+}
```
sam-goldman, line 10: thoughts reusing an existing contract in `MyContracts.sol`? imo it's better if we reuse contracts when possible to avoid having a bunch of contracts like this that are almost the same

packages/plugins/test/mocha/cli/propose.spec.ts
sam-goldman: > We should start following a pattern where we label which tests are related to which issues, ideally, we would also organize regression tests together into suites. I've tried to start that pattern here. The Foundry devs do this, and it's really nice for keeping things organized. It also makes it easy to find additional context on specific tests, which is super useful if you ever make a change that breaks them.

i like the idea of labeling tests with tickets, but i think i need clarity on how strictly we want to adhere to this pattern. for example, say the easiest way to implement a regression test is to add slightly more coverage to an existing test case, or to an existing helper function that a lot of test cases use (like `assertValidProposalRequest`). should we make an entirely different test case in this scenario? i think this deserves its own discussion, so i made a note to talk about it on fri

sam-goldman (approved):
Looks good, but I don't understand how our existing proposal tests didn't already detect this case, since the `sphinx` modifier seems to make two calls to `SphinxUtils`: `safeAddress()` and `fetchAndValidateConfig`. Do you know why?

RPate97:
> Looks good, but I don't understand how our existing proposal tests didn't already detect this case, since the `sphinx` modifier seems to make two calls to `SphinxUtils`: `safeAddress()` and `fetchAndValidateConfig`. Do you know why?

`isRootAccountAccess` already filters out any accesses where the accessor is not the Safe. We don't start pranking the safe in the modifier until after we've already completed both of those calls to SphinxUtils, so they both just get filtered out.

PR #1529, opened 2024-03-06
fix(ct): Prevent infinite loop in configureSphinx function

RPate97:
## Purpose
Resolves an infinite loop that can occur if the user calls `safeAddress()` or some other function that requires the `sphinxConfig` from within their `configureSphinx()` function.

### DIFF
--- a/packages/contracts/test/issues/CHU663/CHU663.s.sol
+++ b/packages/contracts/test/issues/CHU663/CHU663.s.sol
@@ -1,0 +1,23 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "../../../contracts/forge-std/src/Script.sol";
+import "../../../contracts/forge-std/src/Test.sol";
+import { Sphinx } from "../../../contracts/foundry/Sphinx.sol";
+
+contract CHU663_test is Test, Sphinx {
+    address public safe;
+
+    function configureSphinx() public override {
+        sphinxConfig.projectName = "CHU-663";
+        sphinxConfig.owners = [0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266];
+        sphinxConfig.threshold = 1;
+        sphinxConfig.orgId = "test-org-id";
+        safe = safeAddress();
+    }
+
+    function test_configureSphinx_success_calls_safeAddress() external {
+        configureSphinx();
+        assertNotEq(safe, address(0));
+    }
+}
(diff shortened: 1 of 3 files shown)


docs/writing-scripts.md
```diff
@@ -51,6 +51,8 @@ function configureSphinx() public override {
 
 See the [Configuration Reference](https://github.com/sphinx-labs/sphinx/blob/main/docs/configuration-options.md) if you'd like to learn more about these fields.
 
+> Note that we do not recommend calling utility functions such as `safeAddress()` or `moduleAddress()` from your `configureSphinx()` because these utilities rely on fields in the sphinxConfig object. If you attempt to use them in your `configureSphinx()` function, you may run into a `Sphinx: Detected missing Sphinx config` error. The solution is to move the utility function call to the end of your `configureSphinx()` function or remove it entirely.
```
RPate97, line 54: If the user calls `safeAddress()` at the beginning of their `configureSphinx()` function such that the fields in the `sphinxConfig` are not defined, they can get an error message indicating they haven't implemented the `configureSphinx()` function correctly. I just addressed that edge case here instead of updating the error message since I think that situation will be relatively uncommon.

sam-goldman: this is a pretty obscure thing to put in a guide that we want everyone to read. pretty strong preference to put this in the troubleshooting guide or an error message or something

sam-goldman: to be clear, i haven't thought about how we could detect and throw an error for this case specifically. just pointing out what i think is the best ux

packages/contracts/test/issues/CHU663/CHU663.s.sol
```diff
@@ -0,0 +1,23 @@
...
+        sphinxConfig.threshold = 1;
+        sphinxConfig.orgId = "test-org-id";
+        safe = safeAddress();
+    }
+
+    function test_configureSphinx_success_calls_safeAddress() external {
+        configureSphinx();
+        assertNotEq(safe, address(0));
+    }
+}
```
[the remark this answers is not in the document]
RPate97: Generally, I think our current test suite is quite unmanageable as it is largely due to the fact that we're continuously adding more tests to monolithic files, which make it difficult to look for any particular thing. I don't think the current pattern is maintainable in the long term because we will end up with ridiculously huge test files for SphinxUtils.sol and Sphinx.sol. 

This test is an example of a pattern implemented in many other places that is significantly better in my opinion, but it's just one option. I'm happy to talk more about it on Friday.

sam-goldman (approved):
Looks good

PR #1542, opened 2024-03-08
fix(core): Suggest rebuilding with --force if we fail to infer a contract artifact

RPate97:
## Purpose
A temporary solution to the bug where we fail to properly infer a contract artifact where we simply suggest running `forge build --force` if the user thinks we made a mistake.

### DIFF
--- a/packages/core/src/preview.ts
+++ b/packages/core/src/preview.ts
@@ -146,7 +146,8 @@ export const getPreviewString = (
     previewString += yellow(
       `This typically happens when deploying contracts using hardcoded bytecode and no \n` +
         `associated source file. Sphinx will not create a deployment artifact or attempt \n` +
-        `Etherscan verification for any address in the list above.\n\n`
+        `Etherscan verification for any address in the list above.\n\n` +
+        `If you think this is a mistake, try running "forge build --force", then run your Sphinx command again.\n`
     )
   }
 
--- a/.changeset/short-doors-beg.md
+++ b/.changeset/short-doors-beg.md
@@ -1,0 +1,5 @@
+---
+'@sphinx-labs/core': patch
+---
+
+Suggest running forge build with --force if we fail to infer artifact


packages/core/src/preview.ts
```diff
@@ -146,7 +146,8 @@ export const getPreviewString = (
     previewString += yellow(
       `This typically happens when deploying contracts using hardcoded bytecode and no \n` +
         `associated source file. Sphinx will not create a deployment artifact or attempt \n` +
-        `Etherscan verification for any address in the list above.\n\n`
+        `Etherscan verification for any address in the list above.\n\n` +
+        `If you think this is a mistake, try running "forge build --force", then propose again.\n`
```
sam-goldman, line 150: > then propose again

the user could run into this w/ `sphinx deploy`. maybe say something generic, like "run the command again"

sam-goldman (commented):
Seems like a good short term solution. Did you add a note to the ticket for the real fix mentioning that we should revert this PR?

RPate97:
> Seems like a good short term solution. Did you add a note to the ticket for the real fix mentioning that we should revert this PR?

yeah, that's a good idea just did it.

packages/core/src/preview.ts
```diff
@@ -146,7 +146,8 @@ export const getPreviewString = (
     previewString += yellow(
       `This typically happens when deploying contracts using hardcoded bytecode and no \n` +
         `associated source file. Sphinx will not create a deployment artifact or attempt \n` +
-        `Etherscan verification for any address in the list above.\n\n`
+        `Etherscan verification for any address in the list above.\n\n` +
+        `If you think this is a mistake, try running "forge build --force", then run the command again.\n`
```
sam-goldman, line 150: So sorry, after reading this, I think "try running "forge build --force", then run the command again" sorta sounds like we mean that they should run "forge build --force" twice. Maybe we could say "run your Sphinx command again"

sam-goldman (approved):
One last comment, otherwise lgtm

PR #1544, opened 2024-03-08
fix(ct): Prevent owner from being set to the zero address

RPate97:
## Purpose
- Adds a check preventing the user from setting one of their Safe owners to be the zero address or the `SENTINAL_ADDRESS`


packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -489,6 +489,26 @@ contract SphinxUtils is SphinxConstants, StdUtils {
...
 
+    function assertValidOwners(SphinxConfig memory _config) private pure {
+        require(
+            _config.owners.length > 0,
+            "Sphinx: You must have at least one owner in your 'sphinxConfig.owners' array before calling this function."
+        );
+
+        for (uint256 i = 0; i < _config.owners.length; i++) {
+            // We prevent the owners from being either address(0) or address(0x1)
+            // address(0x1) is the SENTINAL_ADDRESS which is a special address used
```
sam-goldman, line 500: nit: SENTINAL_ADDRESS -> SENTINEL_ADDRESS

packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -489,6 +489,26 @@ contract SphinxUtils is SphinxConstants, StdUtils {
...
+            // by Safe to handle their owner implementation. We check for it here for
+            // completeness, but it's unlikely the user would ever actually use that
```
sam-goldman, line 502: I know it's unlikely that the user specifies `address(1)`, but i don't think it's unlikely enough to warrant throwing a misleading error message. e.g. maybe the user will do `address(0), address(1), address(2)` as placeholders for their real owner addresses

if you want to keep things concise, you could continue to use a single if-statement and error message by doing something like:
```
revert(abi.encodePacked("Sphinx: Detected owner that is, " _config.owners[i], ". You are not allowed to use this address as a Gnosis Safe owner"))
```

packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -489,6 +489,26 @@ contract SphinxUtils is SphinxConstants, StdUtils {
...
+            // value so we don't mention it in the error message.
+            if (_config.owners[i] == address(0) || _config.owners[i] == address(0x1)) {
+                revert(
+                    "Sphinx: Detected owner that is address(0). You are not allowed to use the zero address as a Gnosis Safe owner"
+                );
```
sam-goldman, line 507: nit: I could see this restriction being annoying to users who just want to use temporary placeholder addresses as their owners. It may be nice to direct their annoyance to Gnosis Safe since this isn't our restriction. The current error message implies that it's our restriction, or at least doesn't make it clear that it's Gnosis Safe's.

The second sentence in the error message could be something like, "Gnosis Safe prevents you from using this address as an owner"

sam-goldman (commented):
Looks good, just a couple minor comments

packages/contracts/contracts/foundry/SphinxUtils.sol
RPate97: FYI, your code suggestion doesn't work b/c encodePacked doesn't encode address(0) as a full address string (i.e 0x...00), it just treats it as an empty string.

sam-goldman: Ah yeah. Your change works. I think you could also do `vm.toString(_config.owners[i])`

sam-goldman (approved):
Looks good. I think the contracts package needs to be linted so CI passes

PR #1545, opened 2024-03-08
fix(pg): Throw an error if the simulation is going to fail due to a missing hardhat config

RPate97:
## Purpose
Adds logic to detect if the hardhat config we use for the simulation is not in the expected location. This can occur if the user's package manager is configured to store dependency files in non-standard locations. I've confirmed this is not an issue with PNPM, but I believe it would be an issue with some yarn configurations (especially yarn versions 2 and 3). 

I was not able to come up with a good true solution for this which is why this PR just adds logic to throw a better error message. I believe the ideal solution is to switch to using a precompiled NodeJS binary for installation, which would not be subject to this issue.

### DIFF
--- a/.changeset/sixty-worms-agree.md
+++ b/.changeset/sixty-worms-agree.md
@@ -1,0 +1,5 @@
+---
+'@sphinx-labs/plugins': patch
+---
+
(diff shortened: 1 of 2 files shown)


sam-goldman (approved):
Looks good, nice catch

PR #1547, opened 2024-03-10
fix(pg): support scripts that fork networks

sam-goldman:
Review this PR first: [https://github.com/sphinx-labs/sphinx/pull/1548](https://github.com/sphinx-labs/sphinx/pull/1548)

Makes Sphinx compatible with Forge scripts that fork networks via `vm.selectFork` and related cheatcodes.

Users don't need to configure anything. Our system just works.

## Design

Our collection logic is the same as before except it now discards any account accesses that don't belong to the target network. This means that we still always run the user's script once per network. Here's an example:

Say the user's `run` function looks something like this:
```
vm.createSelectFork('ethereum');
MyContract ct = new MyContract();
vm.createSelectFork('optimism');
MyContract2 ct2 = new MyContract2(ct);
```

Say the user proposes on Ethereum and Optimism. Here's what happens in our TypeScript logic:
1. We'll collect the transactions on Ethereum by calling `Sphinx.sol:sphinxCollectProposal`, which will produce a single account access: the first contract deployment. We'll return back to TypeScript.
2. Then, we'll run the user's Forge script again on Optimism, which will produce a single account access: the second contract deployment.

You'll notice that this is inefficient. Specifically, we should only need to call the user's Forge script once since the user is forking within their script. I decided not to implement it this way for two main reasons:

1. My implementation doesn't require the user to specify some sort of `--multi` flag. I think this is important because including the flag would further complicate the "Integrating an Existing Project" guide, especially because the flag is necessary in only a specific scenario, which is when the user forks within their script.
2. When we support linked libraries, we won't always be able to trigger the user's multifork script using a single Forge Script call. This is because the nonce of the user's Gnosis Safe may not be consistent across the networks that they're forking in their script.

## Notes

* I re-integrated Decent's repo, which has a script that forks a network. The integration was easy and the proposal looked correct.
* I manually checked that this design strategy works with Calnix's use case: "Would it be possible to capture the return value of some function that was executed on chainA, to be passed as a param into another function to be executed on chainB?". In the example above, `MyContract2` on Optimism will contain the address of `MyContract` on Ethereum. This is because memory variables are persisted across forks.


packages/contracts/contracts/foundry/Sphinx.sol
```diff
@@ -141,17 +141,24 @@ abstract contract Sphinx {
 
     function sphinxCollectProposal(
         bytes memory _scriptFunctionCalldata,
-        string memory _deploymentInfoPath
-    ) external {
+        string memory _deploymentInfoPath,
+        uint64 _callDepth
```
sam-goldman, line 145: We now pass the call depth as an input variable because this allows us to call `sphinxCollectProposal` in Forge tests, which is how I tested this feature. This is necessary in Forge tests because the default call depth is 3 instead of 2, since the Forge test adds one layer to the depth

packages/contracts/contracts/foundry/Sphinx.sol
```diff
@@ -141,17 +141,24 @@ abstract contract Sphinx {
...
+    ) external returns (FoundryDeploymentInfo memory) {
```
sam-goldman, line 146: We return the `FoundryDeploymentInfo` to make it easier to use `sphinxCollectProposal` in tests

packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -864,13 +864,15 @@ contract SphinxUtils is SphinxConstants, StdUtils {
 
     function getNumRootAccountAccesses(
         Vm.AccountAccess[] memory _accesses,
-        address _safeAddress
+        address _safeAddress,
```
sam-goldman, line 867: There's only a single change to this file: we now check the `chainId` field of the account accesses

packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -1038,27 +1036,38 @@ contract SphinxUtils is SphinxConstants, StdUtils {
...
+    ) private pure returns (bool) {
+        return
+            _access.kind == VmSafe.AccountAccessKind.Create &&
+            _access.chainInfo.chainId == _chainId;
+    }
+
     function fetchDeployedContractSizes(
-        Vm.AccountAccess[] memory _accesses
+        Vm.AccountAccess[] memory _accesses,
+        uint256 _chainId
```
sam-goldman, line 1062: We're going to remove this function and `fetchNumCreateAccesses` when we remove the Moonbeam heuristic, so I didn't add additional tests for them despite slightly increasing their complexity

packages/contracts/foundry.toml
```diff
@@ -23,6 +23,10 @@ src = 'contracts/core'
 [profile.lite]
 optimizer = false
 
+[rpc_endpoints]
+ethereum = 'http://127.0.0.1:42001'
+optimism = 'http://127.0.0.1:42010'
```
sam-goldman, line 28: We use these in a test in `Sphinx.t.sol`

packages/contracts/contracts/foundry/Sphinx.sol
```diff
@@ -184,44 +191,33 @@ abstract contract Sphinx {
...
-            orgId: sphinxConfig.orgId,
-            mainnets: sphinxConfig.mainnets,
-            testnets: sphinxConfig.testnets,
-            saltNonce: sphinxConfig.saltNonce
-        });
-        deploymentInfo.initialState = sphinxUtils.getInitialChainState(safe, ISphinxModule(module));
-        deploymentInfo.nonce = sphinxUtils.getMerkleRootNonce(ISphinxModule(module));
-        deploymentInfo.sphinxLibraryVersion = sphinxUtils.getSphinxLibraryVersion();
-        deploymentInfo.arbitraryChain = false;
-        deploymentInfo.requireSuccess = true;
```
sam-goldman, line 224: Moved all of this to `SphinxUtils.sol:initializeDeploymentInfo` to prevent the `Solc.spec.ts` test suite in the demo package from failing due to a "stack too deep" error

packages/contracts/contracts/foundry/Sphinx.sol
```diff
@@ -274,13 +267,12 @@ abstract contract Sphinx {
         }
 
         vm.revertTo(snapshotId);
-        deploymentInfo.gasEstimates = _sphinxEstimateMerkleLeafGas(
+        deploymentInfo.gasEstimates = sphinxUtils.estimateMerkleLeafGas(
             parsedAccesses,
-            IGnosisSafe(safe),
-            module
+            address(this)
```
sam-goldman, line 272: I moved the `_estimateMerkleLeafGas` function to `SphinxUtils` to prevent the `Solc.spec.ts` test suite in the demo package from failing due to a "stack too deep" error

RPate97 (approved):
Looks good

PR #1557, opened 2024-03-13
fix(pg): update build info cache to include fields for deployed bytecode exact match

sam-goldman:
Updates the build info cache to include deployed bytecode and other relevant fields, which are necessary to implement option 4 in this ticket: [Improve error message when user calls a non-existent function on their script](https://linear.app/chugsplash/issue/CHU-655/improve-error-message-when-user-calls-a-non-existent-function-on-their). Option 4 requires us to make an exact bytecode match on the script's deployed bytecode, but before this PR, it was only possible to check for an exact match on the _init code_ because we didn't store the deployed bytecode in the build info cache.

I added a `_format` field to the build info cache to signify that its structure has changed. I also added logic that handles outdated build info caches for backwards compatibility.

Note: The `findFullyQualifiedNames` function is currently causing the plugins and demo test suite to fail in CI. I'm planning on removing this function in the next PR, so just disregard the failed tests.

### DIFF
--- a/packages/plugins/src/foundry/types.ts
+++ b/packages/plugins/src/foundry/types.ts
@@ -1,7 +1,9 @@
+import { ImmutableReferences, LinkReferences } from '@sphinx-labs/contracts'
 import {
   FoundryBroadcastTransaction,
   FoundryDryRunTransaction,
 } from '@sphinx-labs/core'
+import { ethers } from 'ethers'
 
 export type FoundryTransactionReceipt = {
   transactionHash: string
@@ -73,4 +75,28 @@ export type FoundryToml = {
       key: string
     }
   }
+}
+
+export type BuildInfoCache = {
+  _format: 'sphinx-build-info-cache-1'
+  entries: Record<string, BuildInfoCacheEntry>
+}
+
+/**
+ * @field contracts An array where each element corresponds to a contract in the
+ * `BuildInfo.output.contracts` object. We use this array to match collected contract init code and
+ * runtime bytecode to the corresponding artifact.
+ */
+export type BuildInfoCacheEntry = {
+  name: string
+  time: number
+  contracts: Array<{
+    fullyQualifiedName: string
+    bytecode: string
+    deployedBytecode: string
+    linkReferences: LinkReferences
+    deployedLinkReferences: LinkReferences
+    immutableReferences: ImmutableReferences
+    constructorFragment?: ethers.ConstructorFragment
+  }>
 }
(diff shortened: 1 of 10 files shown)


packages/plugins/src/hardhat/simulate.ts
```diff
@@ -67,6 +67,11 @@ export type simulateDeploymentSubtaskArgs = {
  * `spawn` is prone to input size limits, and EthersJS can't ABI encode extremely large amounts
  * of data (i.e. it'll fail for a Merkle tree that contains 250 contract deployments, where the
  * contract is near the maximum size limit).
+ *
+ * It's important to use a simulation to check that the Merkle leaf gas fields are sufficiently high.
+ * If a Merkle leaf gas field is too low, the simulation will fail because the associated `EXECUTE`
+ * action should fail in the simulation. The simulation must fork the corresponding live network
+ * because the user's transactions may interact with contracts that already exist on-chain.
```
sam-goldman, line 74: This is another [rider](https://linear.app/chugsplash/issue/CHU-594/document-why-we-need-to-use-a-hardhat-node-during-our-simulation)

packages/plugins/src/foundry/types.ts
```diff
@@ -74,3 +76,27 @@ export type FoundryToml = {
...
+ * `BuildInfo.output.contracts` object. We use this array to match collected contract init code and
+ * runtime bytecode to the corresponding artifact.
+ */
+export type BuildInfoCacheEntry = {
+  name: string
+  time: number
+  contracts: Array<{
+    fullyQualifiedName: string
+    bytecode: string
+    deployedBytecode: string
```
RPate97, line 96: We're adding a lot of information to the cache file now (this PR should roughly double its size). It's worth noting that if we increase the size of this file too much, we may encounter issues similar to those we encountered reading extremely large build-info files. 

If you don't remember, we are limited in the size of JSON files we can read using standard JSON.parse() by the maximum string length in NodeJS, which is ~512mb. If our cache exceeds this size, we would have to implement logic to use `stream-json` to read this file.

We don't need to address this issue right now since the cache would only exceed that limit in pretty extreme situations. I am just calling this out as a potential issue with adding more data to the cache file.

RPate97 (approved):
Looks good

packages/plugins/src/foundry/types.ts
sam-goldman: There's already a [ticket](https://linear.app/chugsplash/issue/CHU-392/reduce-the-size-of-the-sphinx-cachejson-file) for that in the backlog, which I just put back in Triage in case we want to handle this. I'm not sure how big the cache file can get during standard usage, so I think it's reasonable to investigate sometime soon

PR #1559, opened 2024-03-13
feat(pg): Supply more cost estimate data in proposal request

RPate97:
## Purpose
This PR adds additional gas estimate-related data to the ProposalRequest object. This data gives us more flexibility to implement different logic for calculating deployment costs for different networks with different fee formulas or custom RPC methods for calculating fees. 

Note that this PR does not resolve any known cases where we do not accurately estimate deployment costs (optimism, blast, polygon zkevm). I think the best option for handling those cases is to provide more information to the website's backend and then implement the necessary cost calculation logic on the website instead of within the plugin. The main reason why I think this is better is b/c we know of at least one case where a custom RPC method is required to calculate the cost estimates (polygon zkevm) properly. The RPC method is optional for node providers (like other trace or debug methods), so we can't guarantee the user's provider will include it.

### DIFF
--- a/packages/plugins/src/foundry/utils/index.ts
+++ b/packages/plugins/src/foundry/utils/index.ts
@@ -16,7 +16,7 @@ import { spawnSync } from 'child_process'
 import {
   BuildInfo,
   CompilerOutputContracts,
-  SphinxTransactionReceipt,
+  SphinxTransactionResponse,
 } from '@sphinx-labs/core/dist/languages/solidity/types'
 import {
   decodeDeterministicDeploymentProxyData,
@@ -55,8 +55,11 @@ import { streamObject } from 'stream-json/streamers/StreamObject'
 import { streamValues } from 'stream-json/streamers/StreamValues'
 import {
   ExecutionMode,
+  NetworkGasEstimate,
   ParsedContractDeployment,
   SphinxJsonRpcProvider,
+  EstimateGasTransactionData,
+  TransactionEstimatedGas,
 } from '@sphinx-labs/core'
 import ora from 'ora'
 import {
@@ -82,7 +85,7 @@ import {
   FoundrySingleChainDryRun,
   FoundryToml,
 } from '../types'
-import { simulate } from '../../hardhat/simulate'
+import { SimulationTransactions, simulate } from '../../hardhat/simulate'
 import { GetNetworkGasEstimate } from '../../cli/types'
 import { BuildInfoTemplate, trimObjectToType } from './trim'
 import { assertValidNodeVersion } from '../../cli/utils'
@@ -1063,6 +1066,21 @@ export const convertLibraryFormat = (
     return `${filePath}:${contractName}=${ethers.getAddress(address)}`
   })
 }
+
+const toSphinxTransactionEstimatedGas = (
+  response: SphinxTransactionResponse
+): EstimateGasTransactionData => {
+  return {
+    to: response.to,
+    from: response.from,
+    data: response.data,
+    gasLimit: response.gasLimit.toString(),
+    gasPrice: response.gasPrice.toString(),
+    value: response.value.toString(),
+    chainId: response.chainId.toString(),
+  }
+}
+
 /**
  * Estimates the gas used by a deployment on a single network. Includes a buffer of 30% to account
  * for variations between the local simulation and the production environment. Also adjusts the
@@ -1070,9 +1088,12 @@ export const convertLibraryFormat = (
  * forks.
  */
 export const getEstimatedGas = async (
-  receipts: Array<SphinxTransactionReceipt>,
+  transactions: SimulationTransactions,
   provider: SphinxJsonRpcProvider
-): Promise<string> => {
+): Promise<{
+  estimatedGas: string
+  transactionsWithGasEstimates: Array<TransactionEstimatedGas>
+}> => {
   // Estimate the minimum gas limit. On Ethereum, this will be 21k. (Technically, since
   // `eth_estimateGas` generally overestimates the gas used, it will be slightly greater than 21k.
   // It was 21001 during development). On Arbitrum and perhaps other L2s, the minimum gas limit will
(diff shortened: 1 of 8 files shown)


packages/plugins/src/hardhat/simulate.ts
```diff
@@ -249,6 +258,40 @@ export const setupPresimulationState = async (
...
+  receipts: Array<SphinxTransactionReceipt>,
+  provider: HardhatEthersProvider
+): Promise<SimulationTransactions> => {
+  // Since the size of receipts array is unbounded, we use pLimit to reduce the number of simultaneous calls.
+  // This reduces the chance of us triggering a rate limit in the RPC provider.
+  const transactions = await Promise.all(
+    receipts.map(async (receipt) => {
+      const response = await limit(async () =>
+        provider.getTransaction(receipt.hash)
+      )
```
RPate97, line 276: See the comment, we should probably try to use plimit whenever we have situations where we're using `Promise.all()` over a potentially large number of rpc provider calls to try to avoid getting rate limited if possible.

packages/plugins/src/foundry/utils/index.ts
```diff
@@ -1086,11 +1107,18 @@ export const getEstimatedGas = async (
...
-    .map((gasUsed) => Math.round(gasUsed * 1.3))
-    .map((gasWithBuffer) => {
+  const transactionsWithGasEstimates = transactions
+    .map((transaction) => {
+      return {
+        transaction: toSphinxTransactionEstimatedGas(transaction.response),
+        estimatedGas: Math.round(Number(transaction.receipt.gasUsed) * 1.3),
+        estimatedBlobGas: Math.round(
+          Number(transaction.receipt.blobGasUsed ?? 0) * 1.3
+        ),
```
RPate97, line 1117: I've included a 30% buffer on the blob gas for consistency. Let me know if you think there's a reason to do something else here.

packages/core/src/languages/solidity/types.ts
```diff
@@ -227,6 +227,7 @@ export type SphinxTransactionReceipt = {
   from: string
   gasPrice: string
   gasUsed: string
+  blobGasUsed: string | null
```
sam-goldman, line 230: The `SphinxTransactionReceipt` is a user-facing type because we include it in our deployment artifacts. Could you update `deployment-artifacts.md`?

sam-goldman: Also, can you update the JSDoc of this type? Specifically, what is the `blobGasUsed`, and when is it included?

packages/plugins/src/hardhat/simulate.ts
```diff
@@ -29,8 +29,19 @@ import {
...
   fetchExecutionTransactionReceipts,
   convertEthersTransactionReceipt,
+  convertEthersTransactionResponse,
+  SphinxTransactionResponse,
 } from '@sphinx-labs/core'
 import { ethers } from 'ethers'
+import { HardhatEthersProvider } from '@nomicfoundation/hardhat-ethers/internal/hardhat-ethers-provider'
+import pLimit from 'p-limit'
+
+const limit = pLimit(5)
```
sam-goldman, line 39: Is there a reason this is defined globally? Seems like it'd make more sense as a local variable in `fetchTransactionResponses`

packages/plugins/src/hardhat/simulate.ts
```diff
@@ -249,6 +260,39 @@ export const setupPresimulationState = async (
...
+ * use this information to calculate deployment cost estimates on the website.
+ */
+export const fetchTransactionResponses = async (
+  receipts: Array<SphinxTransactionReceipt>,
+  provider: HardhatEthersProvider
+): Promise<SimulationTransactions> => {
+  const chainId = (await provider.getNetwork()).chainId
+
+  // Since the size of receipts array is unbounded, we use pLimit to reduce the number of simultaneous calls.
+  // This reduces the chance of us triggering a rate limit in the RPC provider.
```
sam-goldman, line 275: I'm not sure if these requests are actually sent to the remote node b/c the transactions weren't actually executed there. Instead, this may be handled by Hardhat's node, in which case it's not necessary to use p-limit. I think we should keep using it to be safe, but just wanted to point it out

packages/plugins/src/hardhat/simulate.ts
```diff
@@ -249,6 +260,39 @@ export const setupPresimulationState = async (
   return signer
 }
 
+/**
+ * Fetches the transaction response objects for all of the transactions in the simulation. There's some additional
+ * data in the responses that is not in the receipts (data, value, etc) which is not available in the receipts. We
+ * use this information to calculate deployment cost estimates on the website.
+ */
+export const fetchTransactionResponses = async (
```
sam-goldman, line 268: Does this function actually return the blob gas used given that the transactions are executed on a Hardhat node?

packages/plugins/src/foundry/utils/index.ts
```diff
@@ -1063,16 +1066,35 @@ export const convertLibraryFormat = (
...
+
 /**
  * Estimates the gas used by a deployment on a single network. Includes a buffer of 30% to account
  * for variations between the local simulation and the production environment. Also adjusts the
  * minimum gas limit on networks like Arbitrum to include the L1 gas fee, which isn't captured on
  * forks.
  */
 export const getEstimatedGas = async (
-  receipts: Array<SphinxTransactionReceipt>,
+  transactions: SimulationTransactions,
```
sam-goldman, line 1091: Generally, it'd be helpful to add unit tests for functions like this that contain some nontrivial logic. It's hard to feel confident as a reviewer that this is bug free, even if I attempt to mentally walk through the code. It's fine not to add unit tests for this function, but generally, I think this'd help reduce the chance of bugs and make reviews easier

sam-goldman (commented):
Looks good, just had some comments/questions

sam-goldman:
> Note that this PR does not resolve any known cases where we do not accurately estimate deployment costs (optimism, blast, polygon zkevm). I think the best option for handling those cases is to provide more information to the website's backend and then implement the necessary cost calculation logic on the website instead of within the plugin. The main reason why I think this is better is b/c we know of at least one case where a custom RPC method is required to calculate the cost estimates (polygon zkevm) properly. The RPC method is optional for node providers (like other trace or debug methods), so we can't guarantee the user's provider will include it.

This is helpful context, thx

packages/plugins/src/hardhat/simulate.ts
RPate97: I'm not entirely sure, I was planning to test this out once the upgrade went live on Optimism (which just happened today).

RPate97: Yeah, unfortunately, the blob gas is not calculated during the simulation, so I'm just going to remove all the changes related to it.

sam-goldman (approved):
Looks good

PR #1563, opened 2024-03-15
fix(pg): make contract artifact selection more robust

sam-goldman:
Review [this PR](https://github.com/sphinx-labs/sphinx/pull/1557) first

Previously, the `readContractArtifact` function didn't check for a bytecode match, which made it possible for us to select the wrong contract artifact. This PR changes the `readContractArtifact` function to `findContractArtifact`, which checks for an exact bytecode match, and returns `undefined` if it isn't able to find one.

Changing `readContractArtifact` required me to change a decent amount of logic that depended on this function, so it may be easiest to review this PR in more than one phase.

This PR:
* Updates the functions that used `readContractArtifact`, particularly `makeGetConfigArtifacts` and `assertNoLinkedLibraries`
* Updates the `FoundryDeploymentInfo` to include a `scriptDeployedCode` field since this is necessary for `assertNoLinkedLibraries` to find the script's artifact.
* Implements an exact match strategy for the deployed bytecode of a contract, which is also necessary in `assertNoLinkedLibraries`, since the script's actual init code isn't available. (Foundry doesn't provide a way to get the init code of the script after linked libraries have been injected).

In the next PR, we'll use the exact match strategy on the script's deployed bytecode to implement this ticket: [Improve error message when user calls a non-existent function on their script](https://linear.app/chugsplash/issue/CHU-655/improve-error-message-when-user-calls-a-non-existent-function-on-their)


packages/plugins/src/foundry/utils/index.ts
```diff
@@ -310,84 +318,39 @@ export const compile = (silent: boolean, force: boolean): void => {
...
   }
 }
 
-/**
- * Returns an array of the most recent fully qualified names for the given `sourceName`. This
- * function searches the build info cache for the most recent build info file that contains a fully
- * qualified name starting with `sourceName`. Returns an empty array if there is no such fully
- * qualified name in any of the cached build info files.
- */
-const findFullyQualifiedNames = (
```
sam-goldman, line 358: We previously used this function in `assertNoLinkedLibraries`, but I removed it in favor of `findContractArtifact`, which is more reliable since it uses a bytecode match

packages/plugins/src/foundry/utils/index.ts
```diff
@@ -448,6 +420,14 @@ export const makeGetConfigArtifacts = (
...
         time: statSync(path.join(buildInfoPath, fileName)).mtime.getTime(),
       }))
+      // Sort the build info files according to their `time` field, which equals the `mtime`, i.e.
+      // the time that the build info was last modified. We'll use this field to determine the order
+      // in which to search the cache for the build info file and contract artifact that correspond
+      // to each init code supplied to this function. It's possible for the user to modify their
+      // build info files, which will change the `mtime` of the file. However, this is acceptable
+      // because we search for the build info file and contract artifact using an exact match on the
+      // init code. This means the worst case is that we select a valid yet outdated build info and
+      // artifact.
```
sam-goldman, line 430: Closes this ticket: [Document why it's safe to use mtime in makeGetConfigArtifacts](https://linear.app/chugsplash/issue/CHU-751/document-why-its-safe-to-use-mtime-in-makegetconfigartifacts)

sam-goldman:
> There's one thing I want to note, but I don't think it needs to block this PR. It seems pretty redundant to have this logic in addition to the existing logic that links most of the user's contracts to their artifacts using the init code. In principle, it seems like we could just handle everything using the deployed bytecode, which would eliminate a lot of complexity.

Agreed, made a [ticket](https://linear.app/chugsplash/issue/CHU-839/consider-removing-init-code-check-in-favor-of-always-using-deployed) for that

PR #1564, opened 2024-03-16
fix(pg): Avoid recompiling every single time we invoke the users script

RPate97:
## Purpose
Fixes an issue reported by Sablier where the `Validating Networks...` step is taking an extremely long time to run.

## Context
Foundry recently implemented a change where the cache is disregarded if the build info file is requested. This causes a UX issue where if you have build_info = true in your foundry.toml file then the entire project is recompiled every single time you run a script/test/etc. This is obviously a terrible experience. 
https://github.com/foundry-rs/foundry/issues/7379

This causes our plugin to run extremely slow because the users entire project is recompiled every time we call into their script. 

This PR makes the following changes to temporarily address the issue:
1. Remove `build_info = true` from our recommended `foundry.toml` settings 
2. When we call `compile` at the beginning of a script we compile with the `--build-info` flag since the rest of our logic still needs it. Also because we have to recompile the entire project anyway, I've added in the `--force` flag because that makes our later logic to fetch the build info file faster.
3. Whenever we call into the users script in the rest of the plugin, I've added an environment variable `FOUNDRY_BUILD_INFO: 'false'`. This prevents the users entire project from being rebuilt whenever we call into the script if the user has `build_info=true` in their `foundry.toml` file. 

## Note
Obviously, this is not the ideal solution. This PR aims to ensure the plugin continues to be usable in projects where compilation takes a long time, such as Sablier. We should further discuss the ideal solution which may involve implementing logic to use the cache when compiling with `build_info=true` in Foundry if that is reasonable to do.

### DIFF
--- a/.changeset/moody-trees-sing.md
+++ b/.changeset/moody-trees-sing.md
@@ -1,0 +1,5 @@
+---
+'@sphinx-labs/plugins': patch
+---
+
+Avoid recompiling every single time we invoke the users script
--- a/docs/cli-existing-project.md
+++ b/docs/cli-existing-project.md
@@ -189,7 +189,6 @@ ALCHEMY_API_KEY=<your_api_key>
 Update your `foundry.toml` file to include a few settings required by Sphinx. We recommend putting them in `[profile.default]`.
 
 ```
-build_info = true
 extra_output = ['storageLayout']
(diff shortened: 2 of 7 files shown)


sam-goldman (approved):
lgtm

PR #1565, opened 2024-03-16
fix(pg): make hardhat simulation more robust

sam-goldman:
Solves a few recurrent problems in the Hardhat simulation.

## Problems

This PR fixes:
* The "nonce too low" bug originally fixed by `hardhat_reset`.
* 0xBased's bug, where an infinite loop in `hardhat_reset` caused the Hardhat simulation to stall and send thousands of requests to the RPC provider.
* Kevin's bug, where the Hardhat simulation stalled due to a flaky RPC provider.

See the problems in [this ticket](https://linear.app/chugsplash/issue/CHU-768/make-hardhat-simulation-more-robust) for more context on 0xBased's and Kevin's bugs.

## Solution

I fixed these problems by introducing the `SimulationProvider`, which acts as a proxy for the `HardhatEthersProvider`. Every call sent to the `HardhatEthersProvider` is first routed through the `SimulationProvider`. This pattern allows us to implement logic that wraps every RPC call.

I moved our retry logic into the `SimulationProvider`. This means our retry logic now works on the granularity of individual RPC calls. This should speed up the simulation for slower or less reliable RPC endpoints because we aren't restarting the simulation every time a rate limit occurs, or when an iteration of the simulation times out. I previously wasn't able to use a free tier Alchemy API key reliably in the simulation, but now I can.

The `SimulationProvider` solves the three issues above by:
* Using `evm_snapshot` and `evm_revert` to fix the "nonce too low" bug. Specifically, we snapshot before making each call to the `HardhatEthersProvider`, and only revert if the call fails. Our retry logic ensures that we're able to re-execute the call until it succeeds.
* Removing `hardhat_reset` to prevent 0xBased's bug.
* Using `callWithTimeout` to prevent transactions from stalling indefinitely (Kevin's bug).

## Alternative approach

An alternative approach is to implement retries by wrapping the entire simulation in a `callWithTimeout`. However, this would've slowed down the simulation significantly because we would've been restarting the simulation on each iteration. To illustrate, consider what'd happen if the user has a flaky RPC endpoint and is executing a decently large deployment. If they get rate limited on the first iteration, they'll probably get rate limited on subsequent iterations because they're making the same number of requests each time. In the worst case, the rate limits would occur near the end of the execution process each time. If each iteration takes 5 minutes, and we allow 5 iterations, that's 25 minutes.

## Steps to reproduce

All three of these issues are difficult to consistently reproduce. I spent a long time trying to make them reproducible in case we need to revisit them in the future.

[Here](https://github.com/sphinx-labs/sphinx/blob/sg/hardhat-reset-loop/steps.md) are the steps to reproduce 0xBased's bug, and [here](https://github.com/sphinx-labs/sphinx/blob/sg/simulation-nonce-bug/steps.md) are the steps to reproduce the 'nonce too low' bug.

I eventually gave up on making steps to reproduce Kevin's bug because it only seems to happen when the RPC endpoint is degraded, which isn't in my control. The best I've got is the description in [the ticket](https://linear.app/chugsplash/issue/CHU-768/make-hardhat-simulation-more-robust) for this issue, which includes a Loom video.


packages/plugins/src/hardhat/simulate.ts
```diff
@@ -511,6 +457,124 @@ export const handleSimulationSuccess = async (
...
+        // More info on the nonce error is in this pull request description:
+        // https://github.com/sphinx-labs/sphinx/pull/1565.
+        const snapshotIdPromise = target.send('evm_snapshot', [])
+
+        // Call the method on the EthersJS provider.
+        const initialResult = target[prop](...args)
+        // Check if the call is synchronous. If so, return it.
+        if (!(initialResult instanceof Promise)) {
+          return initialResult
+        }
```
sam-goldman, line 487: You may wonder if returning early before awaiting the `snapshotIdPromise` is problematic. The answer is that I don't know. I'm going to make a ticket to investigate it. I don't think it's a big enough concern to block this PR

packages/core/src/utils.ts
```diff
@@ -290,9 +290,9 @@ export const callWithTimeout = async <T>(
     timeoutHandle = setTimeout(() => reject(new Error(errorMessage)), timeout)
   })
 
-  return Promise.race([promise, timeoutPromise]).then((result) => {
+  // Call `clearTimeout` as soon as either of the promises resolve or reject.
+  return Promise.race([promise, timeoutPromise]).finally(() => {
     clearTimeout(timeoutHandle)
-    return result
```
sam-goldman, line 295: Fixes a bug. See the new test case in `utils.spec.ts` for context

packages/plugins/src/hardhat/simulate.ts
```diff
@@ -25,24 +25,41 @@ import {
...
+/**
+ * @property maxAttempts - The maximum number of attempts that Sphinx will make for a single network
+ * request before throwing an error.
+ * @property timeout - The maximum number of time that Sphinx will wait for a single RPC request
+ * before timing out. An RPC request can stall if the RPC provider has degraded service, which would
+ * cause the simulation to stall indefinitely if we don't time out.
+ */
+export const simulationConstants = {
+  maxAttempts: 7,
+  timeout: 90_000,
```
RPate97, line 60: Nit: Since this timeout is for every single call, 90 seconds is a very long timeout period. I think something as low as 10 seconds would be acceptable.

packages/plugins/src/hardhat/simulate.ts
```diff
@@ -511,6 +455,129 @@ export const handleSimulationSuccess = async (
...
+
+              // We revert the Hardhat node state to ensure that there weren't any local state
+              // changes made by the failed RPC request. There should be no state changes because
+              // the call threw an error. This is a precaution against the "nonce too low" error,
+              // which is described in this pull request description:
+              // https://github.com/sphinx-labs/sphinx/pull/1565
+              const success = await target.send('evm_revert', [snapshotId])
+              if (!success) {
+                throw new InvariantError(`Failed to call 'evm_revert'.`)
+              }
```
RPate97, line 544: It's worth noting that this strategy could cause bugs if we're doing multiple state mutating requests simultaneously. Consider the following example:
1. Create a wallet with account A and submit a transaction with that account.
2. Create a wallet with account B and submit a second transaction without waiting for the previous transaction to get finalized. 

It's unclear what would happen if one of these transactions timed out. Depending on when the snapshot is taken for each transaction, you could potentially end up with a mixed-up state. 

This doesn't justify changing how this PR is implemented because we don't do anything like the above situation in the simulation, so this limitation won't cause us problems. However, I do think it's worth noting and documenting that this pattern prevents us from safely doing multiple calls that mutate state at the same time even if doing those calls simultaneously is perfectly valid.

packages/plugins/test/mocha/simulate.spec.ts
```diff
@@ -225,3 +236,186 @@ describe('handleSimulationSuccess', () => {
     }
   })
 })
+
+describe('createHardhatEthersProviderProxy', () => {
```
RPate97, line 240: Nit: It's confusing to me to have tests in `simulate.spec.ts` that don't interact with live networks and run on all PRs. I understand why you included them here since these are logic tests related to the simulation. Still, it would be slightly better to keep the tests that run against live network RPC providers separated into their own file for the sake of organization. I would use a different name for this file, like `live-network.spec.ts`. I would then have a `simulation.spec.ts` file that only includes these new tests.

RPate97 (approved):
Looks good, I had a few minor comments. It's worth noting that this setup does kind of limit what we can safely do in the simulation, but I don't think that's a huge deal. See my comment for more info.

packages/plugins/src/hardhat/simulate.ts
sam-goldman: I reduced it to 60s, but I don't think we should set it much lower than that. Added a bit more context on this to the `simulationConstants` variable:
```
* We set this value to be
 * relatively high because the execution process may submit very large transactions (specifically
 * transactions with `EXECUTE` Merkle leaves), which can cause the RPC request to be slow.
```

packages/plugins/src/hardhat/simulate.ts
sam-goldman: Yep good catch. Documented that. Also, will make a ticket for fixing this bug in Hardhat, since I think that'd resolve this issue as well as [this potential issue](https://github.com/sphinx-labs/sphinx/pull/1565#discussion_r1527251897) that I pointed out

packages/plugins/test/mocha/simulate.spec.ts
sam-goldman: Will keep this unchanged for the sake of speed, but I'll make a ticket to consider re-organizing this test file. Fwiw, our test suite generally follows the pattern of using the same name for the test file and the file being tested. In this case, `createHardhatEthersProviderProxy` is in `simulate.ts`, which is why I put this test suite in `simulate.spec.ts`

PR #1570, opened 2024-03-19
fix(pg): fix simulation bugs

sam-goldman:
This PR fixes several issues in the simulation:
* 0xBased ran into a timeout error in his simulation ([ticket](https://linear.app/chugsplash/issue/CHU-805/increase-timeout-in-simulation)). I fixed this by increasing the timeout from 60 seconds to 2.5 mins.
* Any call from a signer wasn't being routed through our provider proxy, so those calls were particularly prone to rate limits. They're now routed through the proxy.
* The backoff period was denominated in milliseconds, when it should've been denominated in seconds (oof)
* An obscure bug that was made a rate limit possible in the proxy. More details in a comment

### DIFF
--- a/packages/plugins/src/hardhat/simulate.ts
+++ b/packages/plugins/src/hardhat/simulate.ts
@@ -55,11 +55,14 @@ export type SimulationTransactions = Array<{
  * before timing out. An RPC request can stall if the RPC provider has degraded service, which would
  * cause the simulation to stall indefinitely if we don't time out. We set this value to be
  * relatively high because the execution process may submit very large transactions (specifically
- * transactions with `EXECUTE` Merkle leaves), which can cause the RPC request to be slow.
+ * transactions with `EXECUTE` Merkle leaves), which can cause the RPC request to be slow. We also
+ * set it high because the Hardhat provider appears to make a few retries (in
+ * `hardhat/internal/core/providers/http.ts`), which can contribute to the duration of a single RPC
+ * call sent from our provider proxy.
  */
 export const simulationConstants = {
-  maxAttempts: 7,
-  timeout: 60_000,
+  maxAttempts: 10,
+  timeout: 150_000,
 }
 
 /**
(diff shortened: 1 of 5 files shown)


packages/plugins/src/hardhat/simulate.ts
```diff
@@ -482,26 +482,12 @@ export const handleSimulationSuccess = async (
...
-        // https://github.com/sphinx-labs/sphinx/pull/1565.
-        const snapshotIdPromise = target.send('evm_snapshot', [])
-
-        // Call the method on the EthersJS provider.
-        const initialResult = target[prop](...args)
-        // Check if the call is synchronous. If so, return it.
-        if (!(initialResult instanceof Promise)) {
-          return initialResult
+        // Return the result directly if the method isn't asynchronous.
+        if (!sphinxCoreUtils.isPublicAsyncMethod(ethersProvider, prop)) {
```
sam-goldman, line 489: We use this to check whether the property is an asynchronous method. Our strategy of queueing the `initialResult` and `snapshotIdPromise` didn't work because of how Hardhat handles promises. Specifically, this situation led to a rate limit:
1. We queue the initial snapshot ID promise. Note that we don't use await: `const snapshotIdPromise = target.send('evm_snapshot', [])`
2. We queue the proxy call promise. We also don't use await: `target[prop](...args)`
3. We await the snapshot ID promise: `await snapshotIdPromise`. Crucially, this `await` caused the promise from step 2 to be awaited too. This made it possible to hit a rate limit here.

I fixed this by removing the logic that queues these promises in favor of using `isPublicAsyncMethod`, which simplifies this logic

packages/plugins/test/mocha/simulate.spec.ts
```diff
@@ -310,14 +323,15 @@ describe('createHardhatEthersProviderProxy', () => {
...
-      2,
+    // Verify linear backoff timing
+    const expectedDuration = sumEvenNumbers(
       2,
       // We subtract one because we throw the error after the last attempt instead of waiting.
       simulationConstants.maxAttempts - 1
     )
     expect(timeSum).equals(expectedDuration)
+    // Check that the time is denominated in seconds
+    expect(timeSum % 1000).equals(0)
```
sam-goldman, line 334: Fixes the fact that we were sleeping in milliseconds

RPate97 (approved):
Looks good, I think the tests need updating

PR #1574, opened 2024-03-19
fix(pg): Do not force recompile during propose or deploy

RPate97:
## Purpose
Do not force recompile when running propose/deploy command.


sam-goldman (approved):
Looks good, but wanna remove the docs above the `compile` call that explain why we force recompile, since those aren't relevant anymore?

PR #1578, opened 2024-03-19
fix(core): Update Drip versions to handle funding requests

RPate97:
## Purpose
- Updates Drippie configuration so we can call the drip function more frequently (necessary when airdropping funds near the limit). This also makes us better handle situations where the cost of a deployment exceeds the drip size because we can withdraw funds faster. 
- Bumps the Drippie versions for all networks
- Changes the Rootstock decimals value back to 18 (their documentation on this is awful, but I've manually tested this is correct this time)


sam-goldman (approved):
Looks good

PR #1579, opened 2024-03-20
feat(pg): Add execute action to confirm funds airdropped to Safe

RPate97:
## Purpose
Adds in logic to include an action that checks the balance of the safe if the user requests funding for the Safe.

### DIFF
--- a/packages/contracts/contracts/foundry/SphinxUtils.sol
+++ b/packages/contracts/contracts/foundry/SphinxUtils.sol
@@ -1201,7 +1201,7 @@ contract SphinxUtils is SphinxConstants, StdUtils {
      */
     function makeGnosisSafeTransaction(
         Vm.AccountAccess memory _access
-    ) private pure returns (GnosisSafeTransaction memory) {
+    ) public pure returns (GnosisSafeTransaction memory) {
         if (_access.kind == VmSafe.AccountAccessKind.Create) {
             // `Create` transactions are executed by delegatecalling the `CreateCall`
             // contract from the Gnosis Safe.
@@ -1288,7 +1288,7 @@ contract SphinxUtils is SphinxConstants, StdUtils {
         address[] memory _owners,
         uint256 _threshold,
         address _safeAddress
-    ) external {
+    ) public {
         // Get the encoded data that'll be sent to the `MultiSend` contract to deploy and enable the
         // Sphinx Module in the Gnosis Safe.
         bytes memory multiSendData = getModuleInitializerMultiSendData();
@@ -1409,7 +1409,7 @@ contract SphinxUtils is SphinxConstants, StdUtils {
         ParsedAccountAccess[] memory _accountAccesses,
         address _scriptAddress,
         FoundryDeploymentInfo memory _deploymentInfo
-    ) private returns (uint256[] memory) {
+    ) public returns (uint256[] memory) {
         address safe = getGnosisSafeProxyAddress(_scriptAddress);
         address module = getSphinxModuleAddress(_scriptAddress);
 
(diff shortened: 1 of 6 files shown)


packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -1465,6 +1470,98 @@ contract SphinxUtils is SphinxConstants, StdUtils {
...
+     * are running into this case regularly, then we may want to consider addressing this scenario in a
+     * more effective way by creating a new module version that can do this check during the approval
+     * function.
+     *
+     * The execute check works by transferring the required funds from the Safe to itself. We expect that
+     * this transfer will fail if the Safe does not have enough funds. To protect against both #1 and #2
+     * we use the following formula to calculate the required funds:
+     * Funds required = max(actual funds spent by the safe during the deployment, funds requested for safe)
+     */
+    function addBalanceCheckAction(
```
RPate97, line 1500: See the comment

I chose to address scenario 2 as part of this PR mainly because I think giving the user an easy mechanism to fund their safety and encouraging them to transfer funds as part of the deployment, if necessary, is priming them to hit this case. So I felt it was important to go ahead and implement a solution that would provide a good safety net in a wider range of cases where funds are required for a deployment.

packages/contracts/contracts/foundry/SphinxPluginTypes.sol
```diff
@@ -115,6 +115,7 @@ struct FoundryDeploymentInfo {
     bytes encodedDeployedContractSizes;
     uint fundsRequestedForSafe;
     uint safeStartingBalance;
+    uint safeFundsUsed;
```
sam-goldman, line 118: Before I review this PR more thoroughly, there are a couple things that I think need to be addressed

First, my perspective on this feature is that we’re writing un-audited code that directly feeds into logic that’s executed in the user’s Gnosis Safe. In that light, I think we need to be very cautious about how we approach this feature, and treat it more like on-chain logic than off-chain logic. In my opinion, writing un-audited logic that adds an additional `EXECUTE` action actually *decreases* the security profile of Sphinx. However, I think this is outweighed by the fact that it prevents a bug in our backend from messing up a user’s deployment midway when they use `fundSafe`. Outside of this specific situation, I think the security tradeoff points the other direction, so I don’t think we should include the `EXECUTE` leaf for scenario 2. Also, including scenario 2 in this feature appears to add complexity to the `addBalanceCheckAction` function. I think we should simplify this function as much as possible due to its mission critical nature

My second opinion is that I think we should include the `EXECUTE` action in the preview. I don’t think hiding this from the user instills trust in Sphinx. If we don’t include it, users may reasonably wonder what else Sphinx is executing through their Gnosis Safe without their knowledge.

packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -1201,7 +1201,7 @@ contract SphinxUtils is SphinxConstants, StdUtils {
      */
     function makeGnosisSafeTransaction(
         Vm.AccountAccess memory _access
-    ) private pure returns (GnosisSafeTransaction memory) {
```
sam-goldman, line 1204: Do you think we should have an integration test for this feature? Your call. Just floating the idea

packages/core/src/preview.ts
```diff
@@ -1,7 +1,12 @@
 import { yellow, green, bold } from 'chalk'
-import { CREATE3_PROXY_INITCODE } from '@sphinx-labs/contracts'
```
sam-goldman, line 2: The preview should display something like this, right?
```
GnosisSafe<0x...>.call{ value: 0.1 ether }()
```

Did you decide not to include a preview element that says something like "Check balance of the Gnosis Safe"?

packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -1465,6 +1465,65 @@ contract SphinxUtils is SphinxConstants, StdUtils {
...
+     * Say there are already funds in the Safe, the user then proposes a script that requires those
+     * funds, the user executes a transaction via the Safe using a third party interface that reduces
+     * the balance of the Safe. We then attempt to execute the deployment and it fails because the Safe
+     * does not have enough funds.
+     *
+     * This is a specific case of the more general problem that if a deployment depends on some specific
+     * on chain state, the deployment may end up failing if that state changes in between the deployment
+     * being approved and it getting executed.
+     */
+    function addBalanceCheckAction(
```
sam-goldman, line 1485: Just want to point out that the process of transferring native gas tokens from the Gnosis Safe to itself will cause the Gnosis Safe to emit a `SafeReceived` event, which may appear in the standard Gnosis Safe UI or on block explorers. This may be confusing to the user because they’ll see two value transfers occur: the first, where we send the native gas tokens from our backend, and the second, where the Gnosis Safe transfers the funds to itself.

I don't think this is a big deal, but just pointing it out

packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -1465,6 +1465,65 @@ contract SphinxUtils is SphinxConstants, StdUtils {
         return gasEstimates;
     }
 
+    /**
+     * Handles adding an execute action that confirms the Safe has received the requested funding from
+     * our backend. We only include this check if the user requests funds from our backend. This check
+     * just protects the user from an error occurring in our backend which causes the funds to fail to
```
sam-goldman, line 1471: How likely is it that our backend will accidentally transfer funds to the user's Gnosis Safe more than once? I remember a note in the previous PR that mentions it's possible if we call `attemptDeployment` more than once

sam-goldman (commented):
This looks fantastic. Just had some minor comments/questions

packages/contracts/test/Sphinx.t.sol
```diff
@@ -165,6 +168,7 @@ contract Sphinx_Test_Airdrop_Funds is Test, Sphinx, SphinxTestUtils {
         });
 
         assertEq(deploymentInfo.fundsRequestedForSafe, 0.15 ether);
+        assertCorrectFundCheckAction(deploymentInfo, 0.15 ether);
```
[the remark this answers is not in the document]
sam-goldman: Not necessary to do this. Just pointing it out

packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -1465,6 +1465,65 @@ contract SphinxUtils is SphinxConstants, StdUtils {
...
+        }
+
+        ParsedAccountAccess memory checkFundsAccess = ParsedAccountAccess(
+            VmSafe.AccountAccess({
+                chainInfo: VmSafe.ChainInfo(0, _deploymentInfo.chainId),
+                kind: VmSafe.AccountAccessKind.Call,
+                account: _deploymentInfo.safeAddress,
+                accessor: _deploymentInfo.safeAddress,
+                initialized: true,
+                oldBalance: _deploymentInfo.safeStartingBalance +
```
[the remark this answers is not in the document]
RPate97: No, because this action will be executed after we've transferred the requested funds to the safe.

packages/core/src/preview.ts
RPate97: Yeah, we only include the call to transfer funds to the Safe in the preview.

packages/contracts/contracts/foundry/SphinxUtils.sol
RPate97: Very unlikely, I've already solved this case in the executor.

packages/contracts/contracts/foundry/SphinxUtils.sol
RPate97: Yeah, I thought about that. Since this functionality takes advantage of our existing systems for generating execute actions from AccountAccesses and because we already have a forge test to confirm this action will revert on-chain, I don't think an integration test would add any coverage.

packages/contracts/contracts/foundry/SphinxUtils.sol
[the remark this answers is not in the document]
sam-goldman: Ah makes sense. Could you document that?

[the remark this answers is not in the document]
sam-goldman: Gotcha, could you document that?

packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -1465,6 +1465,65 @@ contract SphinxUtils is SphinxConstants, StdUtils {
...
+                    _deploymentInfo.fundsRequestedForSafe,
+                newBalance: _deploymentInfo.safeStartingBalance +
+                    _deploymentInfo.fundsRequestedForSafe,
+                deployedCode: "",
+                value: _deploymentInfo.safeStartingBalance + _deploymentInfo.fundsRequestedForSafe,
```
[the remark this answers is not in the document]
sam-goldman: Could you be a little more specific about why we don't just transfer `fundsRequestedForSafe`? I think the rationale is that `safeStartingBalance + fundsRequestedForSafe` funds are available in the user's Gnosis Safe when the transactions are collected, which means we need to check for an equal amount on-chain, or else the deployment may be underfunded

sam-goldman (approved):
Looks great. Just a couple last minor comments

PR #1580, opened 2024-03-21
fix(pg): hardcode Merkle leaf gas field on Moonbeam networks

sam-goldman:
Hardcodes the Merkle leaf gas value on the Moon networks to be 10,896,000, which is the maximum Merkle leaf gas size that fits in a batch on these networks. (The max batch size is 12M because the block gas limit is 15M).

I tested that this Merkle leaf gas value works properly in two ways:

1. I used the DevOps Platform to successfully deploy `plugins/MyContracts.sol:MyLargeContract` with 50 SSTOREs in its constructor. This contract deployment used ~868k more gas than the DCAHub when deploying both via an EOA on Moonbase Alpha ([more info](https://linear.app/chugsplash/issue/CHU-829/consider-hard-coding-merkle-leaf-gas-on-moonbeam)). I checked a few things about the deployment executed via the DevOps Platform:
* The `gasUsed` of the `MyLargeContract` transaction was 10,282,038, which is significantly lower than the block gas limit.
* The Merkle leaf gas field overestimated the actual gas used in the `GnosisSafe.execTransactionFromModule` call by > 600k, which is plenty of wiggle room. Specifically, the Merkle leaf gas field was hardcoded to 10,896,000, and the total `gasUsed` in the deployment was 10,282,038, which means the subcall to the Gnosis Safe cost less than 10,282,038.

2. I used the DevOps Platform to successfully execute Kevin's deployment on Moonbeam. Caveat: I deployed his contracts via the default `CREATE2` factory instead of Mean Finance's `DeterministicFactory` out of convenience. This means the gas used by my deployment is a few tens of thousands of gas less than his deployment. However, this is fine because there was plenty of wiggle room in his deployment too: the `gasUsed` of the `DCAHubCompanion` transaction was 7,985,022.


packages/contracts/src/networks.ts
```diff
@@ -1,104 +1,27 @@
...
  * @returns
  */
 export const calculateActionLeafGasForMoonbeam = (
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   foundryGas: string,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   deployedContractSizes: DeployedContractSize[],
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   access: ParsedAccountAccess
 ): string => {
```
RPate97, line 20: In the interest of time, this is fine, but we should clean these up if they don't get used in the longer term fix for this.

RPate97 (approved):

RPate97 (approved):
Approved, pending fixes for any CI checks that fail

PR #1585, opened 2024-03-21
fix(pg): enable caching in simulation

sam-goldman:
Our simulation produced several hundred RPC calls because Hardhat's caching was disabled. This PR enables caching, reducing the number of RPC calls by roughly an order of magnitude (~400 -> 40). Expect this to significantly improve the reliabiliity of the simulation as well as its performance, especially when using free RPC providers.

The crucial line in Hardhat that determines the caching is `hardhat/internal/hardhat-network/jsonrpc/client.js:L167`.

You can check the RPC usage by following these steps:
1. Open `sphinx/node_modules/hardhat/internal/core/providers/http.js`

2. Add `const fs = require('fs');` at the top of the file

3. Add the following line to line 135:
```
fs.appendFileSync('logs.txt', `request: ${JSON.stringify(request)}\n`);
```

4. Add this to the beginning of Sphinx's `propose` function:
```
  if (existsSync('logs.txt')) {
    unlinkSync('logs.txt')
  }
  writeFileSync('logs.txt', '')
```

5. Propose a Forge script

6. View logs.txt, which is written to the root of the `plugins` directory.


packages/plugins/src/hardhat/simulate.ts
```diff
@@ -145,10 +148,24 @@ export const simulate = async (
...
+    // Subtracting the reorg size can lead to the following edge case:
+    // 1. User executes a transaction on the live network.
+    // 2. User calls Sphinx's Propose or Deploy command using a script that relies on the state that
+    //    resulted from the transaction in the previous step.
+    // 3. The collection process works correctly because Foundry uses the latest block number.
+    // 4. The simulation uses a block where the transaction doesn't exist yet, causing an error.
+    //
+    // This edge case will naturally resolve itself if the user continues to attempt to
+    // propose/deploy. This is because the corresponding block will eventually be included in the
+    // simulation after there have been enough block confirmations.
```
sam-goldman, line 165: Made a [ticket](https://linear.app/chugsplash/issue/CHU-837/resolve-edge-case-caused-by-using-an-older-block-number-in-simulation) to resolve this edge case. I think this situation is unlikely to happen in practice. The reorg size is pretty modest, e.g. 5 blocks on Ethereum, and 30 blocks on most other networks. I think a reorg size of 30 corresponds to 15 minutes on Rootstock, which perhaps the slowest network we support

packages/plugins/src/hardhat/simulate.ts
```diff
@@ -145,10 +148,27 @@ export const simulate = async (
...
+    //    resulted from the transaction in the previous step.
+    // 3. The collection process works correctly because Foundry uses the latest block number.
+    // 4. The simulation uses a block where the transaction doesn't exist yet, causing an error.
+    //
+    // This edge case is unlikely to happen in practice. The reorg size is pretty small. For
+    // example, it's 5 blocks on Ethereum, and 30 blocks on most other networks. A reorg size of 30
+    // blocks corresponds to 15 minutes on Rootstock, which is one of the slowest networks that
+    // Sphinx supports as of now. If the edge case occurs, it will naturally resolve itself if the
+    // user continues to attempt to propose/deploy. This is because the corresponding block will
+    // eventually be included in the simulation after there have been enough block confirmations.
```
sam-goldman: Made a [ticket](https://linear.app/chugsplash/issue/CHU-837/resolve-edge-case-caused-by-using-an-older-block-number-in-simulation) to resolve this edge case

packages/core/src/networks.ts
```diff
@@ -337,44 +337,3 @@ export const shouldUseHigherMaxGasLimit = (chainId: bigint) => {
     throw new Error(`Unsupported network id ${chainId}`)
   }
 }
-
-/**
- * The number of blocks that Hardhat rewinds when forking the given network. Rewinding the block
- * number protects against chain reorgs. Copied from Hardhat:
- * https://github.com/NomicFoundation/hardhat/blob/caa504fe0e53c183578f42d66f4740b8ec147051/packages/hardhat-core/src/internal/hardhat-network/provider/utils/reorgs-protection.ts
- */
-export const getLargestPossibleReorg = (chainIdStr: string): bigint => {
```
sam-goldman, line 346: I removed this function, which was copied from Hardhat, and instead import the identical function directly from Hardhat. We previously defined it ourselves because I was paranoid about the fact that this function exists in an `internal/` directory in their package. I was concerned that the location could change unpredictably. However, since we hardcode the `hardhat` version and are planning on continuing to do this when we implement the Sphinx binary, I don't think this is a concern. Importing it ensures that we use the correct reorg size, which is necessary to enable caching. (More context to come)

packages/plugins/src/hardhat/simulate.ts
```diff
@@ -145,10 +148,27 @@ export const simulate = async (
   }
 
   if ((await isLiveNetwork(provider)) || (await isFork(provider))) {
-    // Use the same block number as the Forge script that collected the user's transactions. This
-    // reduces the chance that the simulation throws an error or stalls, which can occur when using
-    // the most recent block number.
-    envVars['SPHINX_INTERNAL__BLOCK_NUMBER'] = networkConfig.blockNumber
+    // Use the block number from the Forge script minus the largest possible chain reorg size, which
```
sam-goldman, line 151: Here's the caching explanation

packages/plugins/src/hardhat/simulate.ts
```diff
@@ -145,10 +148,27 @@ export const simulate = async (
...
+    //    resulted from the transaction in the previous step.
+    // 3. The collection process works correctly because Foundry uses the latest block number.
+    // 4. The simulation uses a block where the transaction doesn't exist yet, causing an error.
+    //
+    // This edge case is unlikely to happen in practice because the reorg size is pretty small. For
+    // example, it's 5 blocks on Ethereum, and 30 blocks on most other networks. A reorg size of 30
+    // blocks corresponds to 15 minutes on Rootstock, which is one of the slowest networks that
+    // Sphinx supports as of now. If the edge case occurs, it will naturally resolve itself if the
+    // user continues to attempt to propose/deploy. This is because the corresponding block will
+    // eventually be included in the simulation after there have been enough block confirmations.
```
sam-goldman, line 168: Made a [ticket](https://linear.app/chugsplash/issue/CHU-837/resolve-edge-case-caused-by-using-an-older-block-number-in-simulation) to resolve this edge case

sam-goldman:
@RPate97 requesting your review on this since it's a quick solution that'll improve the reliability of the Hardhat simulation significantly. Not a long term solution ofc

sam-goldman:
> @RPate97 requesting your review on this since it's a quick solution that'll improve the reliability of the Hardhat simulation significantly. Not a long term solution ofc

Actually, scratch that. Canceled the review again. I already wrote the logic that determines the max batch size without the Hardhat simulation. I think that's sufficient to remove the simulation, even though the on-chain simulation doesn't exist yet

sam-goldman:
@RPate97 Fyi this is ready for review

RPate97 (approved):

PR #1608, opened 2024-04-01
fix(pg): Update gas heuristics to support large contracts on Rootstock

sam-goldman:
Adds support for large contracts on Rootstock and its testnet, which both support EIP-2028.

To deploy Mean Finance's `DCAHub`, the IR compilation pipeline must be enabled. I'm going to ask them if this is acceptable. I settled for this solution because the alternative was to stretch our gas heuristics beyond what I considered a safe range.

I kept the Merkle leaf gas buffer unchanged because decreasing it to 7% would've caused the proposal test case for `MyLargeContract` to fail.

I actually increased the batch size heuristic in `estimateGasViaManagedService` and `estimateGasViaSigner` to multiply the estimated gas amount by 1.08 instead of 1.05. I increased this buffer because a buffer of 1.04 caused the `estimateGasViaManagedService` function to underestimate the gas in the Hardhat proposal simulation, leading to a `'SphinxModule: insufficient gas'` error. The fact that our batch estimation logic considered this a valid batch indicates that 1.04  underestimates the actual gas used. So, as a precautionary measure, I increased it to 1.08. This scenario occurred with a Merkle leaf buffer of 9%, a max batch size limit of 99%, and a `DCAHub` that was _not_ compiled via IR.

I used a max batch limit of 99%. This was necessary to deploy the `DeterministicFactory` at its `CREATE2` address. It's important for this contract to have a consistent address since it determines the address of other contracts, so I didn't want to use IR to compile it.

I tested that this works by deploying Mean Finance's `DCAHub` using my [`dca-v2-deployments` repo](https://github.com/sam-goldman/dca-v2-deployments) with the Deploy CLI command. I also checked that the Hardhat simulation succeeds when proposing the entire script.


packages/core/src/utils.ts
```diff
@@ -1344,12 +1345,20 @@ export const getMaxGasLimit = (
...
+  }
   // On Scroll and Scroll Sepolia, set the max gas limit to be 80% of the block gas limit. We set a
   // higher value for these networks because their block gas limit is low (10 million), which means
   // a lower max gas limit could cause large contract deployments to be unexecutable. Transactions
   // that use ~8.5M gas were executed quickly on Scroll Sepolia, so an 80% limit shouldn't
   // meaningfully impact execution speed.
-  if (shouldUseHigherMaxGasLimit(chainId)) {
+  else if (shouldUseHigherMaxGasLimit(chainId)) {
     return (blockGasLimit * BigInt(8)) / BigInt(10)
   }
```
RPate97, line 1363: I would prefer not to break the pattern of having all chain-specific settings defined in the `SPHINX_NETWORKS` array. Could you please refactor this so it doesn't rely on the if statement that checks for rootstock and rootstock testnet?

I recommend the following general approach:
1. Remove the `useHigherMaxGasLimit` field from the `SPHINX_NETWORKS` array.
2. Introduce a new field `maxBlockGasLimit` multiple, an object with two fields, numerator and denominator. 
3. In this function, use the values from the `maxBlockGasLimit` for the network.
4. Make sure Rootstock and Rootstock testnet have 99% multiples. For any networks where `useHigherMaxGasLimit` was true, we should have 80% multiples. 

This approach keeps all configurations inside the `SPHINX_NETWORKS` array and introduces a pattern covering all networks requiring a different maximum batch size.

packages/contracts/src/networks.ts
```diff
@@ -1004,7 +1004,7 @@ export const SPHINX_NETWORKS: Array<SupportedNetwork> = [
     queryFilterBlockLimit: 2000,
     actionGasLimitBuffer: true,
     useHigherMaxGasLimit: true,
-    eip2028: false,
+    eip2028: true,
```
RPate97, line 1007: Since this field is no longer used for any network, I would just remove it. We should also update the `getGasPerCalldata` function so that it no longer includes logic related to this.

RPate97 (changes requested):
Looks pretty good, but this has logic that breaks the `SPHINX_NETWORKS` configuration pattern.

RPate97:
I also tested this out by linking this branch locally in the mean finance repo. I was not able to deploy the DCA hub with these changes. I got the following error:

Screenshot 2024-04-06 at 11 33 24 AM

Note that the two numbers before the error message are the gas values that caused the error to be thrown. The first is the action leaf gas for deploying the DCA hub, and the second is the maximum batch size on the network. 

Please double-check that the changes in this PR work as they are currently. I'm sure you were able to deploy the DCAHub, but possibly all of the changes you used did not get included in this PR.

packages/core/src/utils.ts
sam-goldman: I think simpler approach is to just use the block gas limit to determine the max batch size. This way, we don't need to manually enter a field (or fields) in the `SPHINX_NETWORKS` array for each network.

We could change the `getMaxGasLimit` function to be something like:
- If blockGasLimit <= 7 million: use a 99% multiple
- Else if blockGasLimit < 15 million: use an 80% multiple
- Else: use a 50% multiple

It seems like we're essentially following this pattern already

packages/contracts/src/networks.ts
sam-goldman: Slight preference for keeping it in case there's a network in the future that doesn't support EIP-2028. Not opinionated though

packages/core/src/utils.ts
sam-goldman: This pattern also handles the case where a network changes their block gas limit at some point in the future. (I haven't seen this happen before, but I think it's possible)

sam-goldman:
> I also tested this out by linking this branch locally in the mean finance repo. I was not able to deploy the DCA hub with these changes. I got the following error:
> 
> Screenshot 2024-04-06 at 11 33 24 AM
> Note that the two numbers before the error message are the gas values that caused the error to be thrown. The first is the action leaf gas for deploying the DCA hub, and the second is the maximum batch size on the network.
> 
> Please double-check that the changes in this PR work as they are currently. I'm sure you were able to deploy the DCAHub, but possibly all of the changes you used did not get included in this PR.

I wasn't able to reproduce this. Could you check that you locally linked the `@sphinx-labs/contracts` package in `remappings.txt` and the `@sphinx-labs/plugins` package? Then, if you still run into the error, can you give me exact steps to reproduce it?

packages/core/src/utils.ts
RPate97: That's acceptable to me, I just don't want to have cases where we have logic hanging around that runs based on the target network chain id.

RPate97 (approved):
Looks good. I've tested this on the website.

PR #1609, opened 2024-04-01
feat(core): Improve RPC url setup for secret management

RPate97:
## Purpose
- Updates how RPC URL env variable names are stored in the SPHINX_NETWORKS array to support better using a secret manager
- Updates the names of some RPC URL env variables to be more consistent
- Updates the `plugins` package to use the same env variable names as the `core` package
- Adds scripts to the `core` and `plugins` packages to allow exporting RPC URLs from Infisical
- Changes a few dependencies that are currently installed via GitHub to be dev dependencies since they are not needed in production, and the installation process is not reliable when installing directly from GitHub

## Fetching RPC urls
You can now simply export RPC URLs from Infisical whenever you need them. For more information, see the [RPC URL notion doc](https://www.notion.so/RPC-Endpoints-0ec4ceb92d1d4406848d5030e88cfad6).


packages/plugins/package.json
```diff
@@ -68,6 +67,8 @@
     "hardhat": "2.20.1"
   },
   "devDependencies": {
+    "sphinx-forge-std": "https://github.com/foundry-rs/forge-std.git#v1.7.1",
+    "sphinx-solmate": "npm:solmate@^6.7.0",
```
RPate97, line 71: I just noticed Solmate doesn't need to be a prod dependency, so I moved it over, too.

packages/contracts/src/networks.ts
```diff
@@ -103,8 +103,8 @@ export const SPHINX_NETWORKS: Array<SupportedNetwork> = [
     name: 'ethereum',
     displayName: 'Ethereum',
     chainId: BigInt(1),
-    rpcUrl: () =>
-      `https://eth-mainnet.g.alchemy.com/v2/${process.env.ALCHEMY_API_KEY}`,
+    rpcUrl: () => process.env.ETH_MAINNET_URL!,
+    rpcUrlId: 'ETH_MAINNET_URL',
```
RPate97, line 107: Switching to using a separate env variable for every RPC URL adds a bit of repetition to our .env files but makes the integration with our secret manager much simpler. I think this is a fine tradeoff since we can easily sync our RPC URLs to the mono repo using `yarn sync:rpc`.

sam-goldman (approved):
Looks good

PR #1612, opened 2024-04-04
fix(pg): throw an error if contracts are above the size limit

sam-goldman:
Multiple users have run into this issue, most recently 0xBased today. See [this ticket](https://linear.app/chugsplash/issue/CHU-389/throw-a-helpful-error-if-contracts-are-too-large) for context.

### DIFF
--- a/packages/plugins/src/cli/deploy.ts
+++ b/packages/plugins/src/cli/deploy.ts
@@ -52,6 +52,7 @@ import {
   getSphinxConfigFromScript,
   parseScriptFunctionCalldata,
   readInterface,
+  assertContractSizeLimitNotExceeded,
   writeSystemContracts,
 } from '../foundry/utils'
 import { getFoundryToml } from '../foundry/options'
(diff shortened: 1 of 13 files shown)


packages/plugins/src/cli/deploy.ts
```diff
@@ -304,6 +305,11 @@ export const deploy = async (
     initCodeWithArgsArray
   )
 
+  assertContractSizeLimitNotExceeded(
+    deploymentInfo.accountAccesses,
+    configArtifacts
+  )
+
```
RPate97, line 312: Nit: This is somewhat outside the scope of this issue, but it would be nice if we just had a single function that does all checks like this (i.e., it includes the `assertNoLinkedLibraries` call below). That seems more maintainable to me.

RPate97 (approved):
Looks good, just left one minor nit.

packages/plugins/src/cli/deploy.ts
sam-goldman: We're going to remove `assertNoLinkedLibraries` when we support linked libraries, so I don't think it's worth refactoring this just yet

PR #1622, opened 2024-04-07
feat(pg): Add support for Darwinia Pangolin

RPate97:
## Purpose
Adds support for Darwinia Pangolin


sam-goldman (approved):
Looks good, core tests aren't passing though

PR #1634, opened 2024-04-09
fix(pg): Hardcode merkle leaf gas on Darwinia Pangolin

RPate97:
## Purpose
Uses a hardcode Merkle leaf gas on Darwinia Pangolin.

## Testing
I've tested this, and it resolves the issue Echo ran into. I've also tested that it works when deploying a contract around the maximum size limit (I deployed `MyContracts.sol:MyLargeContract` from the plugins package).


sam-goldman (approved):
Looks good. You could delete the `calculateActionLeafGasForMoonbeam` function

PR #1643, opened 2024-04-10
fix(pg): Lower hardcoded merkle leaf gas on Pangolin

RPate97:
## Purpose
Fixes an issue where the hardcoded Merkle leaf gas on Pangolin was greater than the maximum batch on Pangolin. This is a temporary solution. 

I think we should discuss and then implement a better solution that takes into account the variable batch size that is set using the `getMaxGasLimit` function.

### DIFF
--- a/packages/contracts/src/networks.ts
+++ b/packages/contracts/src/networks.ts
@@ -1178,6 +1178,6 @@ export const SPHINX_NETWORKS: Array<SupportedNetwork> = [
     legacyTx: false,
     actionGasLimitBuffer: false,
     eip2028: true,
-    hardcodedMerkleLeafGas: (14_000_000).toString(),
+    hardcodedMerkleLeafGas: (11200000).toString(),
   },
 ]
--- a/packages/core/src/networks.ts
+++ b/packages/core/src/networks.ts
@@ -140,6 +140,10 @@ export const calculateMerkleLeafGas = (chainId: bigint, foundryGas: string) => {
   const network = SPHINX_NETWORKS.find((n) => n.chainId === chainId)
 
   if (network?.hardcodedMerkleLeafGas) {
+    if (BigInt(foundryGas) > BigInt(network?.hardcodedMerkleLeafGas)) {
+      throw new Error('Transaction too large to be executed')
+    }
+
     return network.hardcodedMerkleLeafGas
   } else {
     return foundryGas
--- a/packages/core/src/utils.ts
+++ b/packages/core/src/utils.ts
@@ -1351,7 +1351,7 @@ export const getMaxGasLimit = (blockGasLimit: bigint): bigint => {
   // million will have a max batch size of 5.68 million (= 80% * 7.1 million), which is too low.
   if (blockGasLimit <= BigInt(8_500_000)) {
     return blockGasLimit
-  } else if (blockGasLimit <= BigInt(13_500_000)) {
+  } else if (blockGasLimit <= BigInt(20_000_000)) {
     return (blockGasLimit * BigInt(8)) / BigInt(10)
   }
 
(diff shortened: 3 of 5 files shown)


packages/contracts/src/networks.ts
```diff
@@ -1178,6 +1178,6 @@ export const SPHINX_NETWORKS: Array<SupportedNetwork> = [
     legacyTx: false,
     actionGasLimitBuffer: false,
     eip2028: true,
-    hardcodedMerkleLeafGas: (14_000_000).toString(),
+    hardcodedMerkleLeafGas: (11200000).toString(),
```
sam-goldman, line 1181: In what situation was the previous value too high? Weren't you able to deploy `MyLargeContract` with it?

packages/core/src/utils.ts
```diff
@@ -1351,7 +1351,7 @@ export const getMaxGasLimit = (blockGasLimit: bigint): bigint => {
   // million will have a max batch size of 5.68 million (= 80% * 7.1 million), which is too low.
   if (blockGasLimit <= BigInt(8_500_000)) {
     return blockGasLimit
-  } else if (blockGasLimit <= BigInt(13_500_000)) {
+  } else if (blockGasLimit <= BigInt(20_000_000)) {
```
sam-goldman, line 1354: Can I get some context on this? Or is this temporary?

packages/core/src/networks.ts
```diff
@@ -140,6 +142,12 @@ export const calculateMerkleLeafGas = (chainId: bigint, foundryGas: string) => {
   const network = SPHINX_NETWORKS.find((n) => n.chainId === chainId)
 
   if (network?.hardcodedMerkleLeafGas) {
+    if (BigInt(foundryGas) > BigInt(network?.hardcodedMerkleLeafGas)) {
+      throw new InvariantError(
+        'Transaction too large to be executed on Pangolin'
```
sam-goldman, line 147: I don't think this should be an `InvariantError` because this type of error is appended with "Should never happen". It seems like this error will be triggered if the user has a very large transaction

sam-goldman, line 147: The error message says Pangolin but it can be triggered on other networks, e.g. Moonbeam

sam-goldman (approved):
I need more context to understand why you made these changes. I don't think this'll break our logic though, so I approved it

packages/core/src/utils.ts
RPate97: This is temporary while we work out a more comprehensive solution.

packages/contracts/src/networks.ts
RPate97: This bug occurred because of the following happened

First, in my PR to fix the original Pangolin issue, I set the hardcoded limit on Pangolin to 14 million. I also set the field `useHigherMaxGasLimit` to true on Pangoline. On Pangolin the block gas limit is 20 million. As a result, the maximum batch size on Pangolin was 20 million * 0.8 = 16 million. That is significantly higher than the hardcoded limit of 14 million, so everything thing works. 

Then your rootstock gas limit PR removed the `useHigherMaxGasLimit` field and replaced it with a function that dynamically sets the max batch size based on the block gas limit on the target network. If the target network block gas limit is greater than 13.5 million than the max batch size was set to 50% of the block gas limit. So on Pangolin the max batch size was 20 million * 0.5 = 10 million which is less than the hardcoded value of 14 million. As a result, it became impossible to propose any deployments on Pangolin b/c the hardcoded gas limit was always higher than the maximum batch size. 

It's worth noting that since the simulation tests are super flaky and I'm used to them failing, I just went ahead and did the release for the rootstock PR even though they were not passing (they failed due to another issue, not specifically this). So I also should have been more patient and retried the simulation tests before doing the release.

PR #1650, opened 2024-04-10
fix(pg): disable polygon zkevm action transaction batching

RPate97:
## Purpose
- Adds logic to allow disabling action batching during the execution process.
- Fixes an issue where artifact generation could fail for deployments that involved a large number of transactions. 
- Disables batching on the following networks: Polygon zkevm (and testnet), Linea (and testnet), Moonbeam, Moonriver, Moonbase Alpha, Scroll (and testnet), Rootstock (and testnet), Rari (and testnet), Darwinia Pangolin

### These networks were chosen because they fall into these categories:
- Is a ZK Rollup
- Is in the list of networks on [CHU-861](https://linear.app/chugsplash/issue/CHU-861/investigate-calldata-length-restrictions-on-certain-networks) where the calldata size is a problem
- Is a network where we use a hardcoded merkle leaf gas limit

### DIFF
--- a/packages/core/src/actions/execute.ts
+++ b/packages/core/src/actions/execute.ts
@@ -40,6 +40,7 @@ import {
 } from '../utils'
 import {
   implementsEIP2028,
+  isActionTransactionBatchingEnabled,
   shouldBufferExecuteActionsGasLimit,
 } from '../networks'
 import {
@@ -299,11 +300,15 @@ const findMaxBatchSize = (
     throw new Error(`Must enter at least one Merkle leaf.`)
   }
 
+  const maxBatchSize = isActionTransactionBatchingEnabled(chainId)
+    ? leaves.length
+    : 1
+
   // Start with the smallest batch size (1) and incrementally try larger batches. We don't start
   // with the largest possible batch size and incrementally try smaller batches because the gas
   // estimation logic ABI encodes the Merkle leaves, which can be very slow for large amounts of
   // data.
-  for (let i = 1; i <= leaves.length; i++) {
+  for (let i = 1; i <= maxBatchSize; i++) {
     if (
       !isExecutable(
         leaves.slice(0, i),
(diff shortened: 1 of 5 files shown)


packages/core/src/artifacts.ts
```diff
@@ -609,9 +609,10 @@ const makeExecutionArtifact = async (
   merkleRoot: string,
   provider: SphinxJsonRpcProvider
 ): Promise<ExecutionArtifact> => {
-  const ethersResponses = await Promise.all(
-    receipts.map((rcpt) => provider.getTransaction(rcpt.hash))
-  )
+  const ethersResponses: (ethers.TransactionResponse | null)[] = []
+  for (const rcpt of receipts) {
+    ethersResponses.push(await provider.getTransaction(rcpt.hash))
+  }
```
RPate97, line 615: Using `Promise.all` here causes issues if we have a lot of transactions in a deployment. I noticed this causing issues when testing Mean Finances deployment on Polygon ZKEvm

packages/core/src/actions/execute.ts
```diff
@@ -299,33 +300,37 @@ const findMaxBatchSize = (
...
-        // size.
-        return i - 1
+  if (isActionTransactionBatchingEnabled(chainId)) {
+    // Start with the smallest batch size (1) and incrementally try larger batches. We don't start
+    // with the largest possible batch size and incrementally try smaller batches because the gas
+    // estimation logic ABI encodes the Merkle leaves, which can be very slow for large amounts of
+    // data.
+    for (let i = 1; i <= leaves.length; i++) {
+      if (
+        !isExecutable(
```
sam-goldman, line 310: I think we still need to call `isExecutable` for all networks, including those that have `actionTransactionBatching=false`. Otherwise, it'll be possible for us to construct a batch containing a single leaf that's greater than the max batch size limit. This is particularly a concern on Rootstock

sam-goldman (commented):
Looks good, but i think you need to modify `findMaxBatchSize`

sam-goldman (approved):
Looks good

PR #1653, opened 2024-04-11
fix(pg): Bump Merkle leaf gas values on networks other than Rootstock

sam-goldman:
This change makes Rysk's protocol executable on Blast Sepolia. It also resolved 0xBased's issue, where the simulation reverted due to an underfunded transaction.

I didn't bump Rootstock's gas values because it's not necessary (explained in the inline docs) and because it'd prevent Mean Finance from deploying their system on this network.


packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -1170,10 +1170,16 @@ contract SphinxUtils is SphinxConstants {
...
+            // We chose to multiply the gas by 10-20% because multiplying it by a higher number
+            // could make a very large transaction unexecutable on-chain. Since this multiplier
+            // doesn't impact small transactions very much, we add a constant amount too. We use
+            // smaller buffers on Rootstock because gas costs are slightly lower on these networks
+            // compared to the EVM. Also, the block gas limit is significantly lower on Rootstock.
+            if (_deploymentInfo.chainId == 30 || _deploymentInfo.chainId == 31) {
+                gasEstimates[i] = 60_000 + ((startGas - finalGas) * 11) / 10;
+            } else {
+                gasEstimates[i] = 120_000 + ((startGas - finalGas) * 12) / 10;
+            }
```
RPate97, line 1182: I still really don't want to have code that implements special logic based on the chain id. I'm ok with this because I think we are going to remove the merkle leaf gas soon anyway. However, we really cannot perpetuate this pattern. It will not scale to more networks.

RPate97 (approved):

packages/contracts/contracts/foundry/SphinxUtils.sol
sam-goldman: Yeah I agree. I should've mentioned that I think this fine because we're planning to remove the Merkle leaf gas field very soon. I'm not planning on perpetuating this pattern in general

PR #1665, opened 2024-04-11
feat(pg): Include rollup stack information on all relevant networks

RPate97:
## Purpose
- Adds more information on the underlying rollup stack for relevant networks (mode, blast, etc)


packages/core/src/actions/types.ts
```diff
@@ -107,11 +107,10 @@ export type ContractInfo = {
 }
 
 export type EstimateGasTransactionData = {
-  to: string | null
+  to: string
```
RPate97, line 110: Just removed the null option here because it's not actually needed.

packages/plugins/src/foundry/utils/index.ts
```diff
@@ -1089,7 +1089,6 @@ const toSphinxTransactionEstimatedGas = (
     from: response.from,
     data: response.data,
     gasLimit: response.gasLimit.toString(),
-    gasPrice: response.gasPrice.toString(),
```
RPate97, line 1092: Removed the gas price here because it isn't useful for estimating fees on the website. If we need this on the website, we'll just populate it with the current gas price on the network.

PR #1666, opened 2024-04-11
feat(pg): Include rollup stack information on all relevant networks

RPate97:
## Purpose
Fills in rollup stack information for relevant networks. This is useful for making sure our cost estimates are accurate on the website.

### DIFF
--- a/packages/contracts/src/networks.ts
+++ b/packages/contracts/src/networks.ts
@@ -63,8 +63,8 @@ export const SPHINX_LOCAL_NETWORKS: Array<SupportedLocalNetwork> = [
 ]
 
 export type NetworkType = 'Testnet' | 'Mainnet' | 'Local'
-type RollupProvider = 'Conduit' | 'Caldera'
-type RollupType = 'OP Stack' | 'Arbitrum'
+type RollupProvider = 'Conduit' | 'Caldera' | 'None'
+type RollupType = 'OP Stack' | 'Arbitrum' | 'Polygon CDK'
 
 export const SPHINX_NETWORKS: Array<SupportedNetwork> = [
   {
@@ -156,6 +156,10 @@ export const SPHINX_NETWORKS: Array<SupportedNetwork> = [
     actionGasLimitBuffer: false,
     eip2028: true,
     actionTransactionBatching: true,
+    rollupStack: {
+      provider: 'None',
+      type: 'OP Stack',
+    },
   },
   {
     name: 'optimism_sepolia',
@@ -186,6 +190,10 @@ export const SPHINX_NETWORKS: Array<SupportedNetwork> = [
     actionGasLimitBuffer: false,
     eip2028: true,
     actionTransactionBatching: true,
+    rollupStack: {
+      provider: 'None',
+      type: 'OP Stack',
+    },
   },
   {
     name: 'arbitrum',
@@ -210,6 +218,10 @@ export const SPHINX_NETWORKS: Array<SupportedNetwork> = [
     actionGasLimitBuffer: false,
     eip2028: true,
     actionTransactionBatching: false,
+    rollupStack: {
+      provider: 'None',
+      type: 'Arbitrum',
+    },
   },
   {
     name: 'arbitrum_sepolia',
@@ -234,6 +246,10 @@ export const SPHINX_NETWORKS: Array<SupportedNetwork> = [
     actionGasLimitBuffer: false,
     eip2028: true,
     actionTransactionBatching: false,
+    rollupStack: {
+      provider: 'None',
+      type: 'Arbitrum',
+    },
   },
   {
     name: 'polygon',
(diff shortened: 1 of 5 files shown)


packages/contracts/src/networks.ts
```diff
@@ -63,8 +63,8 @@ export const SPHINX_LOCAL_NETWORKS: Array<SupportedLocalNetwork> = [
 ]
 
 export type NetworkType = 'Testnet' | 'Mainnet' | 'Local'
-type RollupProvider = 'Conduit' | 'Caldera'
-type RollupType = 'OP Stack' | 'Arbitrum'
+type RollupProvider = 'Conduit' | 'Caldera' | 'None'
```
sam-goldman, line 66: nit: the `null` type may be preferable instead of a string called "None"

packages/contracts/src/networks.ts
```diff
@@ -63,8 +63,8 @@ export const SPHINX_LOCAL_NETWORKS: Array<SupportedLocalNetwork> = [
 ]
```
sam-goldman, line 64: Technically out of scope of this PR, but does it matter that Rari is an Arbitrum L3, not an L2? ([Source](https://rarichain.org/))

I'm not sure if this impacts the cost estimate

packages/contracts/src/networks.ts
```diff
@@ -1109,6 +1141,10 @@ export const SPHINX_NETWORKS: Array<SupportedNetwork> = [
     actionGasLimitBuffer: false,
     eip2028: true,
     actionTransactionBatching: true,
+    rollupStack: {
+      provider: 'None',
+      type: 'OP Stack',
+    },
```
sam-goldman, line 1147: Blast is a modified OP stack chain. I don't know whether this impacts the cost estimate

packages/core/test/convert.spec.ts
```diff
@@ -20,11 +20,6 @@ import {
   sleep,
 } from '../src'
 
-// We filter out app chains because we expect the mainnet test to cover them adequately
-const fetchNonRollupStackChains = () => {
-  return SPHINX_NETWORKS.filter((network) => network.rollupStack === undefined)
-}
```
sam-goldman, line 26: This filters out Optimism, Arbitrum, and Polygon zkEVM, which probably shouldn't be filtered out because these stacks won't have coverage at all

packages/core/test/convert.spec.ts
```diff
@@ -82,6 +77,14 @@ describe('Convert EthersJS Objects', () => {
...
     59141: '0x2c9b7cd04f5db3b6ab923dc8e8eabc256f1a1d15467df6014609109965b763fa',
     43: '0x8562f6585e1a345e6a09a214799f0f99ec63e36182da54d32f6e099ae90fe86c',
+    1918988905:
+      '0xa4cf88a9a354eaa9d7aa25ecc607a43cfdc4f82762254d72c4daea5a5899e47a',
+    1380012617:
+      '0x3536ea6046dbcb7b4e85dee0ebff92bca33d53d9defe4684e74a487721e1fb3c',
+    999999999:
+      '0x6d4b16783da5e871ccec22624fdca23417d47197e16c801a72bb10dea93d0ed2',
+    7777777:
+      '0x7f8617c8a7d54ed192c6eff8d995c383594359ff7ceab9584747409df471599d',
```
sam-goldman, line 87: Do you think we should include app chains in this test case? It may not be necessary. If I'm missing context, it'd be helpful to document why we include app chains

sam-goldman (commented):
Looks good, just a few comments and questions

packages/contracts/src/networks.ts
RPate97: I don't think that matters really, I'm going to leave it out until we have a reason why we need that info.

packages/core/test/convert.spec.ts
RPate97: That's why this function was removed. It's no longer used to filter the networks at all.

packages/core/test/convert.spec.ts
RPate97: I think this entire testing pattern is impractical to scale to a large number of app chains. However, I think resolving that is outside of the scope of this PR. I'd rather just run the tests for all network until we implement a better solution for running tests on a large number of networks including lots of app-specific rollups.

packages/contracts/src/networks.ts
RPate97: When we eventually move all of this information into the DB, it'll be better to have a defined value vs null. That provides better type safety when interacting with the DB.

packages/core/test/convert.spec.ts
sam-goldman: My bad, I guess i'm color blind

sam-goldman (approved):

PR #1667, opened 2024-04-12
feat(pg): Register projects via the website

RPate97:
## Purpose
Implements the mono repo logic required to switch to registering projects using the website.

## Architecture
The high-level architecture for project management now looks like this:
* All projects are registered by calling the `createProject` function. All of the configuration option validation we previously did in the mono repo is now done in this function. See the [corresponding website PR](https://github.com/sphinx-labs/sphinx-managed/pull/188) for more information. 
    * It's worth noting that there is no reason why we could not provide a method for registering projects that uses the CLI. It would simply be a small wrapper on top of the `createProject` function. The vast majority of the logic would be shared with the UI registration process. I haven't done that in this PR, but [I've created a ticket for it](https://linear.app/chugsplash/issue/CHU-1018/allow-registration-via-the-cli). 
* The propose API endpoint now only requires a project name and safe address. If no project with that name and Gnosis Safe address combination is found in the DB, then an error is thrown. 
* All the Gnosis Safe configuration information is now stored in the `sphinx.lock` file which we recommend committing to version control. 
    * When the `propose` command is run, we always update the `sphinx.lock` file by querying the website backend. 
    * When the `deploy` command is run, we rely on the current version of the `sphinx.lock` file. A project can exist on the website and not in the local `sphinx.lock` file. In this case, we error recommending the user run `npx sphinx sync` to fetch the latest lock file. 
    * When the user runs tests that use the `sphinx` modifier, we rely on the current version of the `sphinx.lock` file. Like the `deploy` command, it's possible for the project to exist on the website and not in the local `sphinx.lock` file and we also error in this case recommending the user run `npx sphinx sync`. 

## Backward Compatability
This change is designed to be backward compatible on the website. The user will just no longer be able to register new projects via the propose API endpoint. So if a user continues to use the current version of our plugin, and uses a set of configuration options in their script that correctly match a project that is already registered on the website (correct project name, and Gnosis Safe options that resolve to the correct address), then they will be able to continue proposing to the website without issue. Being backward compatible in this way, will ensure that the release process for this change is as smooth as possible. 

However, the changes in the mono repo are breaking changes because the user will need to update their script configuration options. See the [migration guide](https://github.com/sphinx-labs/sphinx/pull/1667/files#diff-79751e908079863a605ef5162e70e2744e84345906f205b431688c94e53182d9) for more information on that.

## Testing
You should be able to test this entire change out by running the website locally. You'll just need to checkout this branch, and the [corresponding branch in the website repo](https://github.com/sphinx-labs/sphinx-managed/pull/188). You'll of course want to make sure that you have the mono repo packages linked in the website repo. 

Once you have that all setup, you should be able to create a project using the UI and then propose using that project name from whatever script you want as long as you have it configured properly. See the [migration guide](https://github.com/sphinx-labs/sphinx/pull/1667/files#diff-79751e908079863a605ef5162e70e2744e84345906f205b431688c94e53182d9) included in this PR for more information on that.

## Todo
- [ ] Confirm that this works on windows


package.json
```diff
@@ -39,7 +39,7 @@
     "clean": "yarn workspaces run clean",
     "lint": "yarn workspaces run lint",
     "lint:fix": "yarn workspaces run lint:fix",
-    "changeset:version:custom": "yarn changeset version && yarn workspace @sphinx-labs/contracts run write-version"
+    "changeset:version:custom": "yarn changeset version && yarn workspace @sphinx-labs/contracts run write-version && yarn workspace @sphinx-labs/plugins run write-version"
```
RPate97, line 42: This allows us to access the current version of the plugins package which we include in the request to the proposal API endpoint. 

Originally, I was going to use this to determine when to deprecate some things in the mono repo related to this change. However, that ended up being unnecessary. I still think it's helpful to know what plugins package versions people are using though, so I decided to keep all of this logic.

packages/contracts/contracts/foundry/Sphinx.sol
```diff
@@ -53,7 +53,7 @@ abstract contract Sphinx {
      * @dev The configuration options for the user's project. This variable must have `internal`
      *      visibility so that the user can set fields on it.
      */
-    SphinxConfig public sphinxConfig;
+    UserSphinxConfig public sphinxConfig;
```
RPate97, line 56: I've renamed `SphinxConfig` to `UserSphinxConfig` in the Solidity plugin since it now has significantly less fields (only `projectName`, `mainnets`, and `testnets`). I've also introduced a new type `InternalSphinxConfig` which is exactly the same as the old `SphinxConfig` type. `InternalSphinxConfig` is used within our plugin and is included in the `DeploymentConfig` object so that this update is not a breaking change to the executor.

packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -276,46 +286,17 @@ contract SphinxUtils is SphinxConstants {
             ISphinxScript(_script).configureSphinx();
         }
 
-        SphinxConfig memory config = ISphinxScript(_script).sphinxFetchConfig();
+        UserSphinxConfig memory config = ISphinxScript(_script).sphinxFetchConfig();
         validate(config);
         return config;
     }
 
-    function assertValidOwners(SphinxConfig memory _config) private pure {
```
RPate97, line 284: We no longer need this check since we do all of this validation when the project is created via the website.

packages/core/src/types.ts
```diff
@@ -7,3 +7,19 @@ export type StoreDeploymentConfig = (
...
+      projectName: string
+      defaultSafe: {
+        safeIdentifier: string
+        owners: string[]
+        threshold: string
+        saltNonce: string
+      }
+    }
+  }
+}
```
RPate97, line 25: This is the format of the `sphinx.lock` file, a couple notes on it:
1. `projectId` is the literal id stored in the DB for this project, it is generated by us and not by the user. We don't currently use it for anything, but I think it could be useful. 
2. I put all of the Gnosis Safe information into a `defaultSafe` object instead of just having top-level fields because I want this type to accommodate future features like importing external safes.

packages/core/src/types.ts
```diff
@@ -7,3 +7,19 @@ export type StoreDeploymentConfig = (
...
+      projectName: string
+      defaultSafe: {
+        safeName: string
+        owners: string[]
+        threshold: string
+        saltNonce: string
+      }
+    }
+  }
+}
```
RPate97, line 25: This is the format of the `sphinx.lock` file, a couple notes on it:
1. `projectId` is the literal id stored in the DB for this project, it is generated by us and not by the user. We don't currently use it for anything, but I think it could be useful. 
2. I put all of the Gnosis Safe information into a `defaultSafe` object instead of just having top-level fields because I want this type to accommodate future features like importing external safes.

packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -324,41 +305,21 @@ contract SphinxUtils is SphinxConstants {
...
-
-        address[] memory duplicateOwners = getDuplicatedElements(_config.owners);
-        require(
-            duplicateOwners.length == 0,
-            string(
-                abi.encodePacked(
-                    "Sphinx: Your 'sphinxConfig.owners' array contains duplicate addresses: ",
-                    toString(duplicateOwners)
-                )
-            )
```
RPate97, line 350: We no longer need these checks since we do all of this validation when the project is created via the website. See the corresponding PR in the website repo for all of that logic. I've also included a comprehensive suite of tests there.

packages/contracts/contracts/foundry/SphinxPluginTypes.sol
```diff
@@ -135,7 +135,13 @@ struct InitialChainState {
     bool isExecuting;
 }
 
-struct SphinxConfig {
+struct UserSphinxConfig {
+    string projectName;
+    string[] mainnets;
+    string[] testnets;
+}
```
RPate97, line 142: You might notice that there is no `orgId` field here. That is because I decided to switch to requiring the user to define an environment variable `SPHINX_ORG_ID` instead. This simplifies some things on our end since we don't have to call into the users script to fetch the org id before updating the `sphinx.lock` file. It also simplifies the UX of some of the CLI commands for the user.

For example, I've introduced a new command in this PR, `sphinx sync,` which fetches the latest `sphinx.lock` file from our backend. Since both the orgId and API key are defined as environment variables, the user doesn't have to supply arguments to this command. Likewise, we could update the `sphinx artifacts` command to only accept a project name instead of both a project name and an org id (though I have not made that change in this PR).

packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -276,46 +279,17 @@ contract SphinxUtils is SphinxConstants {
             ISphinxScript(_script).configureSphinx();
         }
 
-        SphinxConfig memory config = ISphinxScript(_script).sphinxFetchConfig();
+        UserSphinxConfig memory config = ISphinxScript(_script).sphinxFetchConfig();
         validate(config);
         return config;
     }
 
-    function assertValidOwners(SphinxConfig memory _config) private pure {
```
RPate97, line 284: Since registration now occurs via the website, these checks are not necessary here. Instead, I've implemented the same set of checks in the website mutation where the project is created.

packages/contracts/src/addresses.ts
```diff
@@ -187,3 +191,146 @@ export const getCheckBalanceLowAddress = () => {
...
+
+  // Sort the copied array
+  return arrCopy.sort((a, b) => {
+    const aBigInt = BigInt(a)
+    const bBigInt = BigInt(b)
+    return aBigInt < bBigInt ? -1 : aBigInt > bBigInt ? 1 : 0
+  })
+}
+
+export const getGnosisSafeInitializerData = (
```
RPate97, line 214: Since project creation occurs entirely via the website, we need a method for calculating the safe address in JavaScript to store in the DB. This function was originally in the test files, I pulled it into the contracts package and added some tests for it so we can use it in production.

packages/contracts/src/addresses.ts
```diff
@@ -187,3 +191,146 @@ export const getCheckBalanceLowAddress = () => {
...
+  if (owners.length === 0) {
+    throw new Error(
+      "Sphinx: You must have at least one owner in your 'owners' array."
+    )
+  }
+  if (threshold === 0) {
+    throw new Error(
+      "Sphinx: You must set your 'threshold' to a value greater than 0."
+    )
+  }
```
RPate97, line 227: We don't rely on these checks on the website, so I haven't added tests for them. However, I don't see a strong reason to remove them.

packages/contracts/src/addresses.ts
```diff
@@ -187,3 +191,146 @@ export const getCheckBalanceLowAddress = () => {
...
+    // deployment. We don't use them.
+    ethers.ZeroAddress,
+    0,
+    ethers.ZeroAddress,
+  ])
+
+  return safeInitializerData
+}
+
+export const getGnosisSafeProxyAddress = (
```
RPate97, line 308: Since project creation occurs entirely via the website, we need a method for calculating the safe address in JavaScript to store in the DB. This function was originally in the test files, I pulled it into the contracts package and added some tests for it so we can use it in production.

packages/contracts/src/utils.ts
```diff
@@ -258,23 +256,3 @@ const isValidStorageLayout = (storageLayout: any): boolean => {
 export const isNonNullObject = (obj: any): boolean => {
   return typeof obj === 'object' && obj !== null
 }
-
-export const getCurrentGitCommitHash = (): string | null => {
```
RPate97, line 262: I've moved this to the plugins package because this function is not used in contracts or core. It also was causing problems for me when I tried to import other functions (because it uses `execSync` which is not available in browsers).

packages/contracts/test/ScriptConfiguration.t.sol
```diff
@@ -23,23 +23,17 @@ contract ScriptConfiguration_Test is Test, Sphinx, SphinxTestUtils {
     }
 
     function configureSphinx() public override {
-        sphinxConfig.projectName = "test_project";
-        sphinxConfig.threshold = 1;
-        sphinxConfig.orgId = "test-org-id";
-        sphinxConfig.owners = [0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266];
+        sphinxConfig.projectName = "Simple_Project";
```
RPate97, line 26: To keep the `sphinx.lock` file simple, all our tests use a single project `Simple_Project`.

packages/contracts/sphinx.lock
```diff
@@ -0,0 +1,29 @@
...
+        "safeName": "Simple_Project_1",
+        "owners": [
+          "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266"
+        ],
+        "threshold": "1",
+        "saltNonce": "1"
+      }
+    }
+  }
+}
```
RPate97, line 29: To keep this file simple, I only created separate projects when we had scripts that used different configuration options (excluding the project name). I did the same thing in the `plugins` package.

packages/core/src/utils.ts
```diff
@@ -378,6 +380,39 @@ export const fetchSphinxManagedBaseUrl = () => {
     : 'https://www.sphinx.dev'
 }
 
+export const readSphinxLock = async (): Promise<SphinxLock> => {
+  if (!existsSync(SPHINX_LOCK_PATH)) {
+    throw new Error(
+      'Missing `sphinx.lock` file, run `npx sphinx sync` to regenerate this file. We recommend committing it to version control.'
+    )
+  }
```
RPate97, line 388: This error can only be thrown when running the `deploy` command which just reads the local `sphinx.lock` file. When the `propose` command is run, we always fetch the latest `sphinx.lock` file, so this error cannot occur.

packages/contracts/test/issues/CHU663/CHU663.s.sol
RPate97, line 1: This file was duplicated for some reason.

packages/core/src/actions/types.ts
```diff
@@ -160,19 +160,16 @@ export type ProposalRequest = {
...
-  safeInitSaltNonce: string
-  deploymentName: string
+  projectName: string
   chainIds: Array<number>
   projectDeployments: Array<ProjectDeployment>
   gasEstimates: Array<NetworkGasEstimate>
   diff: SphinxPreview
   compilerConfigId: string | undefined
   deploymentConfigId: string | undefined
+  sphinxPluginVersion: string | undefined
```
RPate97, line 172: We send this value to the propose API endpoint so we can track it in analytics. This will help us better understand when we can safely deprecate things in the propose API endpoint.

packages/plugins/package.json
```diff
@@ -30,7 +30,8 @@
     "lint:ts:check": "eslint . --max-warnings=0",
     "pre-commit": "lint-staged",
     "yalc:publish": "yarn build && yalc publish",
-    "sync:rpc": "infisical export --path='/RPC' --env=dev >> .env"
+    "sync:rpc": "infisical export --path='/RPC' --env=dev >> .env",
+    "write-version": "npx ts-node script/write-version.ts > src/cli/version.ts"
```
RPate97, line 34: This script writes the sphinx plugin version constant. It works the same as the `contracts` package script we already have.

packages/plugins/src/sample-project/sample-contracts.ts
```diff
@@ -44,17 +44,11 @@ contract HelloSphinxScript is Sphinx, Script {
...
-        sphinxConfig.threshold = 1;
         sphinxConfig.projectName = "My_First_Project";
     }
 
     function run() public sphinx {
-        // Set the \`CREATE2\` salt to be the hash of the owner(s). Prevents
-        // address collisions.
-        bytes32 salt = keccak256(abi.encode(sphinxConfig.owners));
-        helloSphinx = new HelloSphinx{ salt: salt }("Hi", 2);
+        helloSphinx = new HelloSphinx("Hi", 2);
```
RPate97, line 51: This is a rider, I updated the sample project to just use create instead of create2.

docs/cli-existing-project.md
```diff
@@ -145,60 +153,41 @@ We'll explain the Sphinx modifier in a later guide.
...
 Copy and paste the following `configureSphinx()` function template into your script:
 
 ```sol
 function configureSphinx() public override {
-  sphinxConfig.owners = [<your address>];
-  sphinxConfig.orgId = <Sphinx org ID>;
-  sphinxConfig.projectName = "My_First_Project";
-  sphinxConfig.threshold = 1;
+  sphinxConfig.projectName = "<your_project_name>";
 }
```
RPate97, line 165: We could simplify this even further by having the user define the project name as a top-level constant. However, that is somewhat non-trivial to handle, so I left it out of this PR. I've [created a ticket for this](https://linear.app/chugsplash/issue/CHU-1017/switch-to-configuring-script-project-name-as-a-top-level-constant).

docs/cli-quickstart.md
```diff
@@ -97,12 +106,6 @@ After you run the command, you'll notice several new files:
 - `.env`: A sample `.env` file that contains your credentials.
 - `.gitignore`: A sample `.gitignore` file that contains files and directories generated by Sphinx, Foundry, and Node.
 
-## 6. Test the deployment
-
-```
-forge test
-```
```
RPate97, line 104: This is also a rider.

packages/contracts/test/issues/CHU572.s.sol
RPate97, line 1: This test is no longer required.

packages/contracts/test/issues/CHU663/CHU663.s.sol
RPate97, line 1: This file was duplicated for some reason.

packages/contracts/test/mocha/utils.spec.ts
```diff
@@ -1,35 +0,0 @@
-import child_process from 'child_process'
-
-import { expect } from 'chai'
-import sinon from 'sinon'
-
-import { getCurrentGitCommitHash } from '../../src'
-
-describe('getCurrentGitCommitHash', () => {
```
RPate97, line 8: These tests have been moved to the plugins package along with the `getCurrentGitCommitHash` function.

packages/core/src/utils.ts
```diff
@@ -1112,42 +1130,6 @@ export const findStorageSlotKey = (
   return storageObj.slot
 }
 
-/**
- * Throws an error if the input project name contains invalid characters, which would prevent Sphinx
- * from using it as a file name when writing the deployment artifacts. This function is in
- * TypeScript instead of Solidity because it's impractical to implement regular expressions in
- * Solidity.
- */
-export const assertValidProjectName = (input: string): void => {
```
RPate97, line 1121: All of this validation occurs in the website backend now.

packages/contracts/contracts/foundry/SphinxPluginTypes.sol
```diff
@@ -192,6 +198,18 @@ struct SystemContractInfo {
     address expectedAddress;
 }
 
+struct DefaultSafe {
```
sam-goldman, line 201: What is the default safe? docs would be nice

docs/breaking/registration.md
```diff
@@ -0,0 +1,53 @@
...
+### 2. Configure the `SPHINX_ORG_ID` environment variable
+Sphinx now requires your org id to be defined as an environment variable instead of in your scripts. You'll want to update your `.env` file to include it.
+```
+SPHINX_ORG_ID=<your org id>
+```
+
+> If you are using Sphinx in CI, you will also need to define the `SPHINX_ORG_ID` environment variable in your CI process.
+
+### 3. Fetch your `sphinx.lock` file and commit it to version control
+Sphinx now stores information about your projects in a `sphinx.lock` file. This is an autogenerated file. You should not modify it manually. Whenever you register a new project with Sphinx, you'll want to update your `sphinx.lock` file.
```
sam-goldman, line 27: > You should not modify it manually. Whenever you register a new project with Sphinx, you'll want to update your `sphinx.lock` file.

I understand what you mean here, but these sentences seem to contradict each other. You could change the second sentence to something like: 
> Whenever you register a new project with Sphinx, you'll want to update your `sphinx.lock` file by running the following command:

docs/cli-quickstart.md
```diff
@@ -66,27 +66,36 @@ pnpm:
 pnpm add -D @sphinx-labs/plugins https://github.com/foundry-rs/forge-std.git#v1.7.1 https://github.com/dapphub/ds-test.git#e282159d5170298eb2455a6c05280ab5a73a4ef0
 ```
 
-## 5. Initialize a project
+## 5. Create a new Sphinx project
+Go to the Sphinx website, sign up, and click the "Create Project" button. You'll need to fill in the following fields:
```
sam-goldman, line 70: > I think we should add a hyperlink to "Sphinx website"

docs/cli-quickstart.md
```diff
@@ -66,27 +66,36 @@ pnpm:
...
+* Project Name: A name for your project. For this guide, we recommend using `My_First_Project`.
+* Gnosis Safe Owners: A set of addresses that will own this project. For this guide, you should use a single EOA you control.
+* Gnosis Safe Threshold: The number of signatures required to execute a deployment. For this guide, you should set the threshold to 1.
+* Gnosis Safe Salt Nonce: An optional value that allows you to register multiple projects with the same set of owners. For this guide, you should leave this field blank.
+
+After you've finished creating the project, you'll see your Org Id, API Key, and Project Name. You'll need these values for the next step.
```
sam-goldman, line 76: Org Id -> Org ID

sam-goldman, line 76: > After you've finished creating the project, you'll see your Org Id, API Key, and Project Name.

I think we should finish this sentence with "in the website" or something

sam-goldman (commented):
Finished a first pass. I'm probably halfway done with the review. Looks great so far!

packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -1295,4 +1254,42 @@ contract SphinxUtils is SphinxConstants {
...
+
+        string memory basePath = concatJsonPath(".projects", _config.projectName);
+        bool exists = vm.keyExists(json, basePath);
+        if (!exists) {
+            revert(
+                string(
+                    abi.encodePacked(
+                        "Project with the name ",
+                        bytes(_config.projectName),
+                        " was not found in the `sphinx.lock` file. You need to register this project in the sphinx UI and then run `npx sphinx sync` to generate the latest `sphinx.lock` file. We recommend committing this file to version control."
```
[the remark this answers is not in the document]
RPate97: I don't think it's irrelevant. I included this because I assume the user will inevitably register a project and forget to run `sphinx sync` and commit the changes to version control.

You correctly point that out in this [comment](https://github.com/sphinx-labs/sphinx/pull/1667#discussion_r1577113767).

In this case, the following scenarios can occur:
1. The user proposes using the newly registered project. This works just fine since we fetch the `sphinx.lock` file from the backend during the proposal.
2. The user runs the deploy CLI command using the newly registered project or runs a test that utilizes the `sphinx` modifier. This fails with an error indicating that they should run `sphinx sync` and reminding them that this file should be committed to version control.

docs/breaking/registration.md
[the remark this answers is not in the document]
RPate97: I agree; the user will likely forget to do this. I think it's acceptable b/c it's only an issue for the deploy CLI command or if your tests utilize sphinx. If you're proposing, then everything will work correctly. 

I've also [designed the error messages to account for this scenario](https://github.com/sphinx-labs/sphinx/pull/1667#discussion_r1578281344).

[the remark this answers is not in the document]
RPate97: If you think it would be helpful, I could add a prompt to the website telling the user to run `sphinx sync` in their repo after they've registered a project.

docs/breaking/registration.md
```diff
@@ -0,0 +1,53 @@
...
+### 1. Update your Sphinx Plugin and Sphinx Solidity Library
+
+Update the Sphinx plugin version by bumping the version in your `package.json` file to the latest version `0.32.0`.
+
+Then update the Sphinx Solidity library by running `sphinx install`:
+```
+npx sphinx install
+```
+
+### 2. Configure the `SPHINX_ORG_ID` environment variable
```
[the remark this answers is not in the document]
RPate97: Yeah, I think that's reasonable. I'm okay with moving this to the lock file.

packages/plugins/src/sample-project/sample-foundry-config.ts
```diff
@@ -54,7 +54,12 @@ dist/
...
   sphinxApiKey: string,
+  sphinxOrgId: string,
   alchemyApiKey: string
 ): string => {
-  return `SPHINX_API_KEY=${sphinxApiKey}\n` + `RPC_API_KEY=${alchemyApiKey}`
+  return (
+    `SPHINX_API_KEY=${sphinxApiKey}\n` +
+    `SPHINX_ORG_ID=${sphinxOrgId}` +
+    `RPC_API_KEY=${alchemyApiKey}`
+  )
```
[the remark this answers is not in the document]
RPate97: The demo test suite doesn't use these values because it just deploys using the deploy CLI command. I've updated the test check that they were written correctly.

packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -1295,4 +1254,42 @@ contract SphinxUtils is SphinxConstants {
...
     }
+
+    function concatJsonPath(string memory a, string memory b) public pure returns (string memory) {
+        return string(abi.encodePacked(a, ".", b));
+    }
+
+    function fetchProjectFromLock(address _script) public returns (SphinxLockProject memory) {
+        UserSphinxConfig memory _config = fetchAndValidateConfig(_script);
+        string memory root = vm.projectRoot();
+        string memory path = string(abi.encodePacked(root, "/sphinx.lock"));
```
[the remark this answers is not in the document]
RPate97: I believe so, but I've made a note to test windows before releasing this.

packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -1295,4 +1254,42 @@ contract SphinxUtils is SphinxConstants {
 
         return _deploymentInfo;
     }
+
+    function concatJsonPath(string memory a, string memory b) public pure returns (string memory) {
+        return string(abi.encodePacked(a, ".", b));
+    }
+
+    function fetchProjectFromLock(address _script) public returns (SphinxLockProject memory) {
```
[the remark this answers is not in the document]
RPate97: The check for whitespace should have been carried over to the website, I've added it back in and added a test for it in the website PR.

.changeset/short-cycles-film.md
```diff
@@ -0,0 +1,8 @@
+---
+'@sphinx-labs/contracts': minor
+'@sphinx-labs/plugins': minor
```
sam-goldman, line 3: Before this PR, users commonly ran into errors like this at the end of the proposal process:
> Detected another project with the same owners, salt, and threshold but a different name. We currently do not support changing the name of an existing project. Please update your script to use the correct name:...

I noticed that we throw these types of error messages during project registration now, which is great. Now that this PR is implemented, are there any other errors that users could commonly hit at the end of the proposal process? I'm asking because this was one of the most frustrating aspects of the proposal process for me

sam-goldman (commented):
Some more comments and questions. Mainly minor stuff

packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -1295,4 +1254,43 @@ contract SphinxUtils is SphinxConstants {
...
+
+        string memory basePath = concatJsonPath(".projects", _config.projectName);
+        bool exists = vm.keyExists(json, basePath);
+        if (!exists) {
+            revert(
+                string(
+                    abi.encodePacked(
+                        "Project with the name ",
+                        bytes(_config.projectName),
+                        " was not found in the `sphinx.lock` file. You need to register this project in the Sphinx UI and then run `npx sphinx sync` to generate the latest `sphinx.lock` file. We recommend committing this file to version control."
```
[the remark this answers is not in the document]
RPate97: https://github.com/sphinx-labs/sphinx/pull/1667#discussion_r1578281344

packages/core/src/utils.ts
```diff
@@ -1112,42 +1135,6 @@ export const findStorageSlotKey = (
...
-  // Reserved names (commonly reserved in Windows)
-  const reservedNames = /^(con|prn|aux|nul|com[0-9]|lpt[0-9])(\..*)?$/i
-  if (reservedNames.test(input)) {
-    throw new Error('Project name uses a reserved name in Windows.')
-  }
-
-  // Check for empty names
-  if (input.length === 0) {
-    throw new Error('Project name cannot be empty.')
-  }
```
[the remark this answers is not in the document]
RPate97: https://github.com/sphinx-labs/sphinx-managed/blob/pate/chu-634-require-explicit-project-registration-instead-of-implicitly/website/server/api/graphql/models/projects/mutations/createProject.ts#L192

packages/plugins/src/cli/propose/index.ts
```diff
@@ -290,6 +292,8 @@ export const propose = async (
     'SphinxPluginTypes'
   )
 
+  await syncSphinxLock(undefined, apiKey)
```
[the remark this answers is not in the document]
RPate97: We don't do this in the deploy command, but I've added docs to this line.

.changeset/short-cycles-film.md
```diff
@@ -0,0 +1,8 @@
+---
```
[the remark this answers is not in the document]
RPate97: Unfortunately, [JSON doesn't support comments](https://www.freecodecamp.org/news/comments-in-json/#:~:text=The%20primary%20reason%20why%20JSON,a%20pure%20data%2Donly%20format.). So, we would have to use a different file format (which I don't think is worth it). Or include this as a warning in the object itself, which would look like this: 
```
{
  "warning": "THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.",
  "format": "sphinx-lock-1",
  "orgId": "test-org-id",
  "projects": {
    "Simple_Project": {
      "projectId": "",
      "projectName": "Simple_Project",
      "defaultSafe": {
        "safeName": "Simple_Project",
        "owners": [
          "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266"
        ],
        "threshold": "1",
        "saltNonce": "0"
      }
    }
  }
}
```

docs/breaking/registration.md
```diff
@@ -0,0 +1,48 @@
+# Project Registration Update
+Sphinx is updating the way that you register projects. Instead of specifying your Gnosis Safe owners, threshold and salt nonce in your Solidity scripts, you will now register your projects via the Sphinx UI. In your script, you will simply specify the name of your project.
```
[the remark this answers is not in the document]
RPate97: I was planning to just message everyone directly since that's what we've been doing.

We could do something better where we detect if their script is out of date and prompt them to upgrade, but that's non-trivial, especially if we want to remove the old configuration fields from the script with this change (which is preferable IMO). Since we still don't have many users, I wouldn't bother with it.

That being said, it would be a good idea to keep in mind that integrations may be outdated when following up with users.

docs/breaking/registration.md
```diff
@@ -0,0 +1,48 @@
...
+
+This change is intended to improve the UX of using Sphinx and to open the door for us to implement other feature requests such as sharing Gnosis Safes between Sphinx projects and importing Gnosis Safes that were deployed using alternative interfaces (such as the Gnosis Safe UI) into Sphinx.
```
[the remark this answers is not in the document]
RPate97: I think that's just part of the release process for this change. It's not helpful to me to have a separate ticket for it. If you think it's helpful, though, go ahead.

.changeset/short-cycles-film.md
```diff
@@ -0,0 +1,8 @@
...
+'@sphinx-labs/contracts': minor
+'@sphinx-labs/plugins': minor
+'@sphinx-labs/core': minor
```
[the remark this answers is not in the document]
RPate97: The issue with syncing in the deploy CLI command is that the user then cannot really offboard from our platform.

To be honest, this sort of question is insignificant enough that I really wouldn't worry about it. I would like to get this change out sooner rather than later, if the current way the deploy cli command works is an issue then I'm happy to change it. But I don't think it's worth spending much time on this feature.

packages/core/src/types.ts
```diff
@@ -7,3 +7,20 @@ export type StoreDeploymentConfig = (
...
 
 export type RelayProposal = (proposalRequest: ProposalRequest) => Promise<void>
+
+export type SphinxLock = {
+  format: string
+  orgId: string
+  projects: {
+    [projectName: string]: {
+      projectId: string
+      projectName: string
```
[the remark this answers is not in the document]
RPate97: It can be helpful to have the data available in the object itself rather than requiring it to be pulled from the key. I don't think it matters that the data is repeated since this file is autogenerated.

packages/core/src/types.ts
```diff
@@ -7,3 +7,20 @@ export type StoreDeploymentConfig = (
...
+      defaultSafe: {
+        safeName: string
```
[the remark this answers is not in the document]
RPate97: They are distinct values in the database. This structure is designed to allow you to have safes that are not tied to a specific project.

.changeset/short-cycles-film.md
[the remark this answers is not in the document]
RPate97: I already fixed this. [You are not be allowed to have whitespace in the project name.](https://github.com/sphinx-labs/sphinx-managed/blob/pate/chu-634-require-explicit-project-registration-instead-of-implicitly/website/server/api/graphql/models/projects/mutations/createProject.ts#L208)

RPate97: There are a number of checks that a user could theoretically encounter when the proposal request is sent to the backend. If you are interested, you can [review them all here](https://github.com/sphinx-labs/sphinx-managed/blob/pate/chu-634-require-explicit-project-registration-instead-of-implicitly/website/server/api/propose/index.ts#L561). 

In practice, most of these are checking for things that indicate a bug in our tooling rather than a user error. For example, we still check that the user proposed with a `projectName` and `safeAddress` that belong to the same project. This check should no longer be hit due to user error because the user only defines the project name in their script. It's on us to make sure we're using the correct Safe for that project. 

The only check the user realistically may hit due to their error is the [currently executing project check](https://github.com/sphinx-labs/sphinx-managed/blob/pate/chu-634-require-explicit-project-registration-instead-of-implicitly/website/server/api/propose/index.ts#L585). This error is thrown if they propose a deployment while another one is in the process of being executed.

packages/contracts/contracts/foundry/SphinxUtils.sol
```diff
@@ -276,46 +276,17 @@ contract SphinxUtils is SphinxConstants {
...
+        UserSphinxConfig memory config = ISphinxScript(_script).sphinxFetchConfig();
         validate(config);
         return config;
     }
 
-    function assertValidOwners(SphinxConfig memory _config) private pure {
-        require(
-            _config.owners.length > 0,
-            "Sphinx: You must have at least one owner in your 'sphinxConfig.owners' array before calling this function."
-        );
```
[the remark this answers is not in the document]
RPate97: [Added](https://github.com/sphinx-labs/sphinx-managed/blob/pate/chu-634-require-explicit-project-registration-instead-of-implicitly/website/server/api/graphql/models/projects/mutations/createProject.ts#L123)

.changeset/short-cycles-film.md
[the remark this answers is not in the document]
sam-goldman: I think it's fine not to include it

sam-goldman (approved):
Great stuff

PR #1674, opened 2024-04-14
maint(ci): Bump docker container version

RPate97:
## Purpose
- Bumps the foundry version used in our CI process
- Fixes an issue in our tests that was broken by an update in foundry

### DIFF
--- a/packages/core/src/config/types.ts
+++ b/packages/core/src/config/types.ts
@@ -268,7 +268,7 @@ interface AbstractFoundryTransaction {
     type: string | null
     from: string | null
     gas: string | null
-    data: string | null
+    input: string | null
     nonce: string | null
     accessList: string | null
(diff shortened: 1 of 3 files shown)


packages/core/src/config/types.ts
```diff
@@ -268,7 +268,7 @@ interface AbstractFoundryTransaction {
     type: string | null
     from: string | null
     gas: string | null
-    data: string | null
+    input: string | null
```
RPate97, line 271: The foundry devs changed the `data` field to `input` in the broadcast file. This is a prime example of why we have to be thoughtful about what parts of Foundry we rely on and why it's important that we use Foundry features in their intended ways. If we still relied on the broadcast file, this update would have randomly broken our entire plugin.

PR #1686, opened 2024-04-18
fix(pg): Pass through error if API response unexpected

RPate97:
## Purpose
Causes API errors to be passed through to the console if the response code is unexpected. This should help with debugging API-related issues that are hard to reproduce.

### DIFF
--- a/packages/core/src/utils.ts
+++ b/packages/core/src/utils.ts
@@ -404,9 +404,7 @@ export const relayProposal: RelayProposal = async (
         `Internal server error. Please report this to the developers.`
       )
     } else {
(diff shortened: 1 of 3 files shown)


packages/core/src/utils.ts
```diff
@@ -404,9 +404,7 @@ export const relayProposal: RelayProposal = async (
         `Internal server error. Please report this to the developers.`
       )
     } else {
-      throw new Error(
-        `Unexpected response code. Please report this to the developers.`
-      )
+      throw new e()
```
sam-goldman, line 407: This looks like invalid syntax

PR #1687, opened 2024-04-18
maint(ct): Fix test broken by Foundry update

RPate97:
## Purpose
This fixes a minor issue in one of our tests that was introduced by this [PR in Foundry](https://github.com/foundry-rs/foundry/pull/7653). This issue does not effect our service for any users.

### DIFF
--- a/packages/contracts/test/ManagedService.t.sol
+++ b/packages/contracts/test/ManagedService.t.sol
@@ -209,6 +209,6 @@ contract ManagedService_Test is Test, ManagedService {
 
     function test_RevertIfOwnerIsAddressZero() external {
         vm.expectRevert("ManagedService: admin cannot be address(0)");
-        new ManagedService{ salt: keccak256("1") }(address(0));
+        new ManagedService(address(0));
     }
 }


packages/contracts/test/ManagedService.t.sol
```diff
@@ -209,6 +209,6 @@ contract ManagedService_Test is Test, ManagedService {
 
     function test_RevertIfOwnerIsAddressZero() external {
         vm.expectRevert("ManagedService: admin cannot be address(0)");
-        new ManagedService{ salt: keccak256("1") }(address(0));
```
RPate97, line 212: The update to Foundry to use the create2 factory in the simulation caused this test to break since the error if the create2 factory is used differs from the error we expect on line 211. I resolved this issue just by deploying via the normal create opcode. 

For reference, this is the result if the create2 factory is used:
Screenshot 2024-04-18 at 3 55 21 PM

sam-goldman (approved):
Looks good. I think this PR resolves [this ticket](https://linear.app/chugsplash/issue/CHU-997/investigate-why-contract-package-test-suite-failed-in-ci)

PR #1692, opened 2024-04-25
audit considerations

sam-goldman:
This branch won't be merged anywhere. I'm just using it as an easy way to share design considerations for the upcoming audit.

### DIFF
--- a/executor.md
+++ b/executor.md
@@ -1,0 +1,24 @@
+# The `executor` Merkle leaf field
+
+I think we should remove the `executor` field from the Merkle tree. This means anybody will be able to execute a user's deployment.
+
+I don't think removing this field meaningfully changes the trust assumptions that the user places on the executor. You can read about these trust assumptions in the `SphinxModuleProxy` spec for [buggy executors](https://github.com/sphinx-labs/sphinx/blob/develop/specs/sphinx-module-proxy.md#buggy-executor) and [malicious executors](https://github.com/sphinx-labs/sphinx/blob/develop/specs/sphinx-module-proxy.md#malicious-executor).
+
+Removing this field would introduce the possibility of one edge case, which could add complexity to the relayer. Specifically, it'll be possible for anybody to front-run our relayer's transactions, causing them to error. Here's how this could happen:
+1. We submit a transaction to the mempool at the same time that somebody else submits a transaction to execute the same set of Merkle leaves.
+2. Their transaction gets accepted first, causing an error in the Sphinx Module for our transaction, which will bubble up to the relayer.
+
+I don't think this edge case is likely to happen, but I'm not sure whether our relayer can recover gracefully from this.
+
+If you'd like to understand the rationale for removing this field, read the next section.
+
+## Rationale
+
+If we keep the `executor` field, the following griefing risk can occur, which would significantly degrade the UX of the arbitrary chain feature:
+1. Say we leak the private key of the account that owns the `ManagedService`.
+2. An attacker could take control of the `ManagedService` by changing its owner and removing all of our relayers. This will prevent us from executing any transactions through the `ManagedService`.
+3. The attacker could grief users of the arbitrary chain feature in a variety of ways. Here's a non-exhaustive list, which should hopefully give you a sense of the headache this would cause:
+  * The attacker could make a deployment active, which would require us to create a new Merkle tree with a `CANCEL` leaf. After this new Merkle root is signed, the attacker could finish executing the deployment, invalidating the new Merkle root.
+  * The attacker could partially execute deployments on different chains to varying degrees. For example, it could execute half of a deployment on Chain A and three-quarters of a deployment on Chain B. This could make it infeasible for the user to sign a single new Merkle root.
+    * If the user's deployment uses `CREATE`, the attacker could make it difficult for the user to get consistent addresses by partially executing deployments.
+    * The user could theoretically switch to a new Gnosis Safe, but that would also cause inconsistent addresses across networks if the user deploys via `CREATE` and has already executed deployments on some chains.
--- a/packages/contracts/contracts/core/SphinxModule.sol
+++ b/packages/contracts/contracts/core/SphinxModule.sol
@@ -34,7 +34,7 @@ contract SphinxModule is ReentrancyGuard, Enum, ISphinxModule, Initializable {
     /**
      * @inheritdoc ISphinxModule
      */
-    string public constant override VERSION = "1.0.0";
+    string public constant override VERSION = "1.0.0"; // TODO(later): what version?
 
     /**
      * @dev The hash of the version string for the Gnosis Safe proxy v1.3.0.
@@ -72,12 +72,7 @@ contract SphinxModule is ReentrancyGuard, Enum, ISphinxModule, Initializable {
     /**
      * @inheritdoc ISphinxModule
      */
-    uint256 public override merkleRootNonce;
-
-    /**
-     * @inheritdoc ISphinxModule
-     */
-    bytes32 public override activeMerkleRoot;
+    bytes32 public override latestMerkleRoot;
 
     /**
      * @inheritdoc ISphinxModule
@@ -130,8 +125,6 @@ contract SphinxModule is ReentrancyGuard, Enum, ISphinxModule, Initializable {
     {
         require(_root != bytes32(0), "SphinxModule: invalid root");
 
-        require(activeMerkleRoot == bytes32(0), "SphinxModule: active merkle root");
-
         // Check that the Merkle root hasn't been used before.
         MerkleRootState storage state = merkleRootStates[_root];
         require(state.status == MerkleRootStatus.EMPTY, "SphinxModule: root already used");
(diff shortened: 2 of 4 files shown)


executor.md
```diff
@@ -0,0 +1,24 @@
+# The `executor` Merkle leaf field
```
sam-goldman, line 1: @RPate97 lmk your thoughts

RPate97: The issue you bring up is legitimate, but I wouldn't solve it by removing the `executor` field. That field gives us a lot of flexibility to implement various schemes for selecting the executor without changing the module. I want to preserve that optionality. 

I would go with one of these two options to resolve the issue instead:
1. Update the `ManagedService` contract to be permissionless. This is functionality equivalent to having no specific `executor` but preserves the field. 
2. Do a better job of securing the `ManagedService` contract. This might involve reevaluating how it is deployed so that it is owned by a multisig wallet as soon as it is deployed. 

Both options have pros and cons. I'm happy to hop on a call to discuss further if you want.

sam-goldman: > That field gives us a lot of flexibility to implement various schemes for selecting the executor without changing the module. I want to preserve that optionality.

Could you expand on this? I don't see the benefit of keeping this field with our current system

sam-goldman: I guess this field would be useful if the user wants to be the only one that can execute the deployment? This has never been requested before, and would require us to build additional tooling, but I can understand why a user would want to do this in theory

sam-goldman: Actually, I remember Kelvin mentioning a long time ago that Optimism would probably want to execute their own deployments because they didn't want to rely on a remote executor that wasn't battle-tested. I'm not sure whether they would've cared if execution was permissionless though. I think teams with higher security requirements could reasonably want to be the sole executor

RPate97: The benefit is that it allows us to implement an unlimited number of different ways to determine who can execute the deployment without needing to change the module. Any future feature we implement related to this will be fully backwards compatible.

#### Non-exhaustive list of examples
1. Fully permissionless (this is what you are suggesting anyone can execute the deployment)
2. Fully permissioned (only we can execute the deployment)
3. Decentralized (the executor is selected through some decentralized process like a fee auction or staking)
4. Only the user (we use a specific address owned by the user)

If we remove the `executor` field entirely, we commit fully to option 1. If we ever want to change this, we must update the module. I don't feel great about this because I don't have evidence to indicate that we won't want to do something else in the future. 

So, since we can achieve something functionally equivalent to option 1 while preserving the `executor` field, I would do that to maintain optionality.

RPate97: It's worth noting that option 2 is probably a lot better from a business perspective.

RPate97: See telegram if you want to talk about this more. I would prefer to hop on a call.

migration.md
```diff
@@ -0,0 +1,34 @@
+How should we migrate existing users to the newest version of the Sphinx Module?
```
sam-goldman, line 1: @RPate97 lmk your thoughts

packages/contracts/contracts/core/SphinxModule.sol
```diff
@@ -34,7 +34,7 @@ contract SphinxModule is ReentrancyGuard, Enum, ISphinxModule, Initializable {
     /**
      * @inheritdoc ISphinxModule
      */
-    string public constant override VERSION = "1.0.0";
+    string public constant override VERSION = "1.0.0"; // TODO(later): what version?
```
sam-goldman, line 37: @RPate97 

Here's the new Sphinx Module. This design prioritizes simplicity and backwards compatibility. I haven't cleaned up this contract yet since I wanted to get your feedback on the design decisions first. I also haven't attempted to break the new contract yet.

### Cancellations

The key difference with this Module is that cancellations now occur in the `approve` function instead of a `cancel` function. I chose this design because it simplifies the Sphinx Module and Merkle tree logic.

It's necessary for the new Module to handle the following edge case:
1. Say Deployment 1 contains Transactions A, B, and C.
2. Say Transaction A is executed, but Transaction B reverts, causing the deployment to stall.
3. User signs Deployment 2, which cancels Deployment 1 and executes Transactions B and C.
4. Transaction B from Deployment 1 is executed successfully.
5. Deployment 2 is approved on-chain and executed successfully. This means Transaction B is executed twice, which is potentially dangerous.

This edge case is possible because deployments aren't explicitly marked as `FAILED` on-chain anymore. Notably, the Merkle root nonce is insufficient to prevent this scenario.

In the new design, we resolve this scenario by replacing the `merkleRootNonce` with two new fields in the `APPROVE` leaf: `previousMerkleRoot` and `previousLeavesExecuted`. A new Merkle root can only be approved if:
- The `previousMerkleRoot` matches the `latestMerkleRoot` (which is a new state variable in the Module).
- The `previousLeavesExecuted` matches the number of leaves executed in the `previousMerkleRoot`.

It'll be easy to collect these two fields during the proposal's collection phase.

These fields are essentially a more granular `merkleRootNonce`, so I removed that field because it's not necessary anymore.

### Pre-flight checks

You'll notice that the new Module doesn't contain any logic for pre-flight checks. I'd prefer if pre-flight checks are implemented as `EXECUTE` actions that are executed within the Gnosis Safe. This decreases the complexity of the new Module and allows us to change the implementation of the pre-flight checks without re-auditing the Module. This design decision means that deployments can be active without being executable yet. I'm not sure if this meaningfully increases the complexity of the executor.

### `activeMerkleRoot` -> `latestMerkleRoot`

I replaced the `activeMerkleRoot` state variable with the `latestMerkleRoot` state variable to support the cancellation design described above. The only difference between these two state variables is that the `latestMerkleRoot` is never set to `bytes32(0)`, whereas the `activeMerkleRoot` was set to `bytes32(0)` when a deployment was completed.

RPate97 (commented):
I'll be going through this more thoroughly later, but I just wanted to go ahead and post a comment regarding specifically pre-flight checks. 

> You'll notice that the new Module doesn't contain any logic for pre-flight checks. I'd prefer if pre-flight checks are implemented as EXECUTE actions that are executed within the Gnosis Safe. 

I agree, I want to use `EXECUTE` leaves to implement this functionality to keep it as simple as possible. However, I think we may need to include some relatively minimal logic in the module to ensure we can implement preflight checks using `EXECUTE` actions in a foolproof way.

Namely, I think we need to address the following scenario, which you mentioned to me a while ago:
1. I create a merkle tree which includes a call to contract A to perform some check. I expect this call to revert if the check is does not pass.
2. Some attacker executes the bundle on a network where A has not been deployed.
3. The check does not revert because A does not exist.
4. The attacker is able to execute the entire tree even though the check did not really pass. 

Just calling this out because I don't see anything in this PR related to this case, so I thought you might have forgotten about it.

To make sure this type of scenario is well covered in the audit, we should probably include an invariant in the spec that covers this general case. 

Sidenote: This particular topic might fall under your note:
> I also haven't attempted to break the new contract yet.

So if you want to wait until later to think about this sort of scenario, I think it's reasonable. I personally wouldn't because I think this impacts the spec, but it's really up to you.

sam-goldman:
> Namely, I think we need to address the following scenario, which you mentioned to me a while ago:
> 
> 1. I create a merkle tree which includes a call to contract A to perform some check. I expect this call to revert if the check is does not pass.
> 2. Some attacker executes the bundle on a network where A has not been deployed.
> 3. The check does not revert because A does not exist.
> 4. The attacker is able to execute the entire tree even though the check did not really pass.

Good catch - I added this to the `SphinxModule` and provided an explanation for how I resolved this scenario below. Also, in this initial review, you don't need to go too deep into the security model. I'm just looking for the outcome of this initial review to be either:
* "Nothing immediately jumps out to me as problematic, and I agree with the design decisions as long as they're secure"
* "XYZ design decision seems problematic to me for these reasons: ..." or "There's an obvious security vulnerability with this approach: ..."

I hope this is clear from your perspective. The main purpose of this initial review is to catch obviously incorrect design decisions before I spend a bunch of time testing, updating specs, etc.

I resolved the griefing scenario you mentioned by adding a new field to the `EXECUTE` leaf, `requireIsContract`. If this field is `true`, we check that the `to` field is a contract. Here's how this'd resolve the specific scenario you outlined, where a contract called Contract A must be deployed:
1. Define the pre-flight check contract, `IsContract`, which will look something like this:
```
contract IsContract {
  function check(bytes memory data) external {
    (address target) = abi.decode(data, (address));
    require(target.code.length > 0);
  }
}
```
2. When constructing the deployment, set the first `EXECUTE` leaf to be a call to `IsContract`, and set `requireIsContract` to `true`.
3a. If `IsContract` hasn't been deployed yet, the `SphinxModule` will revert because of the `requireIsContract` check.
3b. If `IsContract` has been deployed but Contract A hasn't been deployed, the call to `IsContract` will throw an error, causing the `SphinxModule` to also throw an error.
3c. If both contracts are deployed, the call will succeed.

packages/contracts/contracts/core/SphinxModule.sol
```diff
@@ -377,90 +269,31 @@ contract SphinxModule is ReentrancyGuard, Enum, ISphinxModule, Initializable {
                 uint256 gas,
                 bytes memory txData,
                 Enum.Operation operation,
-                bool requireSuccess
+                bool requireIsContract
             ) = abi.decode(leaf.data, (address, uint256, uint256, bytes, Enum.Operation, bool));
 
+            // If `requireIsContract` is `true`, check that the `to` field is a contract.
+            require(!requireIsContract || to.code.length > 0, "TODO(docs)");
```
RPate97, line 276: Minor comment: 
I expect an auditor would recommend checking the code hash instead of just checking the length. We know we don't want to do this because it may cause incompatibilities on some zk rollups. So we'll probably want to document this somewhere.

packages/contracts/contracts/core/SphinxModule.sol
RPate97, line 37: IMO, this should be a major version change since the data structure of the Merkle tree is changing.

RPate97 (approved):
Looks good, I don't see anything obviously problematic.

PR #1693, opened 2024-04-25
docs: Add SphinxEscrow contract spec

RPate97:
## Purpose
Adds the spec for the `SphinxEscrow` contract.

### DIFF
--- a/specs/escrow.md
+++ b/specs/escrow.md
@@ -1,0 +1,60 @@
+# `SphinxEscrow` Contract Specification
+
+When deploying via the Sphinx DevOps Platform, the user has the option to pay for their deployments using ERC20 tokens on a single network. Before the deployment begins, the user must grant an allowance on an ERC20 token to the `SphinxEscrow` contract that is sufficient to cover the estimated cost of the deployment. The ERC20 token is then locked in the `SphinxEscrow` contract while the deployment is being executed. After the deployment is completed, the Sphinx backend calculates the exact cost of the deployment and transfers the remainder back to the user.
+
+The Sphinx team owns the `SphinxEscrow` contract.
+
+## Table of Contents
+
+- [Relevant files](#relevant-files)
+- [Use Case](#use-case)
+- [High-Level Invariants](#high-level-invariants)
+- [Function-Level Invariants](#function-level-invariants)
+- [Assumptions](#assumptions)
+
+## Relevant Files
+
+- The contract: [`SphinxEscrow.sol`](todo)
+- Unit tests: [`SphinxEscrow.t.sol`](todo)
+
+## Use Case
+
+### Allow Sphinx to accept crypto payments for the variable cost of deployments
+For Sphinx to accept crypto payments we must be able to guarantee that we will be paid for the cost of the deployment. Since the exact cost of a deployment cannot be exactly known ahead of time, we cannot charge for the deployment upfront. So instead, we estimate the cost of a deployment and require that the user transfer the estimated amount to an escrow contract while the deployment is being executed. After the deployment is completed, we calculate the exact cost and transfer the remaining funds back to the user.
+
+## High-Level Invariants
+- It must be possible to use the `SphinxEscrow` contract with any arbitrary ERC20 contract.
+- It must be possible to use the `SphinxEscrow` contract to transfer ERC20 tokens from an `owner` address to the `SphinxEscrow` contract if the `SphinxEscrow` contract has been granted an allowance.
+- It must *not* be possible to transfer ERC20 tokens from an `owner` address to any address other than the `SphinxEscrow` contract if the `SphinxEscrow` contract has been granted an allowance.
+- It must *only* be possible to transfer ERC20 tokens out of the `SphinxEscrow` contract to the original ERC20 `owner` address or an address with the `DEFAULT_ADMIN_ROLE` on the `SphinxEscrow` contract.
+
+## Function-Level Invariants
+
+#### `constructor(address _owner)`
+- Must grant the `DEFAULT_ADMIN_ROLE` to the specified owner address.
+- Must revert if the `_owner` address is `address(0)`.
+
+#### `function claim(address _owner, address _erc20, uint256 _amount)`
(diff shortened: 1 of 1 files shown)


specs/escrow.md
```diff
@@ -0,0 +1,60 @@
...
+## High-Level Invariants
+- It must be possible to use the `SphinxEscrow` contract with any arbitrary ERC20 contract.
+- It must be possible to use the `SphinxEscrow` contract to transfer ERC20 tokens from an `owner` address to the `SphinxEscrow` contract if the `SphinxEscrow` contract has been granted an allowance.
+- It must *not* be possible to transfer ERC20 tokens from an `owner` address to any address other than the `SphinxEscrow` contract if the `SphinxEscrow` contract has been granted an allowance.
+- It must *only* be possible to transfer ERC20 tokens out of the `SphinxEscrow` contract to the original ERC20 `owner` address or an address with the `DEFAULT_ADMIN_ROLE` on the `SphinxEscrow` contract.
+
+## Function-Level Invariants
+
+#### `constructor(address _owner)`
+- Must grant the `DEFAULT_ADMIN_ROLE` to the specified owner address.
```
sam-goldman, line 34: It looks like we're using OpenZeppelin's `AccessControl` contract, but I only see the `DEFAULT_ADMIN_ROLE` used as a permission (in the `disburse` function). If we're planning on only having a single owner of this contract, we should use `Ownable` instead of `AccessControl`. If, instead, we're planning on allowing multiple addresses to withdraw funds, I think we should give them a specified role like we do in the `ManagedService` contract instead of giving them all the `DEFAULT_ADMIN_ROLE`. I think only one address should have the `DEFAULT_ADMIN_ROLE` because this address has permission to add or remove any other roles. This means that if we have multiple addresses with the `DEFAULT_ADMIN_ROLE` and we leak only one of their private keys, an attacker could remove all owners

specs/escrow.md
```diff
@@ -0,0 +1,60 @@
...
+
+#### `function claim(address _owner, address _erc20, uint256 _amount)`
+- Must revert if `_amount` is greater than the allowance of the `SphinxEscrow` contract on `_erc20` for `_owner`.
+- Must revert if the `_owner` does not have a balance of `_erc20` that is >= `_amount`.
+- A successful call must:
+  - Transfer the `_amount` of `_erc20` to the `SphinxEscrow` contract.
+  - Increment the `escrowBalance` of `_erc20` for `_owner` by `_amount`.
+  - Emit a `Claimed` event in the `SphinxEscrow` contract.
+
+#### `function disburse(address _owner, address _feeRecipient, address _erc20, uint256 _refund, uint256 _cost)`
```
sam-goldman, line 45: In both of these functions, the `_owner` parameter is the user, right? The semantics made me confused because the contract itself is owned by us

specs/escrow.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Function-Level Invariants
+
+#### `constructor(address _owner)`
+- Must grant the `DEFAULT_ADMIN_ROLE` to the specified owner address.
+- Must revert if the `_owner` address is `address(0)`.
+
+#### `function claim(address _owner, address _erc20, uint256 _amount)`
+- Must revert if `_amount` is greater than the allowance of the `SphinxEscrow` contract on `_erc20` for `_owner`.
+- Must revert if the `_owner` does not have a balance of `_erc20` that is >= `_amount`.
```
sam-goldman, line 39: Fyi these invariants may be enforced by the ERC20 standard itself, so it may be redundant to include these checks in our spec/contract. You don't need to worry about that though - I'll investigate this when I implement this contract

specs/escrow.md
```diff
@@ -0,0 +1,60 @@
...
+- A successful call must:
+  - Transfer the `_amount` of `_erc20` to the `SphinxEscrow` contract.
+  - Increment the `escrowBalance` of `_erc20` for `_owner` by `_amount`.
+  - Emit a `Claimed` event in the `SphinxEscrow` contract.
+
+#### `function disburse(address _owner, address _feeRecipient, address _erc20, uint256 _refund, uint256 _cost)`
+- Must revert if the amount of `_erc20` currently in escrow for `_owner` is < `_refund` + `_cost`
```
sam-goldman, line 46: > - Must revert if the amount of `_erc20` currently in escrow for `_erc20` is < `_refund` + `_cost`

This invariant allows for the possibility that only a fraction of the escrowed `_erc20` is sent to the `_owner` or `_feeRecipient`. For example, if the escrowed `_erc20` amount is 20 USDC, and `_refund` and `_cost` are both 5, then an error isn't thrown by this function. Is that intentional? I could see this being useful if the user wants to keep a balance of ERC20 for all deployments instead of adding a new allowance for each deployment. However, this use case isn't mentioned in the "Use Case" section, so it's not clear what the expected behavior is

specs/escrow.md
```diff
@@ -0,0 +1,60 @@
...
+- It must *only* be possible to transfer ERC20 tokens out of the `SphinxEscrow` contract to the original ERC20 `owner` address or an address with the `DEFAULT_ADMIN_ROLE` on the `SphinxEscrow` contract.
+
+## Function-Level Invariants
+
+#### `constructor(address _owner)`
+- Must grant the `DEFAULT_ADMIN_ROLE` to the specified owner address.
+- Must revert if the `_owner` address is `address(0)`.
+
+#### `function claim(address _owner, address _erc20, uint256 _amount)`
+- Must revert if `_amount` is greater than the allowance of the `SphinxEscrow` contract on `_erc20` for `_owner`.
```
sam-goldman, line 38: It looks like anybody can call this function to lock funds in the `SphinxEscrow` contract. This seems benign, but I'm not sure if it could mess up our backend logic. For example:
1. Griefer grants the `SphinxEscrow` contract an allowance
2. Griefer approves deployment
3. Griefer frontruns the `claim` function, causing our call to the `claim` function to throw an error

Unless there's a use case for keeping this open, I'd probably prefer to make it permissioned

sam-goldman (commented):
Looks solid. Got some questions / comments

PR #1699, opened 2024-04-30
Sg/migrate

sam-goldman:
Explores the question, "How should we migrate existing users to the newest version of the Sphinx Module?"

### DIFF
--- a/migration.md
+++ b/migration.md
@@ -1,0 +1,37 @@
+How should we migrate existing users to the newest version of the Sphinx Module?
+
+A couple notes:
+* I think this decision is likely to impact the migration process for future versions of the Sphinx Module.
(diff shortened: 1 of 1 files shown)


RPate97 (commented):
I agree option 3 is ideal, but I think it meaningfully increases the scope of the audit compared to other options. 

There are two why I think this adds a lot of complexity:
1. It immediately increases the complexity of the Merkle tree generation logic since it now needs to generate a tree that can be executed by two separate versions of the module. I don't immediately see a simple solution for dealing with that, so I have to assume it will add meaningful complexity. 
2. One of the main reasons why option 3 is appealing is because it seems like we could create a system that would work for future versions. I think that should be an explicit design goal, and we should devote effort to ensure we create a system that can be cleanly extended to cover migrating between arbitrary different module versions.

So, I agree that we should do option 3, but I think we should go into it understanding that it will probably add meaningful complexity.

PR #1710, opened 2024-05-08
fix(pg): Improve reliability of internal hardhat simulation in monorepos

RPate97:
## Purpose
- Fixes an issue where some mono repo installations would not work correctly because of the location of the `hardhat.config.js` file used to spin up our internal simulation.
- Fixes an issue where the user's version of hardhat may be used for our internal simulation if they have hardhat installed, causing unpredictable behavior.

## Solution
We no longer rely on a hardhat task to create the in-process hardhat node. Instead, we create the internal hardhat process provider from scratch and then convert it into an ethers-compatible provider, which we then use for the simulation. This required creating a slightly modified version of the `EthersHardhatProvider`, which is functionally equivalent but does not require a hardhat project to work. This setup eliminates the brittle process of invoking a hardhat subtask in a child process using the `hardhatRunner` script and an actual `hardhat.config.js` configuration file.

## Riders
This PR also includes a number of changes that resolve issues recently introduced by updates to Foundry and Hardat.
- Fixes [`CHU-1067`](https://linear.app/chugsplash/issue/CHU-1067/deployments-on-anvil-fail-due-to-blob-gas-issue-in-simulation) by updating to hardhat `2.22.3`.
- In hardhat `2.22.3`, a bug that causes the simulation to fail in certain situations when running against local nodes was introduced. I tried to resolve that issue but could not, so I disabled the simulation when the target network is a local node. This also required removing/updating some tests that relied on the simulation. I've [created a ticket](https://linear.app/chugsplash/issue/CHU-1072/re-enable-simulation-step-against-local-nodes) to reverse these changes if/when hardhat fixes the bug.


packages/plugins/script/Sample.s.sol
```diff
@@ -11,7 +11,7 @@ contract Sample is Sphinx {
     MyContract1 myContract;
 
     function configureSphinx() public override {
-        sphinxConfig.projectName = "my_sphinx";
+        sphinxConfig.projectName = "Simple_Project";
```
RPate97, line 14: This is a rider. It just makes the sample file work immediately with the `sphinx.lock` file in the plugins package.

packages/core/src/provider/hardhat-provider.ts
```diff
@@ -0,0 +1,1268 @@
...
+ * the execution process work with this provider, we must convert it a compatible provider. For that purpose, we've created this fork of the `HardhatEthersProvider`.
+ * The `HardhatEthersProvider` is perfect for our use case since it is designed to make the in-process hardhat node compatible with ethers. However, the standard
+ * implementation requires a hardhat project to be used which is problematic for our situation.
+ *
+ * So this fork makes a minor modification to the `HardhatEthersProvider` and `HardhatEthersSigner` classes to allow them to be used outside of a hardhat project.
+ * Where the original implementation relied on reading the hardhat configuration options directly from the `HardhatRuntimeEnvironment` global var, this implementation
+ * accepts a `network` object in the `InProcessEthersProvider` constructor which includes all of the necessary configuration options.
+ *
+ * To review the exact changes we made to this class, you can search `Sphinx` in this file and in the associated `hardhat-signer.ts` file.
+ *
```
RPate97, line 20: Read the comment. 

I don't recommend reviewing anything else in this file, as it's just a copy of the `HardhatEthersProvider` with very minimal modifications. I've left comments indicating the specific changes I made if you'd like to look through those quickly.

packages/core/src/provider/hardhat-provider.ts
```diff
@@ -0,0 +1,1268 @@
...
+
+  constructor(
+    private readonly _hardhatProvider: EthereumProvider,
+    /**
+     * Sphinx: We modified this class to accept an entire Network configuration object
+     * instead of just the name. We later pass this class to the associated signer so that
+     * it does not need to read from the hre global variable.
+     */
+    private readonly _network: Network
+  ) {}
```
RPate97, line 130: See comment.

packages/core/src/provider/hardhat-provider.ts
```diff
@@ -0,0 +1,1268 @@
...
+        accounts[address],
+        // Sphinx: Here we pass the network object to the signer
+        this._network
+      )
+    }
+
+    if (typeof address === 'string') {
+      // Sphinx: Here we pass the network object to the signer
+      return InProcessEthersSigner.create(this, address, this._network)
+    }
```
RPate97, line 168: See comments.

packages/core/src/provider/hardhat-signer.ts
```diff
@@ -0,0 +1,325 @@
...
+
+export class InProcessEthersSigner implements ethers.Signer {
+  public readonly address: string
+  public readonly provider: ethers.JsonRpcProvider | InProcessEthersProvider
+
+  public static async create(
+    provider: InProcessEthersProvider,
+    address: string,
+    // Sphinx: We accept a network object here which is then used to configure the signer instead of reading from the HRE.
+    network: Network
```
RPate97, line 42: See comment

packages/core/src/provider/hardhat-signer.ts
```diff
@@ -0,0 +1,325 @@
...
+      // has a fixed number as the gas config, we use that.
+      // Hardhat core already sets this value to the block gas limit when the
+      // user doesn't specify a number.
+      if (network.config.gas !== 'auto') {
+        gasLimit = network.config.gas
+      }
+    } else if (network.name === 'localhost') {
+      // Sphinx: Since this signer is only intended to be used with an internal hardhat node, I've removed the logic here which
+      // simplifies the changes required to this class and to the `InProcessHardhatProvider`.
+      throw new Error('Only internal in-process network is supported')
```
RPate97, line 58: See comment

packages/plugins/test/mocha/simulate.spec.ts
```diff
@@ -172,48 +172,11 @@ describe('Simulate', () => {
   })
 })
 
-describe('simulateDeploymentSubtask', () => {
```
RPate97, line 175: As long as I understood the purpose of this test correctly, I believe we can drop it since we don't have a `simulateDeploymentSubtask` function anymore. Let me know if you think we do need something to replace it though.

packages/core/src/provider/hardhat-provider.ts
```diff
@@ -0,0 +1,1268 @@
...
+ *
+ * So this fork makes a minor modification to the `HardhatEthersProvider` and `HardhatEthersSigner` classes to allow them to be used outside of a hardhat project.
+ * Where the original implementation relied on reading the hardhat configuration options directly from the `HardhatRuntimeEnvironment` global var, this implementation
+ * accepts a `network` object in the `InProcessEthersProvider` constructor which includes all of the necessary configuration options.
+ *
+ * To review the exact changes we made to this class, you can search `Sphinx` in this file and in the associated `hardhat-signer.ts` file.
+ *
+ * You might wonder why we made a full copy of the both classes just to make these modifications. The reason is that `HardhatEthersSigner` only implements a private
+ * constructor which makes it impossible for us to use subclasses to make the required changes.
+ */
```
RPate97, line 23: Please read the whole comment carefully.

I don't recommend reviewing much else in this file, as it's just a copy of the HardhatEthersProvider with minimal modifications. I've left comments indicating my specific changes, though, if you'd like to look through those quickly.

packages/plugins/package.json
```diff
@@ -65,7 +64,7 @@
     "typescript": "^5.1.3",
     "yargs": "^17.7.2",
     "yesno": "^0.4.0",
-    "hardhat": "2.20.1"
+    "hardhat": "2.22.3"
```
RPate97, line 67: This fixes [CHU-1067](https://linear.app/chugsplash/issue/CHU-1067/deployments-on-anvil-fail-due-to-blob-gas-issue-in-simulation).

.circleci/config.yml
```diff
@@ -26,7 +26,7 @@ jobs:
           command: sh ./scripts/create-placeholder.sh
       - run:
           name: Install dependencies
-          command: yarn --frozen-lockfile --link-duplicates --network-concurrency 1
+          command: yarn --frozen-lockfile --network-concurrency 1
```
RPate97, line 29: Removing this option fixes an issue I ran into after bumping the hardhat version. The option is [known to be unstable](https://github.com/yarnpkg/yarn/issues/2734) and doesn't add much value in our case.

For context, this option is generally used to reduce the size of the installed dependencies in situations where that would help with performance such as when working on a hosted frontend.

packages/plugins/src/cli/index.ts
```diff
@@ -6,4 +6,38 @@ import { makeCLI } from './setup'
 // Load environment variables from .env
 dotenv.config()
 
+let getaddrinfoErrors = 0
+// If an error occurs, we write the error message and stack trace to `stdout` then exit the process
+// with exit code `1`. We write the error to `stdout` instead of `stderr` because `stderr` may
+// contain warnings that were written via `console.warn`, which are indistinguishable from the
+// actual error message in `stderr`. By using `stdout`, we can throw an error that doesn't contain
+// warnings in the parent process.
+process.on('uncaughtException', (error) => {
```
RPate97, line 15: We still need to catch `getaddrinfo ENOTFOUND` errors, but since we no longer invoke the hardhat runner as a separate script, we have to set this up when the CLI is first called.

packages/plugins/test/mocha/cli/deploy.spec.ts
```diff
@@ -404,12 +404,11 @@ describe('Deploy CLI command', () => {
...
-    // collected, but before any transactions are broadcasted. This is worthwhile to test because
-    // the `SphinxModule` doesn't revert if a user's transactions causes the deployment to be marked
-    // as `FAILED`. If the Foundry plugin doesn't revert either, then Foundry will attempt to
-    // broadcast the deployment, which is not desirable.
-    it('Reverts if the deployment fails during the simulation', async () => {
+    // This test checks that an error is thrown if a transaction fails during execution. This is worthwhile
+    // to test because the `SphinxModule` doesn't revert if a user's transactions causes the deployment to be
+    // marked as `FAILED`. If the plugin doesn't revert either, then the user may believe their deployment
+    // completed successfully.
+    it('Errors if the deployment fails during execution', async () => {
```
RPate97, line 411: The previous version of this test relied on the simulation running against local nodes. Since we do expect an error to be thrown during execution if an action reverts when the transactions are being broadcast, I updated this test to reflect that behavior.

packages/plugins/test/mocha/cli/propose.spec.ts
```diff
@@ -192,15 +192,6 @@ describe('Propose CLI command', () => {
...
     ).equals(expectedContractAddress)
-
-    await assertValidGasEstimates(
-      proposalRequest.gasEstimates,
-      networkConfigArray,
-      networks,
-      scriptPath,
-      targetContract,
-      context
-    )
```
RPate97, line 203: Unfortunately, these tests won't pass if the simulation is disabled, which is the case for these tests. I could not find a way around that requirement, so I've disabled them.

I've made a note in the [relevant ticket](https://linear.app/chugsplash/issue/CHU-1072/re-enable-simulation-step-against-local-nodes) that these should be reenabled if/when we can use the simulation on local nodes again.

packages/plugins/test/mocha/foundry/utils.spec.ts
```diff
@@ -209,7 +209,7 @@ describe('Utils', async () => {
       // Fifth contract:
       expect(
         existsSync(
-          `${artifactFolder}/test/deeper/DuplicateContractName.sol/DuplicateContractName.json`
+          `${artifactFolder}/deep/deeper/DuplicateContractName.sol/DuplicateContractName.json`
```
RPate97, line 212: This is a rider. One of our tests was failing because Foundry made a change that caused this artifact to be output in a different directory. I believe they fixed a bug with this change, and it does not appear to affect us outside of this test so I just updated the test.

packages/plugins/src/hardhat/simulate.ts
```diff
@@ -337,14 +255,58 @@ export const fetchTransactionResponses = async (
...
+  chainId: string,
+  blockGasLimit: string,
+  blockNumber: string | undefined,
+  cachePath: string
 ): Promise<{ transactions: SimulationTransactions }> => {
-  // Wrap the Hardhat provider with a Proxy, which implements retry and timeout logic.
-  const provider = createHardhatEthersProviderProxy(hre.ethers.provider)
+  const config: HardhatConfig = resolveConfig('', {
+    paths: {
+      cache: join(process.cwd(), cachePath, '/sphinx'),
```
RPate97, line 268: Previously the cached files would have been written to a `cache` folder that is located with the hardhat project (inside of the `node_modules` folder where the `hardhatRunner` was), but since we now don't use a hardhat project, it will write all the files to a `cache` folder in the users project by default. To keep things simple for the user, I updated it so that the files are written into their Foundry cache folder instead.

sam-goldman (approved):
Looks great

